PR #25, opened 2023-05-02
Font fix for headless linux systems

nichollsh:
Previously, PROTEUS was set up to plot using Arial only. This caused problems when running the code on systems without Arial installed, including the AOPP cluster.

I also noticed a bug in the plotting functions where the 'times' array was compared using ==, which is always consistent. A minor change to the default value of 'times' fixed this.

### DIFF
--- a/utils/cpl_atmosphere.py
+++ b/utils/cpl_atmosphere.py
@@ -178,8 +178,6 @@ title_ha   = "left"
 title_va   = "bottom"
 title_font = 'Arial'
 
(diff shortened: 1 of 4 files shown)


timlichtenberg (changes requested):
All good, aside:
`time_l = [int(time) for time in time_l.split(',')]`
This is going to break if there is only one time set, sometime you want to do this. The former solution prevents a crash in this case.

timlichtenberg (approved):
All good now

nichollsh:
Thank you!

PR #47, opened 2023-08-03
Offline chemistry, minor additions, and bug fixes

### LINKED ISSUE #42: SPIDER call redox chemistry
Goal: Implement SPIDER call sequence to specific redox state + H/C ratio instead of total volatile budget or partial pressure

@djbower Could you provide us with an example to call sequences to SPIDER that do that so we can implement this into PROTEUS? 

Are there incompatibilities we should be aware of with other call variables? The list of used variables is here (line 67 ff.): https://github.com/FormingWorlds/PROTEUS/blob/master/init_coupler.cfg

nichollsh:
### Summary

This branch contains quite a few changes from the master branch, but I have confirmed that it runs PROTEUS fine. The main purpose of this branch was to include a method for running chemistry 'offline' from the main PROTEUS simulations; the script `tools/RunOfflineChemistry.py` takes the output from a PROTEUS run and provides it to VULCAN. This allows us to investigate the response of the upper atmosphere.

### Major features
 * Script to run VULCAN in offline-chemistry mode. This takes data from
 a completed PROTEUS run, and calculates the disequilibrium composition 
 for several samples across the evolution of the planet. 
 
 * Can now set a parameterised initial composition based on fO2, C/H, 
 H2O inventory, and N2 ppm. This uses Dan's script for satisfying both 
 equilibrium chemistry and melt-vapour partitioning functions. Can still 
 manually choose the composition by disabling this in the config. 
 Closes #42 .

 * Several more configuration options for existing features (e.g. to
 disable stellar heating, change tropopause type, choose spectral file).
 
 * Added skin-temperature as option for tropopause temperature.
 
 * Adaptive time-stepping for selecting new SPIDER step size. This can
 be turned off in favour of other time-stepping methods (e.g. const dt).
 
 * Introduction of 'eqm' loops, which run the SPIDER-AEOLUS iterations
 with dt=0 in order to adapt to changes in instellation. This is not 
 effective for optically thick atmospheres and may be redundant soon.
 
 * Can limit the relative change in net upward flux each time-step, to
 prevent oscillations. This isn't physical in many circumstances.

 
 
### Minor features
 * Can change the frequency of updating the instellation/spectrum, and 
 for creating plots.
 
 * Limits on maximum and minimum temperature
 
 * Plotting scripts for visualising offline chemistry results and 
 debugging the code. 
 
 * Introduced 'input' directory containing stellar tracks, example cfg
 files, and stellar spectra.
 
 * Documentation updates, although much more is still needed. Includes
 changes to read-the-docs information.

 * Script for running PROTEUS using Screen. This is helpful for managing
 multiple simultaneous instances of PROTEUS.
 
 * For a given run 'x', files are stored in `output/x/`. JSON and PKL
 data files are stored in `output/x/data/`.
 
 * Changes to information printed to stdout. It was getting a bit
 unreadable with the new features, so I've just tidied this up.
 
 * Increased number of SPIDER levels from 100 to 150.
 
 * Split up some large py files into several smaller ones.
 
 * Different options for initialising offline chemistry. Can provide 
 elemental abundances based on outgassed volatiles, or directly provide
 constant mixing ratios based on outgassed volatiles.
 
 * Calculated stellar spectra are passed to VULCAN during offline 
 chemistry calculations.
 
 * Other scripts in `tools/`.
 
 * Can choose where the boundary condition `F_atm` is calculated: either
 at the surface or at TOA. This will be necessary for future changes.
 
 * More variables stored in the helpfile, which are needed for
 debugging the model.
 
 * Changes to which SPIDER variables are stored in `init_spider.opts` 
 and which are provided at runtime. 
 
 * New folder `examples/` containing example outputs. Most notably, 
 shows the result of running the current default configuration file.
 

### Fixes
 * Several bug fixes relating to stellar evolution calculations. These
 include scaling luminosity with the Baraffe model and also performing
 track interpolation.
 
 * Uses the new spectral file 'Mallard' by default, which can recover 
 a runaway H2O greenhouse effect.
 
 * Stores updated stellar radius as the star ages. Previously was not
 being updated in the helpfile.
 
 * Much-improved method for choosing UV-PL boundary when using Mors 
 evolution code. Previously it was fixed to be at the Balmer jump, 
 but now is chosen (by bisection search) to be at a value that best 
 reproduces the correct Lyman-alpha energy. This is helpful when running
 the model for various stellar classes.
 
 * Added Hamish Innes to contributors section.


nichollsh:
Note that the code experiences some numerical issues with thick atmospheres when the instellation is updated. The solution to this is to set `F_atm_bc=1` and implement an atmosphere model which solves for global radiative equilibrium. I did not include radiative time-stepping in this pull request because my toy model is not appropriate for the master branch.

examples/examples_readme.txt
```diff
@@ -0,0 +1,2 @@
+This folder contains example PROTEUS runs. They are dated to when they were
```
timlichtenberg, line 1: Ok for now, for the next example plots and configs I would suggest to also add the exact PROTEUS version with which they were generated. In this case, I presume this was commit `a1d24fe`.

init_coupler.cfg
```diff
@@ -10,14 +10,15 @@ star_radius_modern          = 0.1192            # R_sun, radius of star (TODAY)
...
-star_spectrum               = spec_trappist-1.txt   # Stellar spectrum file at 1 AU (TODAY)
+star_spectrum               = input/spec_trappist-1.txt   # Stellar spectrum file at 1 AU (TODAY)
 star_btrack                 = BHAC15-M0p090000.txt
 
 
 ##### Planet configuration
 
+
 mean_distance               = 0.700     # AU, star-planet distance
-gravity                     = 9.81      # m/s^2, planet surface gravity
```
timlichtenberg, line 20: Where is gravity now defined/calculated?

timlichtenberg (approved):

timlichtenberg:
@nichollsh I am going to merge the pull request, great work again! Please double check that gravity is correctly read in/calculated, it is deleted from `init_coupler.cfg`. See my inline comment.

init_coupler.cfg
nichollsh: Gravity is now calculated from the mass and radius of the planet when the config file is read by `ReadInitFile()`. I switched over because AEOLUS now takes in planet mass when creating the `atmos` object. Maybe we could change it so that you can pass in either one, and it'll work out the other variable.

examples/examples_readme.txt
nichollsh: I think that's a good idea. These were created with commit [43a98a1](https://github.com/FormingWorlds/PROTEUS/pull/47/commits/43a98a113ced8a14acbbc24370b80a180da68a02), which differs from the merged version only by the ReadTheDocs changes.

PR #67, opened 2024-02-15
Significant model stability improvements, plotting updates, and other changes

### LINKED ISSUE #61: SPIDER fails to find a solution for permanently-molten cases
I have already been investigating this, but will make an issue here in case this continues to be a problem or re-emerges after being 'fixed'.

In some cases, SPIDER will fail to obtain a solution. Either of the two errors are possible in this case:
- `[CVODE ERROR]  CVode At t = 4.95363e+07 and h = 0.000184003, the error test failed repeatedly or with |h| = hmin`
- `PETSC ERROR: Nonlinear solver didn't converge: DIVERGED_FNORM_NAN`

These occur under the following conditions:
- When the flux boundary condition provided to SPIDER (from `F_atm`) is small (i.e. less than ~5 W m-2)
- When the time-step is too large.

Therefore, this issue most often occurs in the case of Type II planets, where the planet should stay molten and the net outgoing flux rapidly tends to zero within ~100 kyr. One way to mitigate this is to include a "steady-state break condition", which I've already done in a development branch. However, the model still fails in some cases.

This is still being investigated, but I expect that it should be resolved soon.

### LINKED ISSUE #65: Optimise spider.py
Currently, `spider.py` makes a lot of file I/O operations which slows down the model increasingly as the simulation progresses. This is because it reads-in all of the prior .json files when doing some of the spider-related stuff. It reads them in several times, depending on the variables being accessed and the calculations being performed, meaning that the performance drops off quickly (super-linearly) with model runtime. It would be a *lot* faster if they were just read-in once, and then analysed together.

nichollsh:
This PR contains some significant changes and some minor changes.

**Major changes**
- Model stability is significantly enhanced. Previously, there were cases where SPIDER would crash due to non-convergence when the fluxes are small and/or the time step was large. This has been resolved by using a form of step-rejection. To summarise this method: within `spider.py`, the SPIDER subprocess is now called via a wrapper function `_try_spider()` which attempts to run the interior model. If SPIDER fails, PROTEUS will detect this (instead of just crashing) and will attempt to run SPIDER again with an tolerance and step-size. This is repeated up to 7 times, re-adjusting the tolerances and step each time. Closes #61.
- Improvements to batch execution via `tools/GridPROTEUS.py`. The model can now be reliably run for a grid of parameters, with the status of each grid point being tracked by a 'status file' that is updated within each output folder. The status of a grid can be analysed with `tools/GridSummarise.py`. There's also a successor to this called `GridPROTEUS2.py` but it is not complete and is considered unsafe to use at the moment.
- A new test for model convergence when the planet enters into a steady state. This occurs when global radiative equilibrium is reached (the so-called ERCS from Salvador+23), but the MO hasn't solidified. The two criteria for this are: that the outgoing flux (`F_atm`) is small, and that the global melt fraction (`phi_global`) is unchanging in time.
- PROTEUS now uses the most up-to-date version of AEOLUS, which itself has undergone some changes. Notably, in fixing a bug in the calculation of gas MMW from the volatile VMRs.

**Minor changes**
- Re-wrote `cpl_global.py` to be much more efficient in terms of file I/O, which comes with a significant performance benefit as the simulations progress, potentially saving days of CPU time. Closes #65. 
- The model time-step is now capped when the net fluxes are small, as large time-steps run the risk of attaining an (unphysically) negative upward net flux rather than asymptotically tending to zero. This limits performance in some cases, but is mostly mitigated by the new convergence criteria outlined above. This is something to work on in the future.
- Re-formulated the colour palette to be accessible for colour-blind people and more consistent across plots.
- Some new plot types which enable interesting analysis (I am particularly proud of `cpl_interior_cmesh.py`). Removed some old plotting scripts which are redundant and/or unsupported.
- Documentation updated with appropriate installation instructions and the new configuration options. 
- In theory this update to PROTEUS makes it compatible with the 'recent' AGNI changes, but that model will also be seeing some significant changes in the future, so this is now less important.
- New notebooks and tools within `tools/` for processing the model outputs and debugging the code.
- Updated example in `tests/` for an Earth-like case with an initial 100 bar pure-steam atmosphere.


nichollsh:
A note on performance...
The model runtime varies significantly depending on the parameters you choose. For reference, the example case in `tests/` took 6.39 hours to complete on an Intel Xeon 6240R CPU. In my tests, the worst-performing case took 29 hours to complete since it evolved for several 100 Myr before converging.

timlichtenberg (approved):

PR #78, opened 2024-03-19
Support for JANUS, logging, GridProteus, and stability improvements

nichollsh:
### Summary
- Support for JANUS now that it has been renamed from AEOLUS.
- Support for MORS now that it has been cloned to FormingWorlds.
- PROTEUS now supports logging directly, so we don't need to rely on screen or awk.
- Spectral file band information is passed to JANUS when required, since it was previously hardcoded.
- Significant improvements to solver time-stepping, so that it can adapt depending on how fast the net radiative flux and the global melt fraction are evolving. 
- Increased default resolution in SPIDER, since using only ~100 levels was not properly resolving the uppermost layers of the MO. This is important when the planet reaches radiative equilibrium at small values of `phi_global`.
- GridPROTEUS now uses a queueing system, so that we can request more grid points than the number of available cores.
- Added a "dummy atmosphere" module, which is not (very) physical but is incredibly helpful for debugging.
- Documentation updates, so that it should be a lot more readable.
- Added functionality for killing PROTEUS when you don't have shell-access (e.g. when it's run by GridPROTEUS). The model will terminate if the file `keepalive` in the output directory is removed.

### DummyAtmosphere
While debugging some convergence issues - which were ultimately fixed by the new time-stepping scheme and the increased interior mesh resolution - I developed a new module for the atmosphere. It is called DummyAtmosphere, and can be found inside `utils/`. This takes the place of JANUS/AGNI by providing values for the upward/downward radiative fluxes parameterised by a value `gamma`. This value is a measure for the radiating temperature of the atmosphere just above the surface, relative to the surface temperature itself. We can therefore represent an optically thick and luminous atmosphere with gamma=1, or an entirely transparent atmosphere with gamma=0. This is not very physical, but it is still useful.

### GridPROTEUS 
It's now possible to set the number of desired processes in the GridPROTEUS script. The script will then only dispatch this many runs at a given time, with the remainder waiting in a queue. This is probably redundant thanks to other tools like [PySlurm](https://github.com/PySlurm/pyslurm), but it works better than the previous implementation. I've tested this on a grid of 700 points.


tools/SolveVolatileAbundances.py
```diff
@@ -268,15 +268,14 @@ def get_total_pressure(pin, fO2_shift, global_d):
 
 #====================================================================
 def atmosphere_mass(pin, fO2_shift, global_d):
-    """Atmospheric mass of volatiles and totals for H, C, and N"""
+    """Atmospheric mass of volatiles and totals for H, C, O"""
```
timlichtenberg, line 271: What about N? It's in there.

timlichtenberg (approved):
Two minor questions on the volatile solubility file

tools/SolveVolatileAbundances.py
```diff
@@ -561,14 +588,16 @@ def equilibrium_atmosphere_GR():
     global_d = get_global_parameters()
 
     
-    pl_m = 8.63 * 5.972e24 # kg
+    pl_m = 8.63 * 5.972e24 # kg, known k2-18b mass
```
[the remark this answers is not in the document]
nichollsh: This script is not used by the main PROTEUS code. It's currently configured for the K2-18b paper, but I agree it should be more generalised if we want to use it for other things in the future. It only propagates within this script.

tools/SolveVolatileAbundances.py
nichollsh: Ah, yes, it is. I meant to add O to this comment (since I added the O calculation to the function) but didn't mean to remove the N from the comment.

PR #85, opened 2024-04-04
Plot updates. Import optimisations. Config updates.

nichollsh:
Version of code used for article.
Minor changes targeted at reducing the number of libraries/files imported by the code.
Plot updates.


input/jgr_grid.cfg
```diff
@@ -60,14 +60,14 @@ dt_initial                  = 8e2    # Inital step size
 # Flux convergence scheme and tolerances for surface equilibration
 flux_convergence            = 0         # 0: off | 1: on | 2: restart
 F_atm_bc                    = 0         # Boundary condition choice for F_atm, 0: TOA | 1: Surface
-F_crit                      = 0.1       # Critical flux, below which attempt to stabilise the evolution (0: disabled)
+F_crit                      = 0.0      # Critical flux, below which attempt to stabilise the evolution (0: disabled)
```
timlichtenberg, line 63: Not necessary anymore?

input/jgr_grid.cfg
```diff
@@ -60,14 +60,14 @@ dt_initial                  = 8e2    # Inital step size
...
 F_eps                       = 1.0       # W m-2
 F_diff                      = 0.1       # flux fraction
 RF_crit                     = 0.01      # depth fraction
 dTs_atm                     = 30        # K
 skin_d                      = 0.01      # m
 skin_k                      = 2.0       # W m-1 K-1
-prevent_warming             = 0         # Require that the planet only cool down over time, 0: disabled | 1: enabled
```
timlichtenberg, line 70: Still find the right equilibrium state?

timlichtenberg (approved):

input/jgr_grid.cfg
nichollsh: With the new time-stepping changes, I've found that this isn't needed. The step size is constrained by the relative change in the flux (and melt fraction), so if the flux is small then only small changes will be allowed to the flux anyway. Probably worth leaving this parameter in for the future, since it might be helpful later on.

input/jgr_grid.cfg
nichollsh: In some cases, they do wish to heat up from the initial surface temperature of 3200 K since they have a net negative flux. However, since the flux is the limited to be very small, the model sees this as the planet entering into a steady state with in only a handful of iterations.

Physically, this means slightly under-estimating the very hottest cases. It does not mean that the outcome of the simulations will be wrong: all cases marked as solidified will always have solidified (independent of this parameter) and all steady-state cases would've done so as well.

I set it to 0 because allowing the interior to heat up is very very slow, and not something that we'd wish to explore anyway.

timlichtenberg: Not for now, but I do think there is a scientifically interesting case exploring this further sometime in the future.

PR #88, opened 2024-05-07
Updates to AGNI interface

nichollsh:
- Interface working again with latest AGNI version
- Removed old files
- Updated config file
- Updated documentation
- Moved outgassing calculation into new file (from spider.py)
- Logger updates


input/rce_demo.cfg
timlichtenberg, line 1: Is this file also a potential test case for a Github action?

timlichtenberg (approved):

input/rce_demo.cfg
nichollsh: That could be something to consider. Although, it's probably easier to put the tests in AGNI.

timlichtenberg (approved):

PR #98, opened 2024-05-29
Rewrite SolveVolatiles outgassing script and other minor changes

### LINKED ISSUE #92: Remove Sciath submodule
This submodule is ostensibly required by SPIDER in order to perform its tests. However, it's possible to run the tests without having to install this extra package. I assume it would also not be integrated into the SPIDER replacement (Aragog).

### LINKED ISSUE #94: Move volatile calculations outside of SPIDER
Outgassing is currently handled by SPIDER. Since it is not handled by the SPIDER replacement (aragog - issue #62), we will need to separate the interior and outgassing models.

In the mean time, we can take initial steps by handling volatile outgassing entirely via the SolveVolatileAbundances script (which does the same thing anyway).

Eventually, SolveVol will be replaced as in issue #82.

nichollsh:
* Removed sciath submodule (closes #92)
* Rewrite SolveVolatileAbundances so that it is applied at *every* PROTEUS step, not just for bootstrapping the atmosphere. This is more physically consistent, and also means that SPIDER does not have to handle any outgassing. Closes #94.
* Doc updates
* Convergence check for when planet intends to heat up at high instellations
* Minimum step requirement


docs/source/usage.rst
```diff
@@ -482,56 +482,50 @@ configuration, but they must all be passed via the config file.
...
      - Greater than zero. 
 
    * - ``X_included``
-     - Flag to include or exclude a volatile X from SPIDER and SOCRATES. This value may be overwritten at runtime but will be saved to the helpfile.  
+     - Flag to include X in the model. For H2O, CO2, and N2 this will always equal 1.
      - False
      - Integer
      - 0: Excluded, 1: Included.
 
-   * - ``X_add_bars``
```
timlichtenberg, line 527: This is outcommented, meaning this is not possible henceforth?

init_coupler.cfg
```diff
@@ -60,7 +60,6 @@ dt_initial                  = 4e2    # Inital step size
 # Flux convergence scheme and tolerances for surface equilibration
 flux_convergence            = 0         # 0: off | 1: on | 2: restart
 F_atm_bc                    = 0         # Boundary condition choice for F_atm, 0: TOA | 1: Surface
-F_crit                      = 0.1       # Critical flux, below which attempt to stabilise the evolution (0: disabled)
```
timlichtenberg, line 63: Why is this gone?

init_coupler.cfg
```diff
@@ -111,9 +114,6 @@ grain_size                  = 0.1
 # N2 partition coefficients, 0: oxidized | 1: reduced
 N2_partitioning             = 0
```
timlichtenberg, line 115: This could go as well now, right?

utils/spider.py
```diff
@@ -482,11 +459,6 @@ def _try_spider( time_dict, dirs, COUPLER_options, loop_counter, runtime_helpfil
             else:
                 UpdateStatusfile(dirs, 20)
                 raise Exception("Invalid time-stepping method '%d'" % COUPLER_options["dt_method"])
-            
-            # Additional step-size ceiling when F_crit is used
-            if abs(run_atm.iloc[-1]["F_atm"]) <= COUPLER_options["F_crit"]:
```
timlichtenberg, line 487: I think it would be better to keep this in, even though not actively using it. Having step size control as a function of both atmospheric flux AND melt fraction change seems good to me.

timlichtenberg (approved):
A few comments, most important on F_crit, i would keep that in.

docs/source/usage.rst
nichollsh: Yes, adding extra bars is not possible alongside the parameterised calculation of partial pressures from C/H, etc. This is because adding these extra bars overspecifies the initial condition, such that these parameters (C/H, etc) lose all meaning. This lead to my confusion with the pure-steam case that I thought was 6 oceans but was actually 30.

init_coupler.cfg
nichollsh: It wasn't being used by the version of PROTEUS that I branched for these changes. However, I've now repurposed this flag further down in the cfg file as a check for when the planet intends to heat up, since we assume secular cooling. I've tried to explain this a bit in the documentation updates.

init_coupler.cfg
nichollsh: Yes

utils/spider.py
nichollsh: I agree that this might be a useful thing to include. However, would it be possible to add this back to the code at a later time? I don't want to modify the code before merging this PR, since it's the version that grid was run with (for the paper).

init_coupler.cfg
nichollsh: Yes

init_coupler.cfg
nichollsh: It wasn't being used in any of the default cases or the grid. However, I repurposed this configuration option further down the cfg file as a check for when the planet intends to heat up, since we assume secular cooling.

PR #101, opened 2024-06-28
MORS updates and minor fixes

nichollsh:
Spectral synthesis is now included in MORS. This previously duplicated code, but with these changes PROTEUS no longer does any of the per-band synthesis, and it's entirely handled by MORS.

There's still code relating to the Baraffe evolution tracks in PROTEUS, which will need to be moved somewhere eventually. I have decided to leave this for now.

Also updated the default cfg file, and some minor fixes to the AGNI wrapper.


lsoucasse:
Hi @nichollsh,

This work sounds great. It is important to update Proteus with new Mors and it lightens the "utils" script which aligns with our development goals.

I have two questions for you:
- Will the Baraffee tracks be moved into Mors as well as a second step?
- Why does the init_coupler.cfg file in main folder differ from the init_coupled,cfg file in the examples/earth_demo folder? Overall, we need to address the storage of input data, input config giles and reference ouput results.

nichollsh:
Thanks @lsoucasse.

I had been working on the assumption that the Baraffe tracks will be dealt with at a later date, and this is to just get Mors working under the recent changes. We could certainly move the tracks into Mors, or maybe even make a separate small module/package for them. I am leaning towards the second possibility.

I am not seeing any differences between the two `init_coupler.cfg` files in [e7e162e](https://github.com/FormingWorlds/PROTEUS/pull/101/commits/e7e162ec90f8d7c14a2acfd5e14584895d1ce73a).

lsoucasse:
> Thanks @lsoucasse.
> 
> I had been working on the assumption that the Baraffe tracks will be dealt with at a later date, and this is to just get Mors working under the recent changes. We could certainly move the tracks into Mors, or maybe even make a separate small module/package for them. I am leaning towards the second possibility.
> 
> I am not seeing any differences between the two `init_coupler.cfg` files in [e7e162e](https://github.com/FormingWorlds/PROTEUS/pull/101/commits/e7e162ec90f8d7c14a2acfd5e14584895d1ce73a).

OK, the two files are indeed the same, I was miscomparing. We need to avoid duplicating this file in future work.

lsoucasse (approved):

PR #102, opened 2024-07-10
Update AGNI, MORS, and sulfur outgassing

nichollsh:
- Updates MORS to latest version, requiring you to download data into a FWL data directory
- Fixes to AGNI interface following model changes. Has been tested versus JANUS.
- Now handles sulfur outgassing through S2 solubility into the melt. This is then converted to SO2 via oxidation in the atmosphere, with eqm constant calculated from JANAF fit.
- Updated COUPLER_options and other cfg stuff, plots, helpfile IO to support S2 and SO2.
- Executes FastChem at runtime via AGNI.

Like N2, S2 must be included in all model runs. However, you can set its partial pressure (or total S abundance) to zero if you don't want to model it.

Does not update JANUS to latest version. For the model to run, you will need to patch JANUS with molar_mass data for S2 since it does not currently support it. This should be done in the next JANUS update.


input/janus_mixed.cfg
```diff
@@ -30,15 +30,15 @@ iter_max                    = 9000
...
+dir_output                  = jans_mixed_hot
 
 # Choose times
 time_star                   = 100.0e6  	# yr, time since star formation
 time_planet                 = 0.        # yr, time since MO start
 time_target                 = 4.567e+9  # yr, target time for MO evolution
 
 # SOCRATES spectral file to use (relative to JANUS folder)
-spectral_file               = AGNI/res/spectral_files/nogit/Dayspring/48/Dayspring.sf
+spectral_file               = JANUS/spectral_files/Dayspring/256/Dayspring.sf
```
lsoucasse, line 42: Seeing this line, I am wondering if you plan to mimic what we did in Janus in AGNI, i.e. downloading/storing spectral data files in the fwl_data repository?

lsoucasse (commented):
Looks good overall though I cannot really assess the sulfur chemistry part. I recommend not to update the Mors module for now because it is not used.

input/janus_mixed.cfg
nichollsh: I am not sure what the best thing is to do with that. The current disk size of AGNI with all the default files is now about 600 MB, which is too large.

I think storing all of the spectral files in a single place makes the most sense, so maybe writing a simple wrapper for the OSF repo in AGNI would be a good idea.

lsoucasse: On the Proteus side, it makes sense to have the data stored once in a place that both AGNI and Janus share access.

lsoucasse (approved):

nichollsh:
Thanks!

PR #107, opened 2024-07-15
SciColourMaps from package

### LINKED ISSUE #96: Switch to colourmaps package
We currently use the Scientific Colormaps, which are good for presenting data objectively and in an accessible manner.

These files are packaged in PROTEUS/JANUS directly at the moment, but it would make more sense to simply use the python package that contains them.

https://pypi.org/project/cmcrameri/

nichollsh:
Now accessing ScientificColorMaps from the conda package, rather than storing them inside the repository directly.

Closes #96.


lsoucasse (approved):
Great work! Lightens the proteus sources.

PR #108, opened 2024-07-18
Baraffe

lsoucasse:
This work affects the stellar heating. It now always points to Mors to get the stellar flux using either Spada data (and the rotation model of original Mors) or the Baraffe data (without rotation).

It removes the "legacy method" (which actually used Baraffe track data).

Input track data are downloaded automatically from the OSF repository (if needed) when running the main script.


nichollsh (approved):
Looks great. You said in the MORS PR that it works when running PROTEUS, so I assume that that tested this PR too.

Also nice to see a move towards fewer passes of COUPLER_options.

lsoucasse:
> Looks great. You said in the MORS PR that it works when running PROTEUS, so I assume that that tested this PR too.
> 
> Also nice to see a move towards fewer passes of COUPLER_options.

Thanks @nichollsh, the main proteus script broke because of this sulfur issue but it went to all new Mors call (setup, dowload data, getting stellar flux) before entering Janus with consistent steallar flux values between Spada and Baraffe data. A more comprehensive test should be done though with updated Janus.

nichollsh:
That makes sense to me. I'll get the sulfur branch ready.

PR #111, opened 2024-07-22
Detangling input/output variables

### LINKED ISSUE #95: Rework runtime model data storage
The way PROTEUS currently stores information is messy, being split between COUPLER_options and runtime_helpfile. This poses a significant risk of introducing unexpected behaviours, and is also decreasing performance. It also makes adding additional modules difficult. This is also the main difficulty in allowing the simulations to be resumed (issue #90).

We should consider storing model variables in a [dataclass](https://www.pythontutorial.net/python-oop/python-dataclass/). This would allow the information to be passed around more easily and transparently. These data can easily be converted to a dictionary and then written to a CSV file.

Dan is already doing something similar with atmodeller. JANUS and AGNI also store model information in a struct-like entity. It would make sense if PROTEUS did the same.

@lsoucasse and @timlichtenberg, what are your thoughts on this?

This is connected to issues #74, #94, #82, #76, #87, #70.

nichollsh:
This PR makes a significant step towards reworking how PROTEUS handles variables. With this version, output variables are stored as a Pandas DataFrame (the helpfile, HF). However, this HF is no longer passed around to as many functions as before, and is never modified by them directly. Calls to submodules (e.g. SPIDER, JANUS) only return the variables which they have calculated, which are then used to update the current working row of the HF. Once an iteration of the model is completed, the new row is committed to the HF and the whole DataFrame is saved to a CSV file. 

Similarly, the COUPLER_options dictionary now exclusively stores model input variables. These are never modified at runtime and are not returned from any functions other than when the configuration file is first read-in. 

The model runs and reproduces the previous results, which is expected given that these modifications are not intended to modify any physics. However, the output columns of the helpfile have different names in some cases, so this may break analysis or post-processing scripts written externally to the code itself.

Closes #95.

Other changes:
- Rework plot_atmosphere function
- Rework plot_fluxes functions 
- Can now change plotting output format (pdf, png, etc) via the configuration file
- Calls to plotting scripts from the terminal have a consistent command line interface
- More documentation inside the code
- Removed some redundant library imports
- Modifications to AGNI wrapper
- Updated vscode settings for workspace
- Optimisations to surface_gases
- Moved default configuration (previously called `init_coupler.cfg`) to `input/default.cfg`
- Some efforts towards a consistent code style, with line lengths of <=92 chars and docstrings.


nichollsh:
Next step after this will be to merge my resume branch, which includes functionality for resuming the simulation from the disk. https://github.com/FormingWorlds/PROTEUS/tree/resume

PROTEUS.env
```diff
@@ -18,8 +18,9 @@ export PATH=$COUPLER_DIR:$PATH
 export PYTHONPATH=$COUPLER_DIR:$PYTHONPATH
 
 # Source SOCRATES if found in SOCRATES directory
-if [ "$(ls -A $COUPLER_DIR/SOCRATES)" ]; then
-    cd $COUPLER_DIR/SOCRATES
+SOC_DIR="$COUPLER_DIR/SOCRATES/"
```
timlichtenberg, line 21: Should we replace all references to "coupler" with "proteus" to make it all consistent? Perhaps in a later commit.

nichollsh: Yeah that's a good idea. In practice this might be a simple find+replace job, but it should maybe be left for a future commit.

utils/vulcan.py
```diff
@@ -4,8 +4,12 @@
 from utils.modules_ext import *
 from utils.helper import *
 
-# run VULCAN/atmosphere chemistry
 def RunVULCAN( atm, time_dict, loop_counter, dirs, runtime_helpfile, COUPLER_options ):
```
timlichtenberg, line 7: If a different functionality is used for VULCAN, delete this function?

timlichtenberg (approved):
Great start with the disentanglement! Many different changes though, so hard to see if there is any minor bugs in practice.

utils/vulcan.py
nichollsh: The current functionality for VULCAN is the offline chemistry stuff. There's no way to run it 'online' at the moment, although this should probably be addressed by issue #6 in the future.

nichollsh:
Thanks Tim! I agree it's hard to tell through all of the changes but it does seem to reproduce the same results as before for the pure-steam Earth example case.

We should think about adding some tests.

PR #112, opened 2024-07-23
Resume simulation

### LINKED ISSUE #90: Resume PROTEUS simulation
Apparently, it is not possible to resume a PROTEUS simulation form an intermediate result. This should be fixed to incorporate tests (see #86 ) that only run for few coupled iterations from a given known state.

nichollsh:
With these changes, PROTEUS is able to resume a simulation from a previous state. This is done by passing the command line flag `--resume`. This will load the previous variables and continue from the last saved point in the output folder, using the configuration file provided by `--cfg`.

This also results in fewer dictionary passes to functions (e.g. time_dict stays within the main loop) which should help to modularise things.

I have tested this under several conditions, and it seems to work well. The implementation is not very complex, thankfully. I hope that this will also make implementing some tests easier. Closes #90.

Other changes:
- Removed unused configuration parameters
- Updated documentation and other formatting changes
- Updated example outputs 
- Renamed `utils/logging.py` to `utils/logs.py` to avoid confusing the the `logging` library


input/jgr_grid.cfg
```diff
@@ -151,7 +145,7 @@ Phi_global                  = 1.0       # Mantle melt fraction initial guess
 CH_ratio                    = 1.0       # C/H ratio
 hydrogen_earth_oceans       = 6.0       # Hydrogen inventory in units of equivalent Earth oceans
 nitrogen_ppmw               = 2.0       # Nitrogen inventory in ppmw relative to mantle mass
-sulfur_ppmw                 = 200.0     # Nitrogen inventory in ppmw relative to mass of melt
+sulfur_ppmw                 = 200.0     # Sulfur inventory in ppmw relative to mass of melt
```
timlichtenberg, line 148: This should be 0, or do you want to repeat the grid for the revision, including S?

utils/agni.py
```diff
@@ -87,27 +88,28 @@ def _try_agni(loop_counter:dict, dirs:dict, COUPLER_options:dict,
...
-        for k in vol_dict.keys():
-            if vol_dict[k] < vmr_min:
-                vmr_min = vol_dict[k]
-                gas_min = k
-        # add all gases as condensates, except the least abundant gas 
-        for k in vol_dict.keys():
-            if k == gas_min:
-                continue 
-            condensates.append(k)
+    # condensates = []
```
timlichtenberg, line 91: why only water? only for now?

utils/vulcan.py
```diff
@@ -4,15 +4,14 @@
 from utils.modules_ext import *
 from utils.helper import *
 
-def RunVULCAN( atm, time_dict, loop_counter, dirs, runtime_helpfile, COUPLER_options ):
+def RunVULCAN( atm, time, loop_counter, dirs, runtime_helpfile, COUPLER_options ):
```
timlichtenberg, line 7: About this simplification of the time dict: is it still possible to define a simulation relative to star formation? i.e., can one set up a simulation at 150 Myr after the formation of the Sun? Will this be displayede in the plots then as starting from 0, or starting from 150 My onwards?

timlichtenberg (approved):
S update ok

utils/vulcan.py
nichollsh: The stellar age is now also stored in the helpfile, so this should be relatively easy to do now, yes. It would simply be a case of plotting values vs `age_stellar` rather than versus `Time`.

PR #114, opened 2024-07-24
Some observational quantities

### LINKED ISSUE #113: Observational metrics
It would be interesting to calculate measures of transit depth and contrast ratio over time, so that we can learn about what these planets might look like at different stages of their evolution.

Calculating transit depth could be done simply by taking the height at 1 mbar and dividing this by the stellar radius. This would be only a bolometric transit depth, and assumes that tau=2/3 at the 1mbar level.

Contrast ratio could be calculated as the ratio of total TOA flux to the stellar emission, scaled to an appropriate distance.

nichollsh:
The model now calculates bolometric transit depth and planet-star contrast ratio, as some simple observable quantities. These are stored at each timestep, and also plotted to a file. Closes #113.

Updated examples to show this. Also fixed a minor library import issue.


timlichtenberg (approved):
Looks good!

nichollsh:
Thanks Tim!

PR #115, opened 2024-07-25
Dummy escape module

nichollsh:
This PR adds basic escape infrastructure, where total elemental mass inventories are updated according to a given escape rate. The escape model and escape rate are specified in the configuration file.

There's also a new break condition, where the model will terminate when the atmosphere mass drops below some small fraction of the initial mass. This represents the volatiles having (almost) entirely escaped. There is a new plot function `plot_elements` to show the mass loss over time.

There's also a stub function for ZEPHYRUS integration, when it's ready.


timlichtenberg (approved):
Looks good to me, should nicely speed up the escape implementation!

PR #117, opened 2024-07-31
Fixed installation instructions : underscore in the step 4 command

EmmaPostolec:
Hi all, I noticed a small thing in the online instructions at step 4 SOCRATES:

```$ ./build-code``` does not work but the correct command is 

```$ ./build_code```

The only difference is the underscore sign _ instead of the minus sign - between 'build' and 'code'. 

I just modified the docs>source>installation.rst file to correct this.

### DIFF
--- a/docs/source/installation.rst
+++ b/docs/source/installation.rst
@@ -81,7 +81,7 @@ Download the framework
 
         $   cd SOCRATES
         $   ./configure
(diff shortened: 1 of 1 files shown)


timlichtenberg (approved):
Looks good, nice! You can now merge the code into master.

PR #123, opened 2024-08-05
Rewrite AGNI interface with JuliaCall and other refactoring

### LINKED ISSUE #118: JuliaCall with AGNI
Currently, AGNI is called as a subprocess through its command line interface. This means that each call to the model needs to re-allocate memory and load data from the disk.

It would be more efficient/faster to use JuliaCall to access the model, where the data can be held in memory by PROTEUS for the entire duration of the simulation.

https://juliapy.github.io/PythonCall.jl/stable/juliacall/

nichollsh:
- The AGNI interface `utils/agni.py` is now operating via [JuliaCall](https://juliapy.github.io/PythonCall.jl/stable/juliacall/) rather than accessing AGNI is a subprocess. This is more efficient and improves model accuracy by avoiding disk reading/writing. Closes #118.
- Updated configs.
- Refactored `COUPLER_options` variables to just `OPTIONS`, since this is more concise and the model name is no longer just "COUPLER".
- Support for running equilibrium chemistry in the atmosphere via AGNI.
- Updated examples (added a script for doing this more consistently in `tools/MakeExample.sh`)

Future work should consider moving some of the content of `agni.py` into the AGNI repository.


tools/MakeExample.sh
timlichtenberg, line 1: This is for generating example test cases I presume?

nichollsh: Yeah, I figured it might be useful to standardise this.

timlichtenberg (approved):
I am especially fond of the name changes wrt to "coupler".

PR #125, opened 2024-08-06
Set up pyproject.toml

### LINKED ISSUE #106: pyproject.toml
The current documentation for PROTEUS has information on which packages need to be installed. This could be made easier by having a `requirements.txt` file. We could then simply run: `conda install --file requirements.txt` to get all of the packages, rather than having the user run a specific command.

stefsmeets:
This PR sets up pyproject.toml for installing the code.

The code can be installed with `pip install -e .` For the moment this just installed the dependencies.

I was tentative to add MORS/JANUS to the dependencies, because it is not clear to me yet how they can be integrated. We can pick that up at a later stage.

I also set up the boilerplate for a Python package (`src/proteus`) where we can move shared code too.

Closes #106

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -17,11 +17,9 @@ mkdocs:
 
 # Optionally declare the Python requirements required to build your docs
 python:
-  install:
-  - requirements: docs/requirements.txt
-# python:
-#    install:
-#    - method: pip
-#      path: .
-#      extra_requirements:
-#       - docs
+python:
+   install:
+   - method: pip
+     path: .
+     extra_requirements:
+      - docs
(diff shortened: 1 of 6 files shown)


stefsmeets:
@timlichtenberg

stefsmeets:
@nichollsh This PR is ready, could you have a look at it?

nichollsh (changes requested):
Looks good as a method for installing dependencies. However, I think it is missing some:
- scipy
- ipykernel
- cmcrameri
- juliacall

Additionally, sympy and natsort are listed in the docs, but I think are not actually used in the code anywhere.

stefsmeets:
@nichollsh Ah, thanks for pointing that out. I overlooked those. I cannot find ipykernel being used anywhere, but I will add the others. I also updated the documentation.

nichollsh (approved):

PR #133, opened 2024-08-09
Atmos wrapper

lsoucasse:
This PR reorganizes the Proteus repository and introduces a wrapper function to access the atmosphere submodules (JANUS/AGNI/dummy):
- All atmosphere-related scripts are now in the `src/proteus/atmosphere` folder.
- The wrapper script `wrapper_atmosphere.py` is for now very basic: it only copies previous atmosphere sub-loop code lines into a RunAtmosphere function. We might want to improve it through an AtmosphereModel class for instance. Also we might want to change the module import such that it only does the import of the submodule it needs.
- The `utils/` and `plots/` directories are moved into `src/proteus/`. It means that the `pip install -e .` command needs to be run to access associated scripts. The `tools` folder remains in the root as it gathers notebooks and pre/post-process scripts.

Everything seems OK on the JANUS side but I didn't check everything on the AGNI side. Could you try this @nichollsh?


pyproject.toml
```diff
@@ -4,13 +4,23 @@ build-backend = "setuptools.build_meta"
...
+	{name = "Harrison Nicholls, email = harrison.nicholls@physics.ox.ac.uk"},
+	{name = "Laurent Soucasse", email = "l.soucasse@esciencecenter.nl"},
+	{name = "Stef Smeets", email = "s.smeets@esciencecenter.nl"},
+	{name = "Dan J. Bower", email = "dbower@ethz.ch"},
+	{name = "Mark Hammond", email = "mark.hammond@physics.ox.ac.uk"},
+	{name = "Patrick Sanan", email = "patrick.sanan@gmail.com"},
+	{name = "Raymond Pierrehumbert", email = "raymond.pierrehumbert@physics.ox.ac.uk"},
+	{name = "Ryan Boukrouche", email = "ryan.boukrouche@astro.su.se"},
+	{name = "Shang-Min Tsai", email = "shangmin.tsai@ucr.edu"},
+	{name = "Hamish Innes", email = "hamish.innes@fu-berlin.de"},
```
timlichtenberg, line 23: Can we add Mariana and Emma here?

src/proteus/utils/vulcan.py
timlichtenberg, line 1: This should also go /proteus/atmosphere/

src/proteus/utils/zephyrus.py
timlichtenberg, line 1: This should also go to /proteus/atmosphere/

nichollsh: Surely this will go under `src/proteus/escape`, when that folder is created?

src/proteus/utils/vulcan.py
nichollsh: Maybe we should distinguish between atmosphere temperature (climate) solvers and atmosphere chemistry? Could have an `src/proteus/atmos_clim` and `src/proteus/atmos_chem` or something like that?

src/proteus/utils/zephyrus.py
lsoucasse: Harrison is right, it should go in `src/proteus/escape`. I did not create all the subdirectories (interior, escape, outgassing). I think it is better to proceed step by step.

timlichtenberg: Ok

src/proteus/utils/vulcan.py
timlichtenberg: Not sure this may overcomplicate it, but I have no strong opinions

lsoucasse: My first thought was that `vulcan.py` would go into `src/proteus/chemistry`. Will you have chemistry model for the interior in the future? In that case it can be ambiguous and Harrison's suggestion is probably better.

timlichtenberg: Yes, in fact, that's the idea. SPIDER/aragog would be the interior equivalent of Janus/Agni+SOCRATES, modeling the physics of energy transport mostly. Then we would have an interior chemistry module(s) that are the equivalent of VULCAN on the atmosphere side.

My idea was to simply put all interior modules into "Interior", and the same with the atmospheric things. If you think it makes sense to further differentiate, however, please do so.

pyproject.toml
lsoucasse: Done!

src/proteus/utils/vulcan.py
lsoucasse: Hum... I think putting everything into `atmosphere` is confusing because if you take for instance Janus and Vulcan, the two modules are not interchangeable but are potentially coupled because they handle different physics. I suggest we follow Harrison's suggestion, is that OK @timlichtenberg?

timlichtenberg: Fine for me.

lsoucasse: Modification done through last commit.

start_proteus.py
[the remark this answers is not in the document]
lsoucasse: I think it is because you have the latest version of Mors installed in your computer. The Mors submodule of Proteus points to commit [3c795f3](https://github.com/FormingWorlds/MORS/commit/3c795f3a2dec683ddc1044ed14fdff245007fce8) and for this commit it works with `/Spada` and `/Baraffe`.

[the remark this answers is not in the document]
lsoucasse: We will solve these kind of issues when we remove the submodules #135.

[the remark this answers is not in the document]
nichollsh: Okay, sure!

nichollsh (approved):
Looks good. The model runs a few iterations on my machine without any errors, although I didn't let it simulate until convergence.

Issue #74, opened 2024-02-27
Rework configuration file


nichollsh:
The current configuration file format is very confusing. It's not clear which parameters are necessary, and how they relate to each other. I've tried to make this clearer in the ReadTheDocs page using a table, but that's not good enough.

Since we want to keep things as "modules", which are swappable, I propose that we use a TOML file for the cfg format with a section for each module being used.

For example...

```
# Example cfg file with some arbitrary values

[runtime]
log_output =  true
output_folder = "path/to/folder"

[planet]
mass = 10.0
radius = 2.0
surface_bc = 2

[star]
mass = 1e5
radius = 1e4
spectrum = "path/to/file"

[modules]

    [modules.interior]
    enabled    = true
    selection  = "C_SPIDER"
    melt_curve = "path/to/file"

    [modules.atmosphere]
    enabled   = true
    selection = "JANUS"
   
        [modules.atmosphere.initial]
        use_params = false 
        H2O_bar    = 30.0

    [modules.escape]
    enabled   = false

```

See more info here: https://toml.io/en/
This connects with #62, #68, #66, #56, and #71.

nichollsh:
One thing that might also be useful: having multiple ways to specify the planetary parameters. It would be handy to be able to say `planet=Earth` or `planet=TRAPPIST1b` rather than specifying mass, radius, etc, directly.

stefsmeets:
+1 on using toml for the config format. I think striving for a model like in the [JANUS config](https://github.com/FormingWorlds/JANUS/blob/master/src/janus/data/tests/config_instellation.toml) is a very good idea. 

1. Move from .cfg to .toml
2. Add [tables](https://toml.io/en/v1.0.0#table) for planet, star, etc
3. Add validation

I also want to move away from storing the data in a dictionary towards a more structured representation of the config with validation, probably in the form of a dataclass.

For validation then, I think we have 3 options:

1. Python dataclasses with a big `__post_init__` (most similar to what [we have now](https://github.com/FormingWorlds/PROTEUS/blob/bc6a55d5b127f40b8ad90656657485d12e032533/src/proteus/utils/coupler.py#L314)). Simple, not as powerful as attrs.
2. [`attrs`](https://www.attrs.org/en/stable/overview.html) library, sort of a middle-ground between pydantic and dataclasses. Has some [validation](https://www.attrs.org/en/stable/api.html#module-attrs.validators), but it is optional.
[...]

PR #142, opened 2024-08-20
Refactor start_proteus.py and move config to toml

### LINKED ISSUE #141: Refactor `start_proteus.py`
As discussed at yesterdays meeting, I'm opening this issue to track the progress with refactoring `start_proteus.py`.

The goal is to move all the logic etc into the library so that it can be re-used elsewhere. The goal is to have 3 entry points:

1. `start_proteus.py`, this will work in the same way as before. It will just be a lot smaller. This can be used to specify config variables, paths etc, but the main code will be elsewhere.
2. `proteus start`, this will be a CLI tool that starts proteus. The advantage is that it can be started from anywhere and does not require the user to git clone the repo. This will be more relevant as we move away from the submodules.
3. Python api, e.g. in a jupyter notebook. This is somewhat similar to 1. This will enable more exploratory use of the code, e.g. by importing the proteus class into your code: 

```python
from proteus import Proteus

handler = Proteus(cfg='input/default.cfg')
handler.start()
```

stefsmeets:
This PR refactors the `start_proteus.py`. This includes an update to the CLI, in line with what we did for janus / mors.

```console
proteus start --config input/default.toml
```

or

```console
python start_proteus.py
```

I also started to move the config to toml. Moving to toml was fairly straightforward and only meant minor changes in the cfg files itself. Many parts of the code read the config directly, which is turned out to be more work than I had anticipated.

Next steps would be to break down the `Proteus.start()` function into isolated methods, but I want to tackle that in a follow-up PR to avoid making too many changes at once.

## TODO
- [x] Update `plot/cpl_*` files to use toml
- [x] Update `utils/plot_offchem.py` to use toml
- [x] Add tests
- [x] Fix CI workflow

Closes #141 
Addresses #74


stefsmeets:
Hi @lsoucasse @nichollsh Could you please review this PR and let me know what you think?

The main files to look at would be `src/proteus/proteus.py`, `src/proteus/cli.py`, `src/proteus/utils/coupler.py`.

Because of the toml change, I also had to update all the code that reads in the config directly, see `src/proteus/plot`, and specifically `_cpl_helpers.py` for how I refactored it.

Happy to talk you through the PR if necessary.

nichollsh:
Looks good. These changes make sense to me, and it's nice to have a consistent CLI for PROTEUS.

However, doesn't this make `start_proteus.py` redundant? Why include it at all if there's already the `click` interface for running the model via `proteus start`?

docs/usage.md
```diff
@@ -5,12 +5,12 @@
...
 ```console
-python start_proteus.py --cfg [cfgfile]
+proteus start --config [cfgfile]
 ```
 
 Where `[cfgfile]` is the path to the required configuration file. If
-`--cfg [cfgfile]` is not provided, then the default configuration
-located at `input/default.cfg` will be used. Pass the flag `--resume` in
+`--config [cfgfile]` is not provided, then the default configuration
+located at `input/default.toml` will be used. Pass the flag `--resume` in
```
lsoucasse, line 13: The config file is always expected by the cli function.

lsoucasse (changes requested):
Thanks @stefsmeets, Nice to switch to toml and to have the cli for running proteus. I agree for now the start_proteus.py script is redundant but once we split the big running function into different methods it might be worth having it.
I tested all the commands, it seems to run fine. Just need to update the documentation because the path to config file is mandatory for the cli.

stefsmeets:
Cheers both, agree on the `start_proteus.py`. I kept it, because it signals how to call the code from Python. And it is part of the current workflow for some people. I think it does not hurt to keep it around for a while, but if you both think we can remove it we can also remove it.

lsoucasse:
> Cheers both, agree on the `start_proteus.py`. I kept it, because it signals how to call the code from Python. And it is part of the current workflow for some people. I think it does not hurt to keep it around for a while, but if you both think we can remove it we can also remove it.

I think we can keep it for now.

stefsmeets:
@nichollsh @lsoucasse Could you check again and approve? This PR is ready to be merged.

lsoucasse (approved):
Good for me.

PR #149, opened 2024-08-26
Add integration test for PROTEUS

stefsmeets:
This PR adds an integration test for PROTEUS using the dummy config. This bumps the coverage to about 60%. It takes about 40 seconds on my PC.

I tried the default config (with JANUS), but this is extremely slow (over 6 min) even with `max_iter=0`. I think it would be better to test this module and others separately, or once we have a better way to control the number of iterations.

### TODO
- [x] Fix CI workflow

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -26,7 +26,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
         with:
-          submodules: 'true'
+          submodules: 'false'
 
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v4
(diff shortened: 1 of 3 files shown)


stefsmeets:
I tried to get the integration test to work on the CI, but the installation procedure is too complicated, because several tools need to be compiled from scratch (most notably PETSC with SUNDIALS support for SPIDER). This already took close to 30 minutes. This makes the workflow incredibly painful to set up, debug and maintain, so I chose to skip it. The test runs fine locally.

stefsmeets:
@nichollsh would you care to have a look at this PR and let me know what you think?

nichollsh:
I don't understand why the model takes >6 minutes with JANUS, but this is something to be investigated separately. If it works fine with the dummy config then that's good for now.

lsoucasse:
Very nice to have this test set up! I just notice that the workflow install Mors and Janus from current master branch status (without using it though) while we rather want to install a specific version. But this will be fixed soon when we install them through the package dependencies after making the appropriate releases.

stefsmeets:
@lsoucasse Exactly! And pulling all the submodules takes a while, whereas mors and janus are fairly light.

PR #150, opened 2024-08-26
Janus wrapper

lsoucasse:
This pull request clarifies the Janus wrapper. It now better handles the initialisation, and the update of the atmospheric object, either because the iteration variables have changed or because the stellar spectrum has changed.


src/proteus/proteus.py
```diff
@@ -423,27 +420,24 @@ def start(self, *, resume: bool = False):
...
 
                 log.info("Updating stellar spectrum")
                 match self.config["star_model"]:
                     case 0:
                         synthetic = mors.synthesis.CalcScaledSpectrumFromProps(
                             star_struct_modern, star_props_modern, hf_row["age_star"] / 1e6
                         )
-                        fl = synthetic.fl  # at 1 AU
-                        wl = synthetic.wl
+                        hf_row["fl"] = synthetic.fl  # at 1 AU
```
nichollsh, line 431: I am not sure that this is a good idea. At the moment `hf_row` is only designed to store scalar quantities, since it is then appended to the end of the helpfile dataframe and CSV file. These `fl` and `wl` arrays are 1D, so it will not be possible to simply append them to the helpfile in this manner.

I would suggest keeping these variables standalone for now, and we can later think about handling them better.

nichollsh (changes requested):
Looks good except for how the wl and fl arrays are being handled. I would suggest leaving these as they are for now, and then we can think about how to handle these things (similarly for the `atm` objects).

src/proteus/proteus.py
lsoucasse: Wanted to lighten the call to the functions but I get your point. Fixed in new commit.

nichollsh (approved):

PR #151, opened 2024-08-27
Add plots to CLI

### LINKED ISSUE #143: About plotting functions in `src/proteus/plot/ `
Hi all, I was wondering about the plot functions in `src/proteus/plot/cpl_*.py`. I know that some of these get run in the Proteus main loop to produce the plots.

But all of these also have entry points (`if name == __main__`), which indicates to me that they are also supposed to be run directly. Could you tell me how this is this meant to work?

I was thinking we could fold those into the proteus CLI to make these funcions more accessible.

e.g.

```
proteus plot interior --config init_coupler.toml 
# or
proteus plot sflux --config init_coupler.toml 
```

And at the same time, by structuring this a bit more, we could expose it via a python api:

```python
from proteus import Proteus

handler = Proteus(config_path='init_coupler.toml')
handler.plot('interior', **kwargs)
# alternative
handler.plot_interior(**kwargs)
```

Let me know what you think.

stefsmeets:
This PR refactors the plots and exposes them through the CLI. The plots can also be invoked directly.

Example:

```console
$ proteus plot interior global --config init_coupler.toml
Plotting: global
Plotting: interior
```

which is the equivalent of:

```console
python src/plot/cpl_interior.py init_coupler.toml
```

List all plots:

```console
$ proteus plot --list
atmosphere atmosphere_cbar elements escape fluxes_atmosphere
fluxes_global global heatingrates interior interior_cbar
interior_cmesh observables offchem_grid_cross offchem_species
offchem_time offchem_year sflux sflux_cross stacked
```

Closes #143 

### Todo
- [x] Check if it is possible to add some tests
- [x] Test on real data

Tested:
- atmosphere
- elements
- escape
- fluxes_atmosphere
- fluxes_global
- global
- interior
- interior_cbar
- interior_cmesh
- observables
- sflux
- sflux_cross
- stacked
  
Not tested (data not available, needs vulcan run):
- atmosphere_cbar
- offchem_species
- offchem_time
- offchem_year

Not available through CLI yet:
- heatingrates
- offchem_grid_cross


stefsmeets:
For testing the plots, in the absence of a full dataset, I would like to follow-up with a sample output dataset to work with, e.g. from `proteus start --config inputs/default.toml`. The data can be hosted in a separate repo, and we can use the [matplotlib.testing](https://matplotlib.org/stable/api/testing_api.html#module-matplotlib.testing.compare) module to help regression test the outputs.

stefsmeets:
Hi @lsoucasse @nichollsh , this PR is ready. could you have a look at this PR and let me know what you think?

I changed all the plotting codes so that they have a consistent entry point (`plot_xxx_entry`), so that they take a `Proteus` object. There are no functional changes. All functions can still be invoked directly.

Other than that, most of the changes are of this kind:
- Remove dead code, separators
- Add types to plot functions
- Move the code from `if __name__ == '__main__'` block to entry.

nichollsh (approved):
Looks good. Makes sense to have a consistent way to interact with the code, both for plotting and running. Seems like the runtime plotting is mostly unchanged, but I agree it would be good to have some tests for it in the future.

PR #155, opened 2024-08-29
Installation and logging updates

### LINKED ISSUE #137: Update default logger instance name
For consistency with JANUS, the default logger name should be changed to "FWL" rather than "PROTEUS", and this name should be encouraged across all of the FWL programs currently under development.

This means that, when used together, these programs can use the same logger instance without having to do any ugly hacks.

### LINKED ISSUE #145: Moving to miniforge or venv
My IT department has recently discovered that some parts of the Anaconda software are no longer free as they have changed their licensing. This applies to the default conda channel, and therefore also to miniconda. Any organisation which has more that 200 employees, even universities and even for research need to pay for a licence which is in the region of $50 per person per month.

You can read about this here:
https://www.anaconda.com/blog/is-conda-free
https://legal.anaconda.com/policies/en/

It does not apply to conda-forge.
https://conda-forge.org/blog/2020/11/20/anaconda-tos/

I have therefore been asked to immediately discontinue all use of conda, and instead use miniforge or venv+pip. I don't know how this applies to other organisations and their specific cases.

Switching to miniforge is pretty simple and might only require modifying the documentation to nudge the user towards it instead of miniconda.

@stefsmeets @lsoucasse what do you think?

### LINKED ISSUE #152: Make VULCAN an optional module
Currently VULCAN is included in the core installation process. It would be better to make it an optional module like AGNI, to streamline things and potentially make room for alternative models.

nichollsh:
1. Updated SOCRATES submodule to source from https://github.com/nichollsh/SOCRATES as previously discussed. This will be the main (and development) repository for SOCRATES, and the FWL version can act as a 'backup'.

2. Removed VULCAN as submodule. Added docs for downloading it as an 'optional' module, like AGNI. Closes #152.

3. Updated docs to use venv+pip rather than conda as the recommended installation method. Closes #145.

4. Updated JANUS and MORS to latest versions.

5. All loggers are created with names in the format `"fwl."+__name__`, which closes #137. Logging handlers should be setup by executables to handle all `"fwl"` loggers, which will include PROTEUS and JANUS so far.


nichollsh:
@stefsmeets please could you explain to me why the code style action is failing? I think it is something to do with linting that it can't automatically correct?

nichollsh:
@stefsmeets sorry for the continued development after requesting your review! However, I think that this is now actually ready for review.

src/proteus/utils/logs.py
```diff
@@ -57,19 +57,22 @@ def format(self, record):
...
+    custom_logger = logging.getLogger(logger_name)
+    # custom_logger.handlers.clear()
 
     if os.path.exists(logpath):
         os.remove(logpath)
 
     level = str(level).strip().upper()
     if level not in ["INFO", "DEBUG", "ERROR", "WARNING"]:
-        level = "INFO"
+        raise Exception("Invalid log level '%s'"%level)
```
stefsmeets, line 75:
```suggestion
        raise ValueError(f"Invalid log level: {level}")
```

src/proteus/utils/logs.py
```diff
@@ -57,19 +57,22 @@ def format(self, record):
...
+
+    logger_name = "fwl"
 
     # https://stackoverflow.com/a/61457119
+    print("Setup logger as %s"%str(logger_name))
 
-    custom_logger = logging.getLogger("PROTEUS")
-    custom_logger.handlers.clear()
+    custom_logger = logging.getLogger(logger_name)
+    # custom_logger.handlers.clear()
```
stefsmeets, line 68:
```suggestion
```

src/proteus/utils/logs.py
```diff
@@ -57,19 +57,22 @@ def format(self, record):
         return formatter.format(record)
 
 # Custom logger instance
-def SetupLogger(logpath:str="new.log",level:str="INFO",logterm:bool=True):
+def setup_logger(logpath:str="new.log",level:str="INFO",logterm:bool=True):
+
+    logger_name = "fwl"
 
     # https://stackoverflow.com/a/61457119
+    print("Setup logger as %s"%str(logger_name))
```
stefsmeets, line 65: Is this a debug statment that you meant to remove?

```suggestion
    print(f"Setup logger as {logger_name}")
```

docs/installation.md
```diff
@@ -203,3 +182,30 @@ They are
...
+
+    ```console
+    make
+    cd ../../
+    ```
+3. Install Python dependencies
+
+    ```console
+    pip install sympy
+    ```
```
stefsmeets, line 211: Although I understand why this is needed, it would be better if Vulcan would declare its own dependencies, either through an `requirements.txt` or `pyproject.toml`. May be worth to contribute a patch to that code, so that it can be installed like mors or janus (e.g. `pip install vulcan/.`). That also avoids PROTEUS having to set the PYTHONPATH for Vulcan in `proteus.env`.

stefsmeets (approved):
Looks good, nice work! I left some small suggestions in the comments.

stefsmeets (approved):

Issue #135, opened 2024-08-09
Remove Janus, Mors and Socrates submodules


lsoucasse:
Now that the release workflow is set up for Janus and Mors, we can delete these submodules and add them to the dependency list. We need however to be version-specific in the pyproject.toml to point to the correct version, especially at this moment where things are quite changing.

For Socrates, both Janus and Agni provides download/install function. We can thus delete the submodule but need to harmonize access from Janus/Agni and solve #134 first.

timlichtenberg:
Before we do this: how does this change the developer installation path? Can you still have all submodules within the PROTEUS directory without cloning everything manually?

lsoucasse:
For a developer install, you can still clone the Mors or Janus repo anywhere in your computer (it does not have to be within the PROTEUS directory). Then you run `pip install -e .` in Mors or Janus and this will install your "own" local version.

stefsmeets:
`pyproject.toml` specifies the requirements on janus and mors, so if these are satisfied by a local installation like @lsoucasse suggested, then this will also work. If they are not found, `pip install .` in the proteus directory will pull them from pypi.

nichollsh:
How will SOCRATES be handled in this case, since it won't be picked up by the pip command?

PR #158, opened 2024-08-30
Remove submodules

lsoucasse:
This PR removes Janus and Mors as submodules and install them from PyPI instead as dependencies of the Proteus package. The version installed is specified in the pyproject.toml file.
Socrates is also deleted as a submodule. I used Harrison's script to get it from github and compile it. It thinks it is sufficient for now.

### DIFF
--- a/.github/workflows/code-style.yaml
+++ b/.github/workflows/code-style.yaml
@@ -3,10 +3,10 @@ name: Code style
 on:
   push:
     branches:
-      - master
+      - main
   pull_request:
     branches:
-      - master
+      - main
     types:
       - opened
       - reopened
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -3,10 +3,10 @@ name: Tests for PROTEUS
 on:
   push:
     branches:
-      - master
+      - main
   pull_request:
     branches:
-      - master
+      - main
     types:
       - opened
       - reopened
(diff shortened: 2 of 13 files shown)


nichollsh:
@lsoucasse please can you clarify for me what happens in the developer case?

If you clone JANUS and run `pip install -e .` on that package in your environment. Does this still satisfy the dependency in the PROTEUS `pyproject.toml` file? Or will it overwrite your local JANUS and install the version from pypi?

lsoucasse:
> @lsoucasse please can you clarify for me what happens in the developer case?
> 
> Say you clone JANUS and run `pip install -e .` on that package in your environment. Does this still satisfy the dependency in the PROTEUS `pyproject.toml` file? Or will it overwrite your local JANUS and install the version from pypi?
> 
> Stef clarified a similar question previously ([#135 (comment)](https://github.com/FormingWorlds/PROTEUS/issues/135#issuecomment-2283425566)) but how does this change now that a particular release is specified in the pyproject file?



> @lsoucasse please can you clarify for me what happens in the developer case?
> 
> Say you clone JANUS and run `pip install -e .` on that package in your environment. Does this still satisfy the dependency in the PROTEUS `pyproject.toml` file? Or will it overwrite your local JANUS and install the version from pypi?
> 
> Stef clarified a similar question previously ([#135 (comment)](https://github.com/FormingWorlds/PROTEUS/issues/135#issuecomment-2283425566)) but how does this change now that a particular release is specified in the pyproject file?

Hi @nichollsh,
In this case I think what happens is that it will overwrite the PyPI install and proteus requirements will still be satisfied as long as the version number has not changed.
With changing version number we have to experiment, but you will always have the option to do pip uninstall/install locally to make sure which version you are actually using.

nichollsh:
Okay, makes sense! Thanks

nichollsh (approved):
Looks good!

PR #161, opened 2024-08-30
Emission spectrum calculation and other minor changes

### LINKED ISSUE #154: Emission spectrum calculation
It would be a useful metric to have an easy way to generate emission spectra. This could be done on the fly or post-processed with a high resolution spectral file, or both.

It should be simple enough to do this, now that JANUS and AGNI are outputting NetCDF files with the same variables and formatting.

### LINKED ISSUE #160: Editable dependencies broken?
Now that JANUS and MORS have been removed as submodules of PROTEUS, they are automatically installed as dependencies from pypi. This is useful for non-developers.

However, a developer will want an editable install which makes a symbolic link from the environment directory to wherever the modules have been cloned to. 

Installing JANUS (or MORS) with `pip install -e .` initially works as expected. However, if you run the `pip install` step for PROTEUS' `pyproject.toml`, it will overwrite the JANUS and MORS installations with ones from PyPI. This makes them no longer editable.

Additionally, it is not trivial to then reinstall the editable versions. Simply re-running `pip install -e .` in the MORS and JANUS folders does not replace the PyPI versions. A workaround is to manually delete MORS and JANUS from the `site-packages` directory, and then run the editable install again.

Overall, this is pretty clunky. @lsoucasse and @stefsmeets do you see the same behaviour?

nichollsh:
- New functionality for plotting emission spectrum (closes #154)
- Automatically plots emission at model exit/convergence
- Clarifies developer installation instructions (closes #160)
- Updated gitignore for when modules are cloned inside PROTEUS directory
- Minor improvements to AGNI wrapper, with some refactoring to fit code style
- Checks that SPIDER executable exists before trying to run it

Developer installation instructions probably need further work but this at least explains the difference between the editable and 'normal' install steps.

### DIFF
--- a/src/proteus/atmos_clim/agni.py
+++ b/src/proteus/atmos_clim/agni.py
@@ -7,6 +7,7 @@ import os
 
 import numpy as np
 from juliacall import Main as jl
+from scipy.interpolate import PchipInterpolator
 
 from proteus.utils.constants import dirs, volatile_species
 from proteus.utils.helper import UpdateStatusfile, create_tmp_folder, find_nearest, safe_rm
@@ -17,7 +18,7 @@ log = logging.getLogger("fwl."+__name__)
 # Constant
 AGNI_LOGFILE_NAME="agni_recent.log"
 
-def SyncLogfiles(outdir:str):
+def sync_log_files(outdir:str):
     # Logfile paths
     agni_logpath = os.path.join(outdir, AGNI_LOGFILE_NAME)
     logpath = GetLogfilePath(outdir, GetCurrentLogfileIndex(outdir))
@@ -38,7 +39,7 @@ def SyncLogfiles(outdir:str):
     with open(agni_logpath, "w") as hdl:
         hdl.write("")
 
-def ActivateEnv(dirs:dict):
+def activate_julia(dirs:dict):
 
     log.debug("Activating Julia environment")
     jl.seval("using Pkg")
@@ -62,7 +63,7 @@ def ActivateEnv(dirs:dict):
     log.debug("AGNI will log to '%s'"%logpath)
 
 
-def ConstructVolDict(hf_row:dict, OPTIONS:dict):
+def _construct_voldict(hf_row:dict, OPTIONS:dict):
     vol_dict = {}
     for vol in volatile_species:
         if OPTIONS[vol+"_included"]:
@@ -77,7 +78,7 @@ def ConstructVolDict(hf_row:dict, OPTIONS:dict):
     return vol_dict
 
 
-def InitAtmos(dirs:dict, OPTIONS:dict, hf_row:dict):
+def init_agni_atmos(dirs:dict, OPTIONS:dict, hf_row:dict):
     """Initialise atmosphere struct for use by AGNI.
 
     Does not set the temperature profile.
@@ -119,7 +120,7 @@ def InitAtmos(dirs:dict, OPTIONS:dict, hf_row:dict):
         input_star =    sflux_path
 
     # composition
-    vol_dict = ConstructVolDict(hf_row, OPTIONS)
+    vol_dict = _construct_voldict(hf_row, OPTIONS)
 
     # set condensation
     condensates = []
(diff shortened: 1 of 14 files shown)


src/proteus/atmos_clim/agni.py
```diff
@@ -328,16 +361,16 @@ def RunAGNI(atmos, loops_total:int, dirs:dict, OPTIONS:dict, hf_row:dict):
                             conduct=False, convect=True, latent=True, sens_heat=True,
 
                             max_steps=130, max_runtime=900.0,
-                            conv_atol=1e-3, conv_rtol=1e-2,
+                            conv_atol=1e-3, conv_rtol=2e-2,
```
timlichtenberg, line 364: Should these solver tolerances be hardcoded or rather go into the config file?

src/proteus/plot/cpl_emission.py
timlichtenberg, line 1: Do you have examples of this plot type?

src/proteus/atmos_clim/agni.py
nichollsh: It should do, yeah. I think it could go in the `[atmosphere]` section of the TOML configuration file once we have reformatted it in that way (issue #74).

tools/GridPROTEUS.py
```diff
@@ -346,19 +350,18 @@ def run(self,num_threads:int,test_run:bool=False):
 
          # Thread targget
```
timlichtenberg, line 363: typo

src/proteus/atmos_clim/agni.py
nichollsh: Or even in a template configuration file of some kind

tools/GridPROTEUS.py
```diff
@@ -473,8 +476,9 @@ def _thread_target(cfg_path):
     # Define parameter grid
     # -----
 
-    cfg_base = os.path.join(os.getenv('PROTEUS_DIR'),"input","t1c.cfg")
-    symlink  = "/network/group/aopp/planetary/RTP035_NICHOLLS_PROTEUS/outputs/t1c_v3"
+    cfg_base = os.path.join(os.getenv('PROTEUS_DIR'),"input","t1c.toml")
+    # symlink  = "/network/group/aopp/planetary/RTP035_NICHOLLS_PROTEUS/outputs/t1c_v3"
+    symlink = "/dataserver/users/formingworlds/nicholls/model_outputs/t1c_v4"
```
timlichtenberg, line 493: Should this be a bit more generic in `main`?

src/proteus/atmos_clim/agni.py
timlichtenberg: Ok, let's leave them here for now then.

src/proteus/plot/cpl_emission.py
nichollsh: The colour scale looks a little odd because it's trying to plot with log-time mapping, and the evolution is quite short (because I stopped it).

timlichtenberg (changes requested):
Some minor changes needed, indicated in inline comments.

src/proteus/plot/cpl_emission.py
timlichtenberg: Great! So this is the on-the-fly functionality, but does not yet include a high-res re-run + plot?

tools/GridPROTEUS.py
nichollsh: I agree. Made an issue for this https://github.com/FormingWorlds/PROTEUS/issues/162

src/proteus/atmos_clim/agni.py
nichollsh: See https://github.com/FormingWorlds/PROTEUS/issues/162

nichollsh:
Thanks!

src/proteus/plot/cpl_emission.py
nichollsh: This is currently configured to run at model termination with the runtime spectral file, hence the low resolution. 

Maybe we could add a tool which post-processes all simulated TP and Compositions from a single run with AGNI using a high resolution spectral file.

PR #170, opened 2024-09-04
Fixes to environment in files and docs

### LINKED ISSUE #169: Set SOCRATES environment using JANUS function
The user should only be required to set `RAD_DIR` for the model to be able to find SOCRATES. However, the current configuration requires the user to add it the `$RAD_DIR/bin` directory to their `PATH`.

This can be best resolved by making sure that the file `janus.set_socrates_env` is imported by PROTEUS before it attempts to do any radiative transfer calculations through JANUS.

AGNI is not impacted by this issue.

nichollsh:
Closes #169.
Other minor changes.


lsoucasse (approved):
Looks good. Approve it.

PR #179, opened 2024-09-13
High-resolution postprocessing and download data

### LINKED ISSUE #178: Download albedo data for AGNI
AGNI supports spectrally resolved surface albedo and emissivity calculations.
These data need to be downloaded from the OSF.

nichollsh:
* Significant improvements to postprocessing scripts, which allow the user to recalculate the radiative fluxes from a PROTEUS using a high resolution spectral file. This can now be applied to a grid of models more easily, with the grid points run in parallel.
* Added common telescope bandpasses to `plot.py` for easy comparison between model results and telescope instruments
* Model calculates the bulk density that would be observed in transmission, at runtime. Stored in the helpfile alongside other variables.
* Safety checks for when hydrostatic integration fails with JANUS. Does not attempt to calculate observables in these cases, since they are meaningless.
* Downloads surface albedo data from OSF (closes #178). Full functionality on this part depends on #74.
* Updated workflows to use Cache version 4


src/proteus/utils/plot.py
timlichtenberg, line 1: Great! I suggest to add:

PLATO: https://link.springer.com/article/10.1007/s10686-020-09660-1
"PLATO_blue": 500-675 nm
"PLATO_red": 675-1125 nm

LIFE: https://www.aanda.org/articles/aa/pdf/2022/08/aa40366-21.pdf
4-18.5 um

HWO: https://ntrs.nasa.gov/api/citations/20240006497/downloads/HWO%20Engineering%20View%20Status%20Plans%20Opportunities.pdf
100nm - 2500nm

nichollsh: Thanks a lot! I couldn't easily find references for other instruments, hence the generic IR bands from wikipedia. I will add these.

timlichtenberg (approved):
Great additions, will be very useful!

timlichtenberg (approved):

PR #181, opened 2024-09-18
Zephyrus

lsoucasse:
This work introduces a wrapper to interface with the escape models and connects with Zephyrus as a PyPI package.


lsoucasse:
Hi @EmmaPostolec,

This is ready for review.

I just spotted an issue in the current main. When creating the star class [here](https://github.com/FormingWorlds/PROTEUS/blob/1eace0566761285af844498ab4452c4d1aa1a831/src/proteus/utils/escape.py#L113), it gives the star age which does not make sense as we are requesting the whole track instead and do age interpolation during the simulation. In the new version I am now creating the star object with mass and rotation rate only.

EmmaPostolec (approved):
Perfect ! Thanks @lsoucasse

PR #183, opened 2024-09-18
Fixes to dummy atmosphere and other minor changes

nichollsh:
- Fixes `rho_obs` key error when running dummy atmosphere
- Added new observing filters
- Restored old script for downloading spectra from MUSCLES. This must have been deleted in the past, but I needed it for getting the L 98-59 spectrum.


lsoucasse (approved):
Good, thanks!

Issue #126, opened 2024-08-06
Generic submodule handling


nichollsh:
We recently discussed how best to handle many submodules in a simple and transparent manner. For example, in cases where the interior can be handled by either SPIDER or Aragog. 

This can be done by having an `interior.py` file (for example) which handles generalised interior stuff, and then calls the appropriate submodule-specific functions (e.g. `RunSPIDER` in `spider.py`) as necessary. This will reduce the amount of duplicate code and therefore also make things easier to maintain.

Touches on #74 and #62

nichollsh:
I agree. We could format the proteus package directory like so:

- src/proteus/
  - interior/
    - \_\_init__.py
    - spider.py
    - aragog.py
  - atmosphere/
    - \_\_init__.py
    - janus.py
    - agni.py
    - dummy.py

And so on

stefsmeets:
@nichollsh I was just about to make this post, this is exactly what I had in mind as well! 😅

Eventually, all files in the `plot`, `utils`, `tools` directories and `proteus.py` should either move somewhere under `src/janus`. Or, if they are simply scripts that you expect users to edit (i.e. these do not provide any library functions), we can move them under `./scripts` or `./examples`.

timlichtenberg:
The structure above sounds reasonable to me. What do we do with "interface" modules, such as the future atmodeller module, and ZEPHYRUS? Could all go in one of the two, but then MORS becomes a bit weirdly placed.

lsoucasse:
I think we just follow the main proteus script, each tick being a subdirectory:
- stellar flux (Mors)
- escape (Zephyrus, dummy, others...)
- atmosphere(agni, janus, dummy..)
- outgassing (atmodeller, others...)
- interior (Spider, aragog, ...)

timlichtenberg:
Ok, makes sense to me. Let's call "stellar flux" "star" for simplicity?

lsoucasse:
Just to not duplicate the work, I have started this in the atmos_wrapper branch.

Issue #62, opened 2024-01-29
Introduce support for aragog interior model


nichollsh:
The C-based version of SPIDER is being deprecated. There is a new version written in Python being developed (https://github.com/ExPlanetology/aragog), which uses SciPy rather than PETSc and thereby simplifies things significantly.

It will therefore be necessary to implement [pySPIDER](https://github.com/ExPlanetology/aragog) within PROTEUS. Implementation of Dan's new atmosphere/outgassing/EOS code will probably also be necessary.

nichollsh:
Once this has been done we can probably think about re-working the installation method for the framework. At the moment it's very clunky and will make it hard for others to reproduce our results.

Also, this issue makes Issue #43 redundant.

timlichtenberg:
https://github.com/ExPlanetology/aragog/issues/6
https://github.com/ExPlanetology/aragog/issues/7

These two issues are relevant for this. In particular https://github.com/ExPlanetology/aragog/issues/6 needs to be resolved to enable physically meaningful simulations with aragog.

timlichtenberg:
We plan to organise a Zoom meeting with @djbower, myself, @planetmariana, and @planetmariana in ~late Aug/early Sep to start this item.

nichollsh:
Issue #126 should probably be addressed first.

lsoucasse:
Hi @planetmariana @timlichtenberg ,

I have setup a first draft of the aragog interface in this [branch](https://github.com/FormingWorlds/PROTEUS/tree/aragog-constant), see `src/proteus/interior/aragog.py`. For now it runs from some default options and the output solution is not put.

Now I need your help to:
- correctly fill the initial input parameter from proteus options in `SetupAragogSolver`
- update the boundary conditions in `UpdateAragogSolver`
- collect the output from solver solution in `GetAragogOutput`

Maybe you can have a look by next Monday and we go though all this together during our meeting.

stefsmeets:
I'd highly recommend setting this up as a class instead of a global:

```python
class AragogSolver:
    def setup(...):
    def update(...):
    def run(...):
    def get_output(...):
```

Shouldn't take more than 5 min to implement this. This will make it much easier to refactor/update the code at a later stage, and also keep the code decoupled for parallelization.

lsoucasse:
Hi @planetmariana and @timlichtenberg ,

I have setup two inputs for testing aragog:
[...]

PR #185, opened 2024-09-19
Dummy interior model and further work on generalising submodules

### LINKED ISSUE #184: Setup a dummy interior model
I think a dummy interior model would be helpful for testing and also for skipping the PETSC/SPIDER step. Something like the dummy atmosphere model would be valuable. Any idea @timlichtenberg, @planetmariana @nichollsh ?

nichollsh:
- Adds a dummy interior model. This is not meant to be physically accurate, but to emulate the behaviour of a real interior model in terms of cooling and solidification. Closes #184. Runtime is on the order of ~3 minutes.

- Generalised access to interior models (stored inside `interior/`) which also makes way for Aragog support (see #126 and #62). Moved macro time-stepping to `interior/timestep.py` so that it can be shared by all interior models.

- Rewrite `cpl_interior.py` to remove old plotting class. Plotting formulation is now comparable with `cpl_atmosphere.py` and `cpl_stacked.py`.

- Output sampling for plotting routines now uses a single shared function.

- Moved VULCAN wrapper into new `atmos_chem/` directory.

- Updated docs.


nichollsh:
Will mark this ready for review when the examples have updated. This is the best we can do in lieu of tests.

src/proteus/atmos_clim/janus.py
```diff
@@ -199,7 +199,7 @@ def RunJANUS(atm, dirs:dict, OPTIONS:dict, hf_row:dict, hf_all:pd.DataFrame,
 
         # run JANUS
         atm = MCPA_CBL(dirs, atm, trppD, rscatter, method=search_method, atol=tol,
-                        atm_bc=int(OPTIONS["F_atm_bc"]), T_surf_guess=float(T_surf_old)-0.5, T_surf_max=float(T_surf_max))
+                        atm_bc=int(OPTIONS["F_atm_bc"]), T_surf_guess=float(T_surf_old)-0.1, T_surf_max=float(T_surf_max))
```
lsoucasse, line 202: I am a bit afraid about this change. How does this affect all the results stored in the example folder? Is it why the earthdemo helpfile is changin?

nichollsh: This is intended to improve performance, because often the next temperature solution is only ~0.1 K cooler than the previous one. Therefore, by moving the initial guess closer to the previous solution the model should have to do fewer steps in order to determine the next temperature. In theory it shouldn't change the model output since the solver should be finding the same solution regardless of the initial guess.

However, we can revert this and leave it for a later work if you feel that that's best.

lsoucasse (approved):
Changes look good. This interior dummy is really valuable and the interior wrapper will facilitate aragog integration, great work!

I ran the coupled dummy model and it seems to work fine: I end up with an atmosphere escape after 4.37 Myr (3 min computing time). I got some Nan values for species VMR but I guess it is because pressure goes down to zero.

I am wondering if we should change all the files in the example dummy folder (toml file, log file and png plots) ? Once this is merged, we can take easily set up a test that checks output after a couple iterations of the full dummy model.

src/proteus/atmos_clim/janus.py
lsoucasse: Makes sense. I think you can leave it like this.

nichollsh:
Thanks, Laurent! On my computer the dummy configuration also yields complete atmosphere loss after 4.37 Myr, so it's really nice to see we are getting the same results. As you say, the NaN values come about when the entire volatile inventory has been depleted. Maybe we should add a better check for this, but it doesn't really matter too much since it can only happen at the end of the simulation (by definition).

I agree that we should update the examples, especially for the dummy module since it's now actually configured differently (dummy interior vs SPIDER) to the current example. I will do this and then merge the PR.

lsoucasse:
> Thanks, Laurent! On my computer the dummy configuration also yields complete atmosphere loss after 4.37 Myr, so it's really nice to see we are getting the same results. As you say, the NaN values come about when the entire volatile inventory has been depleted. Maybe we should add a better check for this, but it doesn't really matter too much since it can only happen at the end of the simulation (by definition).
> 
> I agree that we should update the examples, especially for the dummy module since it's now actually configured differently (dummy interior vs SPIDER) to the current example. I will do this and then merge the PR.

Attached log file and helpfile if it can helps cross-checking the results.
[proteus_00.log](https://github.com/user-attachments/files/17061783/proteus_00.log)
[runtime_helpfile.csv](https://github.com/user-attachments/files/17061787/runtime_helpfile.csv)

PR #188, opened 2024-09-24
Dummy test

lsoucasse:
This work introduces the first integration test for proteus using the dummy model. It checks the full data stored in `runtime_helpfile.csv`and compares all png plot output.

- The test takes 11 min to run on the CI (instead of 3 min locally).
- The coverage goes from 16% to 40%.
- Results are not reproduced exactly from one run to another. See [this](https://github.com/FormingWorlds/PROTEUS/actions/runs/11012949252/job/30580205428) passing and [this](https://github.com/FormingWorlds/PROTEUS/actions/runs/11015304138/job/30588031577) not with same tolerance 1e-3.


nichollsh:
Thanks for this, @lsoucasse. The coverage improvement is nice to see.

Regarding comparison of the helpfiles: does it tell you which variables are not up to the required tolerance? If it's on variables that don't matter too much, then this could be okay. If it's coming from the dummy test, then maybe it is associated with undefined behaviour in the last iteration?

Regarding the image comparison: if this is causing problems, maybe we can avoid using this method? If the check on the helpfile passes, then that should be sufficient, since the plot is only plotting the data in the helpfile anyway.

nichollsh:
And a note on the model runtime: I have found that the model runs *a lot* faster if you change the plotting period to a larger number.

lsoucasse:
> And a note on the model runtime: I have found that the model runs _a lot_ faster if you change the plotting period to a larger number.

I can experiment with a lower dummy rate and see if it passes with lower tolerance. Would a value of 2e5 be relevant?

We only need the final plot so I guess setting the `plot_iterfreq` key to 0 would be OK?

nichollsh:
Yeah 2e5 kg/s sounds okay. We could probably even do 1e4 kg/s.
Setting `plot_iterfreq=0` should indeed prevent any plotting except in the final step.

stefsmeets:
Hi @lsoucasse I would say that if the plots are not reproducible, then that is a great result of having this test in the first place! I would avoid rescaling / increasing the tolerance to get the tests to pass, and instead mark the tests as xfail until we figure out what is causing the difference.

lsoucasse:
> Hi @lsoucasse I would say that if the plots are not reproducible, then that is a great result of having this test in the first place! I would avoid rescaling / increasing the tolerance to get the tests to pass, and instead mark the tests as xfail until we figure out what is causing the difference.

But to me the plots are reproducible. I extracted the CI results using GH artefacts and I cannot see the difference. The diff plot in the comment above shows the frame and the labels which are artifical differences. I had to resize the images before comparing and I blame this step for creating these artificial differences and high rms.

stefsmeets:
Ah, could be font rendering or anti-aliasing settings.

tests/integration/test_dummy_integration.py
```diff
@@ -0,0 +1,59 @@
...
+    hf_all_ref = ReadHelpfileFromCSV(ref_output)
+
+    assert_frame_equal(hf_all, hf_all_ref, rtol=1e-2)
+
+def test_plot_dummy_integration(test_dummy_run):
+
+    out_dir = PROTEUS_ROOT / 'output' / 'dummy'
+    ref_dir = PROTEUS_ROOT / 'tests' / 'data' / 'integration' / 'dummy'
+
+    for image in IMAGE_LIST:
```
stefsmeets, line 44: Consider splitting these tests using `pytest.mark.parametrize`. See examples here: https://docs.pytest.org/en/7.1.x/how-to/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions

tests/integration/test_dummy_integration.py
```diff
@@ -0,0 +1,59 @@
...
+        out_img = out_dir / image
+        ref_img = ref_dir / image
+        tolerance = 14
+
+        # Resize images if needed
+        out_tmp, ref_tmp = resize_to_match(out_img, ref_img)
```
stefsmeets, line 50: I think we should avoid resizing, and make sure we find the cause of the different rendering between systems. To make sure we do not hold up this PR, we can mark the failing bit with `pytest.xfail` for the moment.

https://docs.pytest.org/en/6.2.x/skipping.html#xfail-mark-test-functions-as-expected-to-fail

tests/integration/test_dummy_integration.py
```diff
@@ -0,0 +1,59 @@
...
+
+def test_plot_dummy_integration(test_dummy_run):
+
+    out_dir = PROTEUS_ROOT / 'output' / 'dummy'
+    ref_dir = PROTEUS_ROOT / 'tests' / 'data' / 'integration' / 'dummy'
+
+    for image in IMAGE_LIST:
+        out_img = out_dir / image
+        ref_img = ref_dir / image
+        tolerance = 14
```
stefsmeets, line 47: If we need to make the tolerance this high, then comparing the plots has little use because we will not catch actual differences.

```suggestion
        tolerance = 14
```

tests/integration/test_dummy_integration.py
lsoucasse: This is what I used at first instance, but the problem is that with parametrize it re-executes the fixture (the integration test) for each image, which does not make sense.

tests/integration/test_proteus_start.py
```diff
@@ -1,29 +0,0 @@
...
-
-
-def test_dummy_run():
-    config_path = PROTEUS_ROOT / 'input' / 'dummy.toml'
-
-    runner = Proteus(config_path=config_path)
-
-    runner.config['log_level'] = 'WARNING'
-    runner.config['plot_iterfreq'] = 0
-    runner.config['iter_max'] = 0
```
stefsmeets, line 23: Consider playing with the number of iterations to speed up the test. On my machine it takes about 13 minutes to run the test, if I set iter_max to 0 it goes to about 22 seconds. 

Even iter_max of 10 or 100 is quite manageable, as it only adds a few seconds of runtime. However, plot_iterfreq is a permormance killer, considering that the default in dummy.toml is 1 (plot every iteration). Consider setting to 0 to only plot at the end.

| iter_max | plot_iterfreq | time (s) |
|-|-|-|
| 0 | 0 | 22 |
| 10 | 0 | 23 |
| 100 | 0 | 28 |
| 10 | 1 | 49 |


iter_max=10 and

tests/integration/test_dummy_integration.py
lsoucasse: OK, will leave it for now. I got the following [error](https://github.com/FormingWorlds/PROTEUS/actions/runs/11013464586/job/30581882764) without resizing.

stefsmeets: Yeah, I understand that you get an error, but this means there is an underlying issue preventing reproducibility to figure out 😁

tests/integration/test_dummy_integration.py
stefsmeets: You can set the fixture to only run once using `scope='module'`

lsoucasse:
With escape rate of 2e4 kg/s I can drop the tolerance down to 1e-4. And plotting only at the end reduces the total test run time to 4 min. I think plot_iterfreq=0 (or at least 50/100) should be the default option in all config files as is slows the code so much.

I set down the tolerance for the plots and flagged it with xfail. I will create an issue for solving this.

tests/integration/test_proteus_start.py
lsoucasse: Set plot_iterfreq to zero.

tests/integration/test_dummy_integration.py
lsoucasse: Done, thanks for the tip.

tests/integration/test_dummy_integration.py
lsoucasse: Will create a separate issue to solve this.

nichollsh (approved):
Nice work. If the test on the CSV file is passing with a reasonable tolerance, then this is good enough for me. Future work can look into checking the figures, but it's less important (to me, anyway) since it only needs to test the ability to plot data, rather than the data itself.

Happy for this to be merged into main if @stefsmeets is also.

PR #189, opened 2024-09-24
Implement new config

### LINKED ISSUE #74: Rework configuration file

stefsmeets:
This PR sets up the schema for the new config using [attrs](https://www.attrs.org/en/stable/index.html).

I made one change from the proposal, I changed `exec`->`params`, because exec is a reserved parameter in Python.

### Todo
- [x] Find way to store docstrings -> Although it is possible to store them on the attribute, this gets very cluttered. I suggest to pick this up in a follow-up PR and add them to the docstring.
- [x] Add more validators -> Follow-up in separate PR
- [x] Regenerate table for documentation -> Combine with docstring PR.
- [x] Convert older configs(?) -> Separate PR

Closes #74


stefsmeets:
Hi @lsoucasse @nichollsh Could you have a look at this PR and let me know what you think?

I want to get this merged first to avoid ballooning this PR, and so that we have a baseline to work from. I will pick up other tasks (validation, documentation, integration) afterwards.

nichollsh (approved):
Looks like a good start! Thanks, Stef.

lsoucasse (approved):
Looks good, thanks! Now we have a clear separation between common options and module specific options.

I just think we should replace all atmos key by atmos_clim to be consistent with the repository structure and allow for future atmos_chem key for chemistry.

lsoucasse (approved):
Great!

PR #191, opened 2024-09-26
Time-averaged orbital distance and minor changes

### LINKED ISSUE #187: Time-averaged orbital distance
The new configuration options include an orbital eccentricity. This is needed for the escape calculations with Zephyrus.

However, we can also make use of this parameter elsewhere. It can be used for calculating tidal heating (if we add a tidal heating calculation) but also to more accurately calculate the orbital distance. 

We currently set the orbital distance equal to the semi-major axis, but this does not perfectly translate when calculating the planetary instellation. The SM axis is not the time-averaged distance. This stack exchange thread has a nice analytical solution for time-averaged orbital distance: https://physics.stackexchange.com/a/715749

`r_t = a(1 + 0.5*e^2)`

From this, we can see that the relative difference in instellation when using a non-zero eccentricity `e` is equal to `4/(e^2 + 2)^2`. For example, `e=0.3` would give an instellation of 0.916x the value that would be found had we assumed `e=0`. This is quite significant.

It would be trivial to add this calculation of time-averaged orbital distance into PROTEUS.

nichollsh:
### Large changes
- Now calculates orbital `separation` as the time-averaged planet-star distance. This accounts for eccentric orbits. This value is stored in the helpfile as a new column in units of metres. Should be accessed instead of the config variable where possible. Closes #187.
- Time-stepping scheme decreases the time-step more aggressively when required, which improves stability when approaching the end of the simulation.
- The steady-state convergence criterion is now triggered more reliably 

### Small changes 
- AGNI now uses the same `min_temperature` as JANUS, which is set in the PROTEUS config file
- Fixed some typos and added some comments. Added a new file `_stars.txt` containing useful information on specific stars, which is often entered into configuration files.
- Interior plot now uses nicer units, to avoid cluttering the axis tick labels
- Saves runtime config file using TOML library, rather than just copying the file. This will be useful if we decide to use a template file in the future. It's otherwise useful for making absolutely sure that the variables in the file in the output folder correspond to the ones stored in memory.
- The `GetStellarSpectrum` tool now ensures that wavelengths are always saved in ascending order, with no duplicates. Not doing this was causing problems when interpolating the spectra generated by this tool.


stefsmeets:
This is an enormous PR, anything specifically that you want me te review?

nichollsh:
Sorry about that! 

Most of the changes picked up by GitHub are actually from me find+replacing the config file key `mean_distance` -> `semimajoraxis` and from updated plots and notebooks.

If you could just look at any changes to `py` files, that would be more than sufficient.

examples/dummy/init_coupler.toml
lsoucasse, line 1: Why do you want to reformat this file at this stage? I realise that beforehand it only duplicates the `input/dummy.toml` file and now the two are different.

input/dummy.toml
lsoucasse, line 1: This will conflict with the new integration test.

examples/dummy/init_coupler.toml
nichollsh: I did this because I found that the current method of copying the file causes problems when trying to resume a simulation (sometimes).

If you tell PROTEUS to resume from the config file in the output folder, it tries to copy the file from a source and destination which are the same. However, if we instead write the config file from the variables in memory then this issue is avoided. 

The only difference produced by this change is that the output files now lack comments and whitespace, but the keys and variables should be unchanged.

examples/dummy/plot_elements.png
lsoucasse, line 1: Maybe the `example/dummy` folder could be deleted as it duplicates the integration test data?

input/dummy.toml
nichollsh: Because of the renamed keys? We could merge the integration test first and then I could fix this PR afterwards.

lsoucasse: I've just merged the integration test. Results need to be updated as some values are changed (escape rate, eccentricity, dt_rtol...).

examples/dummy/init_coupler.toml
lsoucasse: Oh, OK. Do you mean this is a backup file to resume the simulation?

nichollsh: Yes.

I do think there are a couple of reasons to have PROTEUS output its input config file into the output folder. One of them is so that we can see how the model was run. Another is so that we can resume the simulation using the exact some config options which which it was originally run. It's more reliable to do the latter using the copy of the config file, rather than the one in `input/`, since the one in `input/` could potentially have been changed.

lsoucasse (approved):
Thanks Harrison for the changes! To me it is OK, providing the dummy integration test passes.

nichollsh:
Thanks for your review! I probably will have to change the reference data for the test, since we are now using some new keys in the output helpfile. But in theory it should get the same numbers.

input/dummy.toml
lsoucasse: I don't know if it matters much but it seems that the new dummy setup has triggered an oscillating behaviour on the heat flux.

nichollsh: I noticed this too. I think it was an existing issue, because I have had problems with similar behaviour for a while. I don't think it matters too much, since it's just a case of the planet oscillating around the radiative equilibrium state. It gets better if you use more rigorous time-stepping, but I don't know what makes it worse.

nichollsh: It also exists in cases with JANUS/SPIDER. I had previously blamed them for causing the behaviour, but now that we see it in the dummy-only setup then something else must be to blame.

lsoucasse: Interesting. Need to keep this in mind. This does not prevent the test from passing!

nichollsh: I'll make an issue for it, so that we don't forget.

nichollsh: https://github.com/FormingWorlds/PROTEUS/issues/193

PR #195, opened 2024-09-26
Plotting I/O improvements

### LINKED ISSUE #194: Improve plotting speed
Tests have found that `plot_iterfreq` - which sets how often the model should make plots - significant impacts performance. This is not only annoying, but potentially limits the number and resolution of the simulations, given finite resources.

We should optimise the plotting speed. This could be done by avoiding redundant reads/writes to the disk.

nichollsh:
Improvements to plotting speed at runtime by avoiding extra file reads. There are now common functions for reading NetCDF and JSON files from the disk. These are used to read the data into memory only once (or keep it in memory entirely for some cases), rather than having each plotting function individually read the data from the disk. Closes #194.

Note that the improvement scales with the number of data files to be read, and the frequency at which plotting is performed. We should still consider using larger `plot_iterfreq` values for further performance gains.

Should also make things easier when implementing Aragog, since we won't need to read files from the disk at all when making plots at runtime.

Also added some more instrument bandpasses for reference.

### DIFF
--- a/input/default.toml
+++ b/input/default.toml
@@ -38,7 +38,7 @@ time_star                   = 100.0e6  	# yr, time since star formation
 time_target                 = 4.567e+9  # yr, target time for MO evolution
 
 # SOCRATES spectral file to use (relative to FWL_DATA folder)
-spectral_file               = 'spectral_files/Frostflow/256/Frostflow.sf'
+spectral_file               = 'spectral_files/Frostflow/16/Frostflow.sf'
 
 # Stellar heating toggle, 0: disabled | 1: enabled
 stellar_heating             = 1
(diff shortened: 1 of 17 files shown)


lsoucasse (approved):
Thanks Harrison, changes look OK.

I did not see any significant improvement on the running time though compared to previous implementation. Did you find some?

nichollsh:
It doesn't make a difference to the dummy case because the plotting frequency is low, and it's not reading NetCDFS or JSON files anyway. When I ran some longer cases with JANUS, plotting every iteration, it reduced the runtime by between 5 and 10%.

nichollsh:
Thanks for the review!

lsoucasse:
> It doesn't make a difference to the dummy case because the plotting frequency is low, and it's not reading NetCDFS or JSON files anyway. When I ran some longer cases with JANUS, plotting every iteration, it reduced the runtime by between 5 and 10%. There's some variation depending on the number of model levels, and the the number of files it has to read.

Good to know. Thanks.

PR #197, opened 2024-09-28
Remove SPIDER and PETSc submodules

### LINKED ISSUE #196: Remove SPIDER/PETSc submodules
We aim to replace SPIDER with Aragog as the main interior module. This means that we should remove SPIDER (and its dependency PETSc) as Git submodules, and instead source them by other means.

One option is to use a simple script (c.f. `get_socrates.sh`) so that the user (or a workflow) can download them only as required.

@stefsmeets @lsoucasse do you have any other suggestions for how to handle this? Or does using a script sound okay?

nichollsh:
Removed SPIDER and PETSc submodules (closes #196). They can now be installed via scripts in the `tools/` directory. These scripts will download PETSc and SPIDER as zip files from OSF, decompress, and compile them. I have added functions for calling these scripts in `utils/data.py`, which in the future should be made accessible from the CLI. Updated the docs correspondingly. 

In the future, we could also consider having PROTEUS run the `get_petsc.sh` and `get_spider.sh` scripts automatically if the user has selected SPIDER as the interior module of choice. I leave this for a future PR.

Also added some useful constants to `constants.py`.

Some optimisations to the module imports in the interior wrapper. Only imports the required functions, while previously it would import the DummyInterior functions even when using SPIDER for the interior model.

Renamed other files in `tools/` to use consistent capitalisation.


docs/troubleshooting.md
```diff
@@ -6,6 +6,15 @@ encounter errors or other issue that you cannot solve via the standard
 step-by-step guide or the advice below contact the developers (see
 `contact`{.interpreted-text role="doc"}).
 
+## GitHub: Cannot clone module
+
+Have you added your SSH key to GitHub? See these pages for guidance:
+* https://docs.github.com/en/authentication/connecting-to-github-with-ssh/checking-for-existing-ssh-keys
+* https://docs.github.com/en/authentication/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent
+* https://docs.github.com/en/authentication/connecting-to-github-with-ssh/adding-a-new-ssh-key-to-your-github-account
+* https://docs.github.com/en/authentication/connecting-to-github-with-ssh/testing-your-ssh-connection
```
stefsmeets, line 15:
```suggestion
* [Checking for existing SSH keys](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/checking-for-existing-ssh-keys)
* [Generating a new SSH key](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent)
* [Adding a new SSH key to your Github account](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/adding-a-new-ssh-key-to-your-github-account)
* [Testing your SSH connection](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/testing-your-ssh-connection)
```

src/proteus/proteus.py
```diff
@@ -242,7 +239,7 @@ def start(self, *, resume: bool = False):
                 solvevol_target[e] = hf_row[e + "_kg_total"]
 
         # Download basic data
-        download_basic()
+        download_sufficient_data(self.config)
```
stefsmeets, line 242: Love this function name 😅

src/proteus/utils/data.py
```diff
@@ -31,19 +59,131 @@ def DownloadSurfaceAlbedos():
...
+
+    import os
+    import subprocess as sp
+
+    # Get path
+    workpath = os.path.join(dirs["proteus"], "petsc")
+    workpath = os.path.abspath(workpath)
+    if os.path.isdir(workpath):
+        # already downloaded
+        return
```
stefsmeets, line 149: Might want to add a log statement here that it is already downloaded

src/proteus/utils/data.py
```diff
@@ -31,19 +59,131 @@ def DownloadSurfaceAlbedos():
...
+    Download and install SPIDER
+    """
+
+    # Need to install PETSc first
+    get_petsc(dirs)
+
+    log.info("Setting up SPIDER")
+
+    import os
+    import subprocess as sp
```
stefsmeets, line 172: These can be moved to the top of the file

src/proteus/utils/data.py
```diff
@@ -31,19 +59,131 @@ def DownloadSurfaceAlbedos():
...
+
+def get_petsc(dirs:dict):
+    """
+    Download and install PETSc
+    """
+
+    log.info("Setting up PETSc")
+
+    import os
+    import subprocess as sp
```
stefsmeets, line 142: These can be moved to the top of the file

stefsmeets (approved):
Looks good to me, left a few small comments

src/proteus/proteus.py
nichollsh: Haha - thanks. It went through several iterations before reaching this point

nichollsh:
Thanks, Stef!

stefsmeets (approved):

PR #198, opened 2024-09-30
MacOS tests and standalone setup

### LINKED ISSUE #172: Tests: MacOS runner
It would be useful to setup the tests to run a matrix of different Linux and Python versions, in order to catch potential issues which arise only for certain platforms.

C.f. https://github.com/FormingWorlds/MORS/issues/22

nichollsh:
- Removed SOCRATES from tests, since we use dummy atmosphere for this. SOCRATES has its own workflow for testing its compilation.
- Rework download functions and CLI so that PROTEUS doesn't depend on JANUS to do this
- Moved some of the atmos_clim imports into the functions in which they are used. This avoids performing redundant imports when only using the CLI, or when only using a specific climate model.
- Added MacOS to test matrix. Closes #172.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -16,7 +16,7 @@ on:
 
 jobs:
   test:
-    # if: github.event.pull_request.draft == false
+    if: github.event.pull_request.draft == false
     name: Run Coverage and Tests
     strategy:
       matrix:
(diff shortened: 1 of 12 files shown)


lsoucasse (approved):
Changes look good. It is nice to run the tests on MacOS. Also the CLI is now more consistent with both data and submodules being fetched.

PR #199, opened 2024-09-30
Dummy integration

lsoucasse:
I realized that the dummy integration test was not triggered because it was only a fixture of the plotting tests that were allowed to fail. With the current state of the repo, the dummy integration test is actually failing. I revert the dummy set up to a previous config which did not trigger spurious oscillation on the flux. I also switched the star model to use the Baraffe track.


tests/integration/test_dummy_integration.py
```diff
@@ -21,24 +19,25 @@
...
 
-    hf_all = ReadHelpfileFromCSV(output)
-    hf_all_ref = ReadHelpfileFromCSV(ref_output)
+    hf_all = ReadHelpfileFromCSV(out_dir)
+    hf_all_ref = ReadHelpfileFromCSV(ref_dir)
 
-    assert_frame_equal(hf_all, hf_all_ref, rtol=1e-4)
+    assert_frame_equal(hf_all, hf_all_ref, rtol=5e-3)
 
 @pytest.mark.xfail
```
stefsmeets, line 38: Maybe not part of this PR, but consider to only xfail on the assertion, not on any regular failures in the workflow itself

```suggestion
@pytest.mark.xfail(raises=AssertionError)
```

stefsmeets (approved):
Looks good to me! 🎉

nichollsh:
Thanks for noticing this Laurent. Good to see the oscillations have gone away, although it's surprising given that the relative step tolerance is now higher/looser. We should look into what's causing them.

lsoucasse:
Need your approval again @nichollsh

nichollsh (approved):

Issue #208, opened 2024-10-17
Crash with undefined escape model


stefsmeets:
If the escape model is not defined (`config["escape_model"] == 0`), `RunEscape` crashes because `solvevol_target` is undefined. What is the expected result here?

See this function:
https://github.com/FormingWorlds/PROTEUS/blob/181d4ceb790eba15d69daaa229b73857a3738d1b/src/proteus/escape/wrapper.py#L18-L61

lsoucasse:
Actually, RunEscape must not be called if not dummy or zephyrus.

I would remove the if statement l.53 and put a statement like this in `proteus.py`
`if config["escape_model"]:`
`    RunEscape(parameters...)`

stefsmeets:
That will only bring the undefined variable problem to the main loop of proteus as far as I can see, which expects solvevol_target in several places.

nichollsh:
Closed by https://github.com/FormingWorlds/PROTEUS/pull/211

PR #211, opened 2024-10-18
Generalised outgassing wrapper

### LINKED ISSUE #210: Outgassing model wrapper
In preparation for support for atmodeller (#82) we should add a generalised wrapper for accessing outgassing modules (see also #126).

This is separate to a vapourisation module, like Lavatmos or Vaporock, which should come under a different category of module (probably).

nichollsh:
This PR generalises access to the outgassing wrapper in a similar manner to the interior and atmosphere wrappers. This will make it possible to integrate atmodeller, for which there is now a stub file, and also potentially a dummy outgassing module. Closes #210.

Minor fixes:
* Typo in `data.py`
* Default configuration file
* Access config with attributes rather than keys.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,6 +49,9 @@ ARAGOG
 atmodeller
 ATMODELLER
 Atmodeller
+zephyrus
+ZEPHYRUS
+Zephyrus
 
 # misc
 ######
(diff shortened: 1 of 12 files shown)


lsoucasse (approved):
Nice! It also addresses #208.

PR #214, opened 2024-10-19
Comparison with population data, rework planetary vs interior mass/radius

### LINKED ISSUE #212: Compare PROTEUS (grid) output to population data
This paper[1] has a really interesting comparison between a simple evolutionary model and observed exoplanet densities. They match their model input (water mass fraction) to observed density, to constrain the water mass fraction on a given planet.

We could do something similar with PROTEUS, or at least make it simple to compare model outputs to the observed population. A simple plot of planet mass/radius/density on top of the population scatter would be an interesting start. Connected to #204, #162.

The PlanetS database can be accessed here [2], and downloaded in the VOTable (XML) format here[3].

[1] https://arxiv.org/abs/2409.17394
[2] https://dace.unige.ch/exoplanets/
[3] https://dace.unige.ch/downloads/exoplanets/PLANETS.vot

### LINKED ISSUE #213: Distinguish between M_planet and M_interior
Throughout the code we currently refer to the variable `M_planet`. However, in some cases this is used as the total planet mass, and in other cases as the mass of the interior (i.e. everything enclosed by the atmosphere-mantle interface).

We should use two variables to describe this: `M_planet` and `M_int`, where `M_planet = M_int + M_atm`, where `M_int` is the mass of the interior and `M_atm` is the mass of the atmosphere.

This is connected to #212, #50, #62, #82.

nichollsh:
Plots mass/radius/density throughout the model evolution on the mass vs radius diagram using planetary population data. Also plots this as density vs time, which will be particularly interesting for comparison with planets thought to be undergoing escape. The model downloads data from OSF to be able to do this. Closes #212.

Rework planet mass and radius variables. `M_int` now represents all dry mass enclosed in the interior (core+mantle), not including volatiles. `M_planet = M_int + M_atm` is the total planet mass. `R_int` represents the radius of the atmosphere-mantle interface, and is also what the user provides in the config file. `R_planet` no longer exists, but the best analogue for this is `z_obs+R_int`. We can think about how to rework the mass calculations in the future, but this is already somewhat better than treating `M_planet` and `M_int` as if they're the same thing. Closes #213.

Since the new variable adds an extra key in the output csv file, I had to update the dummy test data, but we can see that it's getting the exact same results. This PR comes with a lot of file changes because I updated the examples and renamed variables throughout the code. Feel free to ignore the updated examples, and you can see that most of the changes are find+replace. You can filter to just viewing changes to py files, if that helps.

Other minor things:
- Fixed incorrect axis limits on interior plot
- Make sure that rounding errors don't cause problems with CALLIOPE outgassing. Sometimes the elemental mass would be ~1e-34, which is difficult for calliope to solve in some instances. When the mass is small, it is now rounded to zero, which fixes the problem.


nichollsh:
I don't know why the test is failing now. It seems to be due to some issue reading the configuration file?

The test runs fine when I execute the same commands in the Python REPL, and also works fine when I run `proteus start --config input/dummy.toml`.

timlichtenberg:
Do you want me to review this now, or do you first want to figure out why the tests are failing?

nichollsh:
It would be great if you could review it now, and check from a physical perspective if these changes make sense. 

I think we can potentially discuss this configuration issue with @stefsmeets at the next PROTEUS meeting.

examples/default/plot_emission.pdf
timlichtenberg, line 1: The y-axes on this plot are not formatted correctly by default, they should allow wider variation.

examples/hd63433d/init_coupler.toml
timlichtenberg, line 1: Same here with the comments.

examples/hd63433d/plot_observables.png
timlichtenberg, line 1: These values have shifted a very tiny amount, is this due to how the mass budget is calculated?

examples/hd63433d/plot_population_time_density.png
timlichtenberg, line 1: My suggestion for this plot would be to normalise mass-radius curves of the same radius of the planet at a particular time-step. Here are lots of M-R curves: https://drive.google.com/open?id=1M6MXZ0OscVtF713r86H3MoRRBj60IMUF&usp=drive_fs. 

First thing would be to normalise the black line relative to the 'Zeng_2019/massradiusEarthlikeRocky.txt' at current radius. The water line should be normalised to, for example, 'Zeng_2019/massradius_100percentH2O_300K_1mbar.txt'. Optimally, the latter is tuned to the approx. T_eq of the planet. 

But maybe these additions are too complicated for the current pull request and should be done in another one. :)

examples/hd63433d/plot_stacked.png
timlichtenberg, line 1: This plots needs a legend explaining dashed and solid lines. An additional idea would be to also dash all lines in the atmosphere that are in the water-condensing region.

input/default.toml
```diff
@@ -114,6 +114,8 @@ author  = "Harrison Nicholls, Tim Lichtenberg"
 
 # Planetary structure - physics table
 [struct]
+
+    # Enclosed mass and radius at the atmosphere-interior interface
     mass        = 1.0       # M_earth
     radius      = 1.0       # R_earth
```
timlichtenberg, line 120: Does this also need relabeling?

src/proteus/utils/data.py
```diff
@@ -125,6 +125,21 @@ def download_stellar_spectra():
...
+
+    folder_name = 'Exoplanets'
+    storage = get_osf('fzwr4')
+
+    data_dir = GetFWLData() / "planet_reference"
+    data_dir.mkdir(parents=True, exist_ok=True)
+
+    if not (data_dir / folder_name).exists():
+        log.info(f"Downloading exoplanet population data to {data_dir}")
+        download_folder(storage=storage, folders=[folder_name], data_dir=data_dir)
```
timlichtenberg, line 142: So this is the newest CSV file from DACE? Just to double check.

timlichtenberg (approved):
Looks good overall! I made some suggestions for a few changes in the default plots, but these can also go in a subsequent branch.

examples/default/init_coupler.toml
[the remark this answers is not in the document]
nichollsh: This is a copy of the cfg file which is automatically made by PROTEUS. It loses the comments during this process, but the original one (in `input/`) still has them.

examples/default/plot_emission.pdf
nichollsh: Thanks! I'll fix this

examples/hd63433d/init_coupler.toml
nichollsh: See above

examples/hd63433d/plot_population_time_density.png
nichollsh: This is perfect. I tried to add the Zeng+19 curves myself, but it got quite complicated trying to do this from the paper. I'll add these to OSF.

examples/hd63433d/plot_stacked.png
nichollsh: I agree. This applies to the plot_interior plots as well. I'll make an issue for this too.

input/default.toml
nichollsh: As in renaming the variables?

examples/hd63433d/plot_observables.png
nichollsh: I suppose so. I will look into this.

timlichtenberg: Doesn't seem to important tbh

input/default.toml
timlichtenberg: Yes that's what I meant, but on second look it seems fine.

src/proteus/utils/data.py
timlichtenberg: Agree

timlichtenberg: Maybe label it with the date you downloaded it (in case you haven't done this anyway), for future updates

nichollsh:
Turns out that the reason it is failing is not because of the integration test. It's when the tests check that the toml files in `examples/` for validity that it fails, because the fields which are converted from `"none"` to `None` do not get saved.

See this issue: https://github.com/uiri/toml/issues/346

examples/default/init_coupler.toml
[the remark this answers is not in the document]
nichollsh: The fix to the test below also ensures that the 'backup' made of the config file contains the comments.

src/proteus/utils/data.py
nichollsh: I put the date and other information inside the file

examples/hd63433d/plot_interior.png
timlichtenberg, line 1: Great!

examples/hd63433d/plot_stacked.png
timlichtenberg, line 1: Here as well, nice!

timlichtenberg (approved):

nichollsh:
Thanks Tim!

stefsmeets:
> Turns out that the reason it is failing is not because of the integration test. It's when the tests check the toml files in `examples/` for validity that it fails, because the fields which are converted from `"none"` to `None` do not get saved. See this issue for a description of a similar problem: [uiri/toml#346](https://github.com/uiri/toml/issues/346)

This is a somewhat awkward side-effect from toml not recognizing a 'null' value: https://github.com/toml-lang/toml/issues/30

I think copying is also the way to go for the moment, depending on how much we lean on converters in the future. TOML really seems to be made as a one-way format (toml -> dict). If we want to go back (dict -> toml), the alternative is having a custom TomlEncoder to go from None -> 'none'.

Issue #204, opened 2024-10-07
PROTEUS grid-search or forward model


timlichtenberg:
For moving towards an inverse method of PROTEUS sometime down the road, we need to consider a computationally feasible approach to run many models to fit a given set of observations.

To give an example of the problem: Let's assume a given exoplanet has the following known/observed parameters with uncertainties: stellar age, orbital distance, planet radius, planet mass, transmission/emission spectrum. Given these parameters, we would like to compute the best-fitting PROTEUS models over a set of input parameters, and then compute a goodness-of-fit metric. This is essentially the description of an atmospheric retrieval, only that PROTEUS simulations are way too computationally expensive as to perform 100k+ simulations.

I am not certain yet what is the best strategy to approach this problem. Here are a few that have some opportunities and drawbacks:
- A modified chi-squared or [r-squared](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.r2_score.html) algorithm to compute some measure of goodness-of-fit for an arbitrary grid. E.g. [Madhusudhan & Seager (2009)](https://iopscience.iop.org/article/10.1088/0004-637X/707/1/24/pdf).
- Train a machine learning model on the simulation data and use the machine learning model for the retrieval, e.g., [Ardevol Martinez et al. (2024)](https://www.aanda.org/articles/aa/abs/2024/01/aa48367-23/aa48367-23.html).
- Brute-force retrieval approach, e.g., nested sampling, MCMC, or some other random sampler, ignoring the computational cost, and possibly achieving a pretty low-confidence result.

timlichtenberg:
The machine learning method of Ardevol Martinez (2024) seems potentially useful to me for this in principle. It can work together with an arbitrary "atmospheric code" (which could be PROTEUS) and does not need a large amount of training data to be generated upfront. Code here: https://github.com/franciscoardevolmartinez/FlopPITy.

nichollsh:
I agree that this would be incredibly powerful. I can imagine that running an MCMC (or similar method) would be tricky because of the slow runtimes. When we are ready to look into this, maybe we could involve someone who has experience doing retrievals with large models?

timlichtenberg:
[...]

Issue #162, opened 2024-08-30
Dynamic coding for GridPROTEUS


nichollsh:
Currently, a large amount of the configuration of GridPROTEUS.py is hardcoded in the `if __name__` statement. This would ideally be configured in another manner, such as by a configuration file or with a CLI.

stefsmeets:
I would suggest to hold off major work on this until we refactored the proteus (single) run code. My hope is to fold `GridProteus` into there as well eventually.

EmmaPostolec:
Hi, I edited tools/grid_proteus.py with new parameters that I wanted to test for the escape (Pxuv, efficiency and semimajoraxis) and I ran into the following error:

````
[2024-11-13 11:25:32] Sleeping...
[2024-11-13 11:25:32]     7 
[2024-11-13 11:25:33]     6 
[2024-11-13 11:25:34]     5 
[2024-11-13 11:25:35]     4 
[2024-11-13 11:25:36]     3 
[2024-11-13 11:25:37]     2 
[2024-11-13 11:25:38]     1 
[2024-11-13 11:25:39]  
[2024-11-13 11:25:39] Writing config files
[2024-11-13 11:25:45] Starting process manager (75 grid points, 40 threads)
[2024-11-13 11:25:45]  75 queued (100.0%),   0 running (  0.0%),   0 completed (  0.0%)
[2024-11-13 11:25:45] Uncaught exception
Traceback (most recent call last):
  File "/Users/emmapostolec/Documents/PHD/SCIENCE/CODES/PROTEUS/tools/grid_proteus.py", line 492, in <module>
    pg.run(40, test_run=False)
  File "/Users/emmapostolec/Documents/PHD/SCIENCE/CODES/PROTEUS/tools/grid_proteus.py", line 405, in run
    threads[i].start()
  File "/Users/emmapostolec/miniconda3/envs/proteus/lib/python3.12/multiprocessing/process.py", line 121, in start
    self._popen = self._Popen(self)
                  ^^^^^^^^^^^^^^^^^
  File "/Users/emmapostolec/miniconda3/envs/proteus/lib/python3.12/multiprocessing/context.py", line 224, in _Popen
    return _default_context.get_context().Process._Popen(process_obj)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

Issue #134, opened 2024-08-09
Harmonize access to Socrates from PROTEUS


lsoucasse:
At the moment JANUS and AGNI have their own tools to download and install SOCRATES. When you use PROTEUS, you should be able to switch JANUS / AGNI while having only one SOCRATES code installed.

nichollsh:
Definitely agree. AGNI only needs the RAD_DIR environment variable to be set.

lsoucasse:
Just tested to swap from Janus to Agni using the same Socrates repository. It just works fine. When installing the Julia Agni package it just adds the julia wrapper into the socrates repo. I think we just need to remove the get_socrates command in the documentation when using AGNI, assuming socrates is already there as it is a core module.

nichollsh:
That's great - glad to hear it works nicely. Presumably we will then add a get_socrates script of some form into PROTEUS?

lsoucasse:
Yes, when we remove Socrates as a submodule we need to write a cli command that runs the download and the basic compiling of Socrates. Then if you install agni it will build the julia wrapper on top of it.

lsoucasse:
Closing this issue, I will address the documentation update when working on #135.

PR #220, opened 2024-10-21
Generalise stellar wrapper, dummy star module

### LINKED ISSUE #202: Dummy star module
It would be useful to have a dummy module for stellar evolution, just as we have for the interior and atmosphere. This could be implemented alongside a generalised stellar wrapper.

This module could easily generate a stellar spectrum using a Planck function with a fixed stellar temperature. Setting the temperature to zero is equivalent to turning the star 'off'.

nichollsh:
This PR adds a dummy star module. It calculates the stellar spectrum and bolometric luminosity assuming that the star emits like a blackbody with a temperature and radius specified by the user. Added `dummy_star.toml` configuration to demonstrate this, but there isn't a test for this yet. Closes #202.

It also reworks the stellar module handling to generalise calculations between the mors and dummy modules. This mostly means moving functions into `src/proteus/star/` and calling these from `proteus.py`. This also meant moving a lot of variables in `proteus.py` into the `Proteus` class, which will also be useful for later work (e.g. #204, #162).

Minor things:
* Fixes numerical issue when volatile elemental inventory is small. Sometimes floating point errors mean that 0 kg of a given element can become ~1e-13 kg or so, which is problematic.
* Renamed `Lbol` config variable to `lum_now`, to make it clear that this is the *modern* luminosity of the star.
* Added `phys.py` for containing physics functions.
* Removed `start_proteus.py` since it is made redundant by the CLI.
* Fixes some of the deprecation warnings when accessing the config.


timlichtenberg (approved):
Looks good to me, but @lsoucasse should also give it a quick check.

lsoucasse (approved):
Nice changes! I will probably extend the baraffe_track property of proteus into a star_track to store either spada or baraffe data for the whole run.

The dummy star simulation seens to run fine on my laptop.

nichollsh:
That makes sense - I guess this is covered by #134?

nichollsh:
Thanks both!

PR #222, opened 2024-10-22
Add mass-radius curves to population plots

### LINKED ISSUE #216: Update mass-radius curves in population plots
Normalise mass-radius curves of the same radius of the planet at a particular time-step. Here are lots of M-R curves: https://drive.google.com/open?id=1M6MXZ0OscVtF713r86H3MoRRBj60IMUF&usp=drive_fs. 

First thing would be to normalise the black line relative to the 'Zeng_2019/massradiusEarthlikeRocky.txt' at current radius. The water line should be normalised to, for example, 'Zeng_2019/massradius_100percentH2O_300K_1mbar.txt'. Optimally, the latter is tuned to the approx. T_eq of the planet. 

_Originally posted by @timlichtenberg in https://github.com/FormingWorlds/PROTEUS/pull/214#discussion_r1807307834_

nichollsh:
Plots mass-radius curves from Zeng+2019 on population plots, so the user is provided with information on how the planet's simulated bulk density compares to different compositions and materials. These data are downloaded from OSF. Closes #216.

Minor things:
* Fixed bug where Tsurf doesns't update when running JANUS with 'fixed' surface condition.
* Fixed typo where density was only being calculated with `radius=z_obs`, but it should have been `radius=z_obs+R_int`.
* Clarified that the default case is running without escape enabled.


nichollsh:
Will convert from draft when examples cases have updated.

nichollsh:
It's somewhat concerning that the MacOS/Py3.10 test failed at first, and then it passed when I re-ran it. Maybe something we should look into at a later stage.

nichollsh:
@stefsmeets this is ready for review now that the test has passed.

src/proteus/atmos_clim/janus.py
```diff
@@ -101,7 +101,7 @@ def UpdateStateAtm(atm, hf_row:dict, trppT:int):
             Dictionary containing simulation variables for current iteration
     """
 
-    atm.setSurfaceTemperature(hf_row["T_surf"])
+    atm.setSurfaceTemperature(hf_row["T_magma"])
```
lsoucasse, line 104: I never got the distinction between T_surf and T_magma but the surface temperature returned by the interior models is indeed T_magma.

lsoucasse (approved):
Looks fine, nothing to add.

PR #225, opened 2024-10-23
Escape config

lsoucasse:
This PR fixes the zephyrus escape runs with the new config. Two new config keys for zephyrus have been created.

I renamed the omega key to rot_pctle to avoid confusion. The spada star tracks are now created using this percentile instead of omega to avoid inconsistent set up. You can easily go form one to the other using
- `print(mors.Percentile(Mstar=1.0, Omega=10.0))` to get the percentile from omega
- `print(mors.Percentile(Mstar=1.0, percentile=50.0))` to get omega from the percentile


nichollsh (approved):
Looks good - thanks @lsoucasse. I think renaming the parameter to pctle definitely makes more sense.

EmmaPostolec:
Sorry, I saw this PR too late. Thanks for all the changes @lsoucasse, it makes much more sense with the new variable names.

PR #203, opened 2024-10-01
Config implementation

stefsmeets:
This PR updates the code to use the new style config. To help with the transition, I added a compatibility layer with the old style config, so that a dict lookup (like `OPTIONS['...']`) will still work for the most part, but the idea is to completely move over to the new format.

I added a superannoying warning to help with motivation :smiling_imp: 

I also updated all the old input tomls to the new format.

The most tricky bits are around the years, which changed from yr -> Gyr, and modules, which changed from integers to enums. Please pay extra attention to those.

Closes #74

### TODO

- [x] Get dummy test to pass
- [x] Get other tests to pass
- [x] Add some tests for config reader
- [x] star age changed (yr -> Gyr), baraffe and spada expect different conversions
- [x] Test non-dummy modules (zephyrus, spider, mors, janus, ...)

PR #226, opened 2024-10-23
Fix deprecation warnings from config update

### LINKED ISSUE #209: Fix deprecation warnings from config refactor
Follows up from #203 

There are a bunch of lookups that still need to be fixed. Some take more than a find/replace.

stefsmeets:
This PR addresses some of the deprecation warnings introduced by #203. So far I fixed all the ones in the dummy output.

Closes #209  

### Todo

- [x] Add tests for calliope outgas converter
- [x] Refactor other outgas/vol included settings to new converter
- [x] Test with default and other configs


nichollsh (approved):
Looks good! 

I ran 30 iterations of the default configuration on my computer and it didn't show any warnings relating to the old config. I also can't find any relevant instances in the code where it's accessed like a dictionary, so we can probably go ahead with removing the compatibility layer. The only places where it accesses the config in the old way is with the run offline chemistry stuff, which needs to be entirely overhauled anyway.

stefsmeets:
Cool, I'm off the rest of the week so I will pick this up later 😅

timlichtenberg:
Awesome, see you back soon! :-)

PR #227, opened 2024-10-23
 Implementation of Rxuv computation for escape

EmmaPostolec:
Draft pull request for escape changes


EmmaPostolec:
Hi all, 
So, as @nichollsh suggested during the last PROTEUS meeting, I choose to extract p_xuv and z_xuv from the atmos_clim directory. I made some changes in the files in this directory (janus.py, agni.py and wrapper.py) to store those values in hf_row["p_xuv"] and hf_row["z_xuv"]. 

To get p_xuv, I use the function find_nearest() to get the pressure value (in Pascal) in the atmosphere as close as Pxuv choose by the user (in bars). I used the same index to find the corresponding scale height, z_xuv. 
To compute R_xuv, I do R_xuv = hf_row["R_int"] + hf_row["z_xuv"].
I modified the GetHelpFileKeys() function in utils/coupler.py to add z_xuv and R_xuv in the runtime_helpfile.csv file when running a simulation. 

I am currently running some test with default.toml and escape.toml as inputs file. In both cases, I use Janus for the atmospheric module and Pxuv is taken in the escape config file (so I am not sure using default.toml is releveant in this case ? or should I include Pxuv in the default.toml config file ?). In both cases, z_xuv and R_xuv decrease with time. Here are 2 screenshots of this simulation. The red column on the left is the escape rate [kg s-1], the black column is the scale height z_xuv [m] and the blue column on the right is R_xuv [m].

I will need to try to run this with AGNI to be sure this is working fine. Please let me know what you think about the latest changes :) 

Screenshot 2024-11-06 at 4 26 28 PM

nichollsh:
Hi @EmmaPostolec I made two small changes to this branch to (hopefully) fix aragog and also to remove the VULCAN submodule (this should've already been in the gitignore but I had forgotten to add it previously).

Sorry to directly edit the branch, but I thought it would be easier to do this because these aren't changes associated with your new additions.

nichollsh (changes requested):
Nice work @EmmaPostolec!

I left a couple of comments, but there are a few other things:
* You need to add the new configuration parameters into the other configuration files in the `input/` folder.
* The dummy atmos_clim module needs to return values for `p_xuv` and `z_xuv` too. This is probably best handled by setting `p_xuv` to the surface pressure and `z_xuv` to zero in the Dummy case.
* You will need to update the test data. I can help with this if you want.
* You also need to fix the code style, since it's complaining about that as one of the PR checks.

EmmaPostolec:
Hi @nichollsh, thanks a lot for your feedback! In my latest commits, I fixed the following:

- You need to add the new configuration parameters in the other configuration files in the input/ folder. ✅
- The atmos_clim dummy module should also return values ​​for p_xuv and z_xuv. It is probably better to fix this by setting p_xuv to surface pressure and z_xuv to zero in the dummy case. ✅

I still need to implement the test and fix the code style problem. I will try to do it tomorrow. Maybe we can discuss it together if you have time?

nichollsh:
You can see the code style issue by viewing the output of the workflow here:
https://github.com/FormingWorlds/PROTEUS/actions/runs/11724157887/job/32657418698?pr=227

I think it's just a case of changing the order of the imports?

EmmaPostolec:
Yes indeed, it was just an import issue with the order of the functions. I just fixed the problem. Thanks! :)

EmmaPostolec:
Next (and final step I hope) is to update the test data files before merging 🤓

EmmaPostolec:
- To fix soon : Running dummy.toml simulation does not work → an error is displayed, and the simulation does not start :
```
/Users/emmapostolec/miniconda3/envs/proteus/lib/python3.12/site-packages/calliope/solubility.py:151: RuntimeWarning: divide by zero encountered in log10
  ppmw = 10 ** (-0.738 + 0.876 * np.log10(p) - 5.44e-5 * p_total)
/Users/emmapostolec/miniconda3/envs/proteus/lib/python3.12/site-packages/calliope/solve.py:458: RuntimeWarning: invalid value encountered in scalar divide
  outdict[s+"_vmr"] = outdict[s+"_bar"]/outdict["P_surf"]
```

nichollsh (approved):

EmmaPostolec:
@timlichtenberg I think this is ready for merging :)

src/proteus/utils/coupler.py
```diff
@@ -270,7 +270,7 @@ def GetHelpfileKeys():
             "bond_albedo", # bond albedo [1]
 
             # Escape
-            "esc_rate_total", # [kg s-1]
+            "esc_rate_total", "z_xuv", "R_xuv", # [kg s-1], [m], [m]
```
nichollsh, line 273: Add `p_xuv` here

nichollsh (changes requested):

nichollsh (approved):

timlichtenberg (approved):
Looks good to me!

timlichtenberg:
@lsoucasse We did two reviews now, seems good, perhaps you could have a look at the rough overall changes, btu we'll discuss it anyway next week in the PROTEUS meeting.

EmmaPostolec:
https://tenor.com/view/merge-pull-request-github-meme-pr-gif-8902990345370093805

lsoucasse:
> @lsoucasse We did two reviews now, seems good, perhaps you could have a look at the rough overall changes, btu we'll discuss it anyway next week in the PROTEUS meeting.

I was about to approve the PR too. Good job @EmmaPostolec.
One thing you  might think of is that now janus and agni expect zephyrus to be activated while one might want to run these atmosphere models without escape. It would be worth adding an "if zephyrus activated" statement to the Pxuv computation, and if not fill the Pxuv entry with the surface pressure. I think it kind of goes with issue #232.

timlichtenberg:
This seems like a good suggestion for a quick new fix/PR @EmmaPostolec :)

nichollsh:
Ah yes good catch - thanks @lsoucasse!

EmmaPostolec:
Thanks a lot for the suggestion @lsoucasse ! I opened a new issue (see #258 ) and I will try to do this today or tomorrow :)

PR #228, opened 2024-10-23
Download spectral files as required

### LINKED ISSUE #173: Download spectral files as required
Some spectral files are automatically downloaded and stored in the `$FWL_DATA` directory. However, this does not necessarily mean that the spectral file that you *want* is downloaded. It would be ideal if PROTEUS automatically downloaded the appropriate file from OSF if it cannot find it.

This is connected with #74, since we could have the configuration file specify a required spectral file "group" (e.g. Frostflow) and "resolution" (e.g. 256) rather than a file path which may or may not exist.

nichollsh:
Rework CLI for getting spectral files. Only downloads the spectral file required for the config that was provided (+Frostflow4096 for postprocessing). Closes #173.

Minor things:
* Tweak axis on population plot
* Access config with attribute rather than with key


lsoucasse (approved):
Nice! Also fixes some config warning.

nichollsh:
Thanks @lsoucasse! Apologies for the PR "spam".

Issue #215, opened 2024-10-19
Rock vapour calculation


nichollsh:
Currently, PROTEUS only supports volatile outgassing, but at higher temperatures we should expect significant amounts of rock vapour to be produced. Integrating these into the radiative transfer would be complex, but they could quite easily be incorporated into calculations of MMW and bulk density.

We can calculate vapour partial (surface) pressures at a given temperature and fO2 using codes such as Vaporock and Lavatmos. These codes are somewhat annoying to setup, since they depend on ThermoEngine, but I have been successful in doing this on my Mac laptop.

We would have to think carefully about how to include these gases alongside the currently supported volatiles.

timlichtenberg:
Agree, we had similar issues in Björn Koops' BSc thesis earlier this year when trying out Vaporock. At the meeting on Wednesday, we should discuss whether pre-computed look-up tables may provide an entry. 

Why do you think that integrating them into the radiative transfer would be complex? If we have the opacities for the major gases, SOCRATES should be able to compute k-tables?

nichollsh:
Pre-computing this is a nice idea, particularly if we are assuming BSE composition since that would only leave 2 variables (Tsurf, fO2) for the table's axes.

I think it might be complex because SOCRATES doesn't currently support any refractory species. It uses hardcoded gas IDs defined in [gas_list_pcf.f90](https://github.com/nichollsh/SOCRATES/blob/4d7ec2afe8b26671987aa1a678362c6834b8e19b/src/radiance_core/gas_list_pcf.F90#L26), so we would need to manually add all of the required gases to this list. This is easy to do, but it would mean diverging from the Met Office SOCRATES distribution somewhat.

nichollsh:
The currently supported refractory species are: TiO, VO, Na, K, FeH, CrH, Li, Rb, Cs

timlichtenberg:
Mh, yeah, these are clearly not sufficient. But yes, perhaps assuming BSE and condensing the axes would make sense. I wonder if the axes should be (Tsurf, Psurf, fO2), but that's a thing to discuss on Wednesday with Yamila's group further.

nichollsh:
Notes from meeting with Yamila's group:
* Lavatmos can calculate rock vapour partial pressures - can probably use a look up table to do this easily.
* Installing lavatmos might be a pain - depends on ThermoEngine.
* We can probably integrate this into CALLIOPE directly.
[...]

PR #229, opened 2024-10-23
Support for arbitrary gases

### LINKED ISSUE #224: Support arbitrary gases
Before tacking #215, we should consider implementing adding support for arbitrary gases. Implementing this would make working on #215 much easier.

This would mean MMW calculations, plot colours, semantic changes to labels, etc.

nichollsh:
Introduces support for arbitrary gases (closes #224). This is mostly in preparation for #215. Several of these changes might need to be reflected on the CALLIOPE side, but we can think about this later.

**Outline**:
- Refactor labels and variables to refer to 'gases' rather than 'volatiles'.
- Remove `input/_stars.txt` since this is now included in the docs.
- Allow for gases beyond those handled by CALLIOPE.
- Allow for gases beyond those marked as included/excluded in the config.
- Correction to transit radius calculation (`R_observed = z_obs + R_int`). There was some inconsistency throughout the code on this, but it should be consistent now.
- Systematically generate colours for plotting, so we don't need to specify a colour for every single gas. This would become infeasible if we start using fastchem which will speciate the atmosphere into lots of gases.
- Systematically generate labels for plots.
- Support for refractory elements when calculating metallicity and other quantities.
- Fixed `grid_proteus.py` although we should consider reworking this in the future (#204, #162).
- Directly handle cases where atmosphere escapes (`P_surf=0`) as this was causing divide-by-zero errors, resulting in the helpfile CSV missing some columns in the last row.


lsoucasse (approved):
Great! I like the new get_color function. Also having the zero pressure bug fixed is a good thing.

nichollsh:
Thanks @lsoucasse!

Issue #167, opened 2024-09-02
Harmonize access to Mors and stellar object


lsoucasse:
Getting the bolometric flux and the star radius with Mors and the Spada model (case 0) does not seem to take the rotation rate as an input parameter as it should. Also, it would make more sense to harmonize the access to stellar data between submodules (i.e. the bolometric flux for the atmospheric modules is get with the Value function  while the XUV flux for the escape module is get through the Track function).

This issue can be addressed together with designing a wrapper for stellar modules #126 .

lsoucasse:
This is also important now we use a fix time stepping to access Spada data. We need to make sure to create star object only once.

lsoucasse:
I would like to address this issue but the current state is very confusing to me:
- Why do we currently set up a star radius as an input parameter in the config while it is actually an output of Mors (either Spada/Baraffe data)? same remark with lbol_now and Teff?
- When the get_new_spectrum function is called, it takes some star properties computed at the init step (star radius, various fluxes...) that are never updated in time. Is is the normal behaviour?

I was thinking of deleting the get_property function in the mors synthesis module and instead generating the whole track data on the proteus side and get the properties from there as time evolves. Does that sound good @nichollsh ?

nichollsh:
A few of these variables are the *modern* values - i.e. those observed at the current time. These are needed for knowing how to scale the modern spectrum backwards in time. This is related to the properties which are only calculated at the init step - we need to know what the star looks like now, in order to know how to scale the modern (observed) spectrum backwards in time.

The Teff and radius parameters are also needed for the dummy star module, although we could feasibly move those into `config.star.dummy`.

lsoucasse:
OK, I understand these are frozen. But I think it would be better to extract them from the whole track data otherwise it will not be consistent. Workflow would be:
- create the star track data from mass and rotation rate
- extract properties at initial time (modern properties if I get it correctly) for spectrum synthesis
- extract properties as time evolves to be stored in helper file and used by other submodule (instellation, xuv flux for escape...)
Would that be OK?

nichollsh:
Yeah I think that makes more sense. 
However, for the dummy star module we will still need to provide radius and Teff, so those can go into `config.star.dummy` I suppose.

lsoucasse:
> Yeah I think that makes more sense. However, for the dummy star module we will still need to provide radius and Teff, so those can go into `config.star.dummy` I suppose.

[...]

PR #230, opened 2024-10-24
Star wrapper consistency

lsoucasse:
This work restructures attributes of the star config and consistently uses Mors/Spada when this model is triggered.
- The config now better displays which parameters are used by what (dummy/mors) although this could be further refined between mors/baraffe and mors/spada
- When using Mors/Spada, the whole track (time evolution) data are generated for given star mass and rotation percentile and star properties (luminosity, radius, Teff) are computed from there.  Modern star properties for spectrum synthesis are also computed from the track which required little adaptation (and new release) of the mors package.
- These data are used consistently by the zephyrus escape model.


lsoucasse:
This PR works together with Mors PR [here](https://github.com/FormingWorlds/MORS/pull/24).

The idea for the review is to approve the mors one first, then do a release and update proteus dependencies before merging this one.

nichollsh:
Presumably you will need to re-run the tests on this one once merging the MORS changes. I can imagine that taking the 'modern' values from the tracks might yield different results compared to using the user-provided ones.

This is not to say it's wrong - I think it's more consistent - but maybe the test data will change.

nichollsh (approved):
Nice work - looks good. I think it's nice to unify the way the two tracks are treated. I agree with the way the input parameters have been re-organised. Does this change the status of #167?

Feel free to merge this once the MORS one is merged - the tests are passing on that branch now.

lsoucasse:
> Presumably you will need to re-run the tests on this one once merging the MORS changes. I can imagine that taking the 'modern' values from the tracks might yield different results compared to using the user-provided ones.
> 
> This is not to say it's wrong - I think it's more consistent - but maybe the test data will change.

Thanks @nichollsh,

Actually the dummy test uses baraffe tracks so it will not be affected.  I ran the dummy test with spada data and got very similar results which gives some confidence in the implementation. Maybe worth to set a new test soon that uses spada data together with escape.

I am expecting little changes in the results with this new implementation. The individual euv flux values may change significantly in using the whole track, but we can hope that the ratios we are actually using are not. On the other hand, the bolometric flux is not altered.

lsoucasse:
> Nice work - looks good. I think it's nice to unify the way the two tracks are treated. I agree with the way the input parameters have been re-organised. Does this change the status of #167?
> 
> Feel free to merge this once the MORS one is merged - the tests are passing on that branch now.

Yes it is closing the issue.

lsoucasse:
Made a new Mors release and updated pyproject.toml requirements. Could you approve this again @nichollsh ?

nichollsh (approved):

PR #234, opened 2024-10-25
Fix unit conversion in dummy star and add test

### LINKED ISSUE #233: Fix unit conversion issue for dummy star and add test
The stellar radius used to calculate the dummy stellar spectrum is currently passed in units of R_sun, but it should be in units of metres. This is a simple mistake, but we should try to make sure that these things don't come up so easily.

We should fix this error and add a test for this component of the code.

nichollsh:
Fixes unit conversion error on stellar radius provided to dummy star module.

Updates the test to use the dummy config, which runs a few iterations of PROTEUS with this module. I've tried to configure the cfg file for this test to touch on different bits of code compared to the existing dummy test. It only increases coverage to 47%, but we are now testing a lot more of the model outputs (i.e. everything related to the star wrapper and the dummy_star module). The test checks for:
* Helpfile values
* Statusfile content
* Stellar spectrum file content

Closes #233


src/proteus/star/wrapper.py
```diff
@@ -130,7 +130,7 @@ def get_new_spectrum(t_star:float, config:Config,
     # Dummy case
     if config.star.module == 'dummy':
         from proteus.star.dummy import generate_spectrum
-        wl, fl = generate_spectrum(config.star.dummy.Teff, config.star.dummy.radius)
+        wl, fl = generate_spectrum(config.star.dummy.Teff, R_star)
```
lsoucasse, line 133: I removed passing R_star as an argument in my last PR as it is a constant for the dummy module... but I missed the unit conversion.

I recommand not passing R_star as an argument but rather replace l133 with
`wl, fl = generate_spectrum(config.star.dummy.Teff, config.star.dummy.radius * Rsun)`

lsoucasse (changes requested):
Hi @nichollsh,
I suggest not passing constant argument and correcting rstar unit within the function instead.
Also to me, I would rather use the dummy star test as the dummy test (as it covers all dummy modules). Hopefully we can set up soon a new integration test with together janus/mors/zephyrus/calliope/aragog.

nichollsh:
Thanks @lsoucasse. I'll rework the test with this in mind.

nichollsh:
@lsoucasse can you check this now please?

lsoucasse (approved):
Great! Now we have a "full dummy" integration test. Thanks @nichollsh.

PR #235, opened 2024-10-25
Aragog constant

lsoucasse:
This work introduces Aragog as an alternative module for interior modelling.

It uses constant physical properties and runs fine athough the coupled results seem to not behave properly, see [here](https://github.com/FormingWorlds/PROTEUS/issues/62#issuecomment-2431755202).

I suggest we create new issues to debug coupled behaviour and expand from this current state (add config parameters, use look up tables, plot results).


lsoucasse:
I am off next week so do not hesitate to change/approve/merge in my absence @planetmariana @timlichtenberg . I also add you @nichollsh as reviewer as you already managed to run this branch.

timlichtenberg:
Thanks Laurent! All three of us @planetmariana @nichollsh and me should have a look through this and add a review. Once that is done, we can merge it in the next days.

nichollsh:
This is what I get from the dummy config when replacing the dummy interior with aragog. I think this is generally looking okay - the planet at least cools down and the melt fraction decreases. 

plot_global_log

nichollsh (approved):
Personally, I am happy for this to be merged. Nice work @lsoucasse. I agree that the next steps of the implementation can be handled by another PR.

Issue #236, opened 2024-10-25
Aragog submodule


lsoucasse:
We now have a basic version of Aragog implemented within Proteus. This issue lists the next step towards an actual replacement of C-Spider.

- [ ] Debug coupled interactions as it seems to misbehave, see [here](https://github.com/FormingWorlds/PROTEUS/issues/62#issuecomment-2431755202). We could also compare Spider and Aragog on a single run to gain insight on what is going on here @planetmariana.
- [ ] Write plotting routine to get similar output as Spider
- [ ] Use look up tables for thermophysics properties downloaded from the OSF ExPlanetology/aragog#7
- [ ] Introduce core cooling boundary condition ExPlanetology/aragog#8 
- [ ] Set up an integration test with core modules (Janus/Mors/Zephyrus/Caliope/Aragog)

lsoucasse:
Also I realize I did not implement the "attempt" loop as in spider, where tolerance/dt is adjusted when solver fails. Maybe it can help.

nichollsh:
I have made some progress on the second point. See the plot below. We can think about merging these changes into aragog and PROTEUS once the current aragog-related PR is merged.

<img width="1120" alt="image" src="https://github.com/user-attachments/assets/6449b8cf-8f79-4830-adfb-28c78507b0d6">

PR #239, opened 2024-10-28
Read aragog output files and make interior plots

nichollsh:
Calls the `Output.write_at_time()` function in Aragog, which will write a NetCDF file at each PROTEUS iteration - this means we will have 1 interior and 1 atmosphere NetCDF file at each step. Generalised the manner in which these data are accessed between SPIDER and Aragog, so we don't have duplicated plotting code. Added functionality to read and plot Aragog model results when coupled to PROTEUS. This will complete one of the tasks in https://github.com/FormingWorlds/PROTEUS/issues/236

Requires the corresponding Aragog branches to be merged first.

Bump Mors version - support for Python 3.13


nichollsh:
Encountered a slightly odd issue with the Ruff linting. Running `ruff check --fix` locally produces import blocks which fail on the CI. Solution was to just manually adjust the whitespace in the import blocks until the CI was happy.

nichollsh:
Hi @timlichtenberg - could you review this please? If you want to test it, you will need to use my aragog branch: https://github.com/ExPlanetology/aragog/tree/hn/writefiles

timlichtenberg (approved):
Looks good to me! My only overall worry is to retain capabilities for SPIDER, once Aragog is fully integrated we have to do tests to maintain this.

PR #243, opened 2024-10-29
Atmosphere model wrapper

### LINKED ISSUE #238: Updated atmosphere interface parameters
The following parameters should be added to the config file in the AGNI section, since they are modelled by AGNI and could be interesting to consider more deeply
- `surf_material`, the type of material used to set the surface albedo
- `solve_atol` and `solve_rtol`, tolerances on the solver solution

The surface albedo parameter that currently exists should be moved into the Janus section. 
The existing parameter `albedo_pl` should yield a warning when used alongside clouds or rayleigh scattering.

We should also consider storing the Bond albedo of the planet in the helpfile. Taking `bond_albedo = F_sct / F_ins` is not correct, since `F_sct` is scaled by the zenith angle while `F_ins` is not. This is an easy mistake to make and will yield significantly incorrect values for the bond albedo of the planet. It should instead be calculated as `ds["fl_U_SW"]/ds["fl_D_SW"]` where `ds` is the atmosphere Dataset generated by AGNI or JANUS.

nichollsh:
* Closes #238.
  - Support passing surface albedo file names to AGNI, allowing spectrally-dependent surface single-scattering properties. 
  - Support passing solver tolerance to AGNI from config files. 
  - Store Bond albedo in output helpfile.
* Fix bug where incorrect tropopause option was being passed to JANUS.
* Warn user when setting configuration `albedo_pl>0` in physically invalid cases.
* Updated documentation.


input/default.toml
```diff
@@ -93,7 +93,7 @@ author  = "Harrison Nicholls, Tim Lichtenberg"
     module  = "mors"
     [star.mors]
         rot_pctle = 50.0    # rotation percentile
-        tracks  = "spada"   # evolution tracks: spada | baraffe
+        tracks  = "baraffe"   # evolution tracks: spada | baraffe
```
timlichtenberg, line 96: Should baraffe be the default?

input/default.toml
```diff
@@ -145,11 +145,14 @@ author  = "Harrison Nicholls, Tim Lichtenberg"
...
+        surf_material   = "greybody"    # surface material file for scattering
+        solution_atol   = 1e-3          # solver absolute tolerance
+        solution_rtol   = 2e-2          # solver relative tolerance
 
     [atmos_clim.janus]
-        p_top           = 1.0e-6        # bar, top of atmosphere grid pressure
+        p_top           = 1.0e-7        # bar, top of atmosphere grid pressure
         spectral_group  = "Frostflow"   # which gas opacities to include
-        spectral_bands  = "256"         # how many spectral bands?
+        spectral_bands  = "16"         # how many spectral bands?
```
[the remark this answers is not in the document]
nichollsh: Ah, no, probably not. I'll change this back.

input/default.toml
nichollsh: I'll change this back too.

timlichtenberg (approved):
Looks good, very nice to have the surface scattering in there now! We should talk about how/when to make use of it at some point.

nichollsh:
Yes definitely! I added it for the purposes of observables, but maybe it can also factor into their evolution through the surface energy balance.

Issue #193, opened 2024-09-26
Oscillatory behaviour of fluxes and surface temperature


nichollsh:
The model sometimes exhibits oscillatory behaviour in `F_atm`, `F_int`, and `T_surf`. This is probably not physical, since it occurs when using the dummy modules. It also occurs when using SPIDER+JANUS.

We should investigate this further. My guess is that it is associated with the time-stepping scheme, which is currently quite crude.

Example from the dummy model in the image below. The behaviour looks particularly bad in this plot because of the log scale on the y-axis. We can see from the OLR lines that the magnitude of the oscillations is quite small (in this case, at least).
![image](https://github.com/user-attachments/assets/599b13ad-177b-48c2-a712-47b1c1908949)

timlichtenberg:
Is this still relevant with the current last version?

nichollsh:
Yeah this is still an issue. It doesn't typically cause problems physically, but we should at least look into using a better time-stepping scheme in terms of model efficiency when the melt fraction is small. Maybe there are techniques that can be borrowed from ESM codes? Somewhat connected to https://github.com/ExPlanetology/aragog/issues/21.

I would suggest we look into this soon, because it could potentially yield significant performance benefits.

nichollsh:
Forcing the model to take large steps shows the following behaviour. This is what we'd expect given the logic above.
<img width="570" alt="image" src="https://github.com/user-attachments/assets/7e4f1340-30ca-437f-833d-f34283fd84ce">

PR #247, opened 2024-11-02
Tests using physical simulations

### LINKED ISSUE #237: Set up an integration test with core modules (Janus/Mors/Zephyrus/Caliope/Aragog)

nichollsh:
We currently have tests using the Dummy simulations. This checks that the module coupling works okay and the dummy modules are working.

It is absolutely necessary to add tests for the physical modules. This PR introduces some initial tests for these, although they will surely be subject to change in the future. It runs PROTEUS with JANUS+Aragog+MORS+CALLIOPE for a few model steps. It does not run until convergence as this would take too long. Workflow runtime is about 5 minutes. Closes #237.

This **increases the coverage from 45% to 59%**. We could get this to ~65% if we included an AGNI test, but this would probably be quite slow.

Implementing these physical tests required adding installation steps for SOCRATES and Aragog to the CI workflow. These use caching to speed things up.

Includes a few bug fixes:
* Do not plot modern spectrum on Dummy stellar flux plots
* Do not try to copy the modern spectrum when using Dummy stellar model

GitHub records that a lot of files were changed, but most of these are changes to the test data. Apologies for the many random commits - it took some work to get the CI workflow working properly. The key thing with this is the order in which the modules are checked-out.


lsoucasse (approved):
This is really great! Thanks @nichollsh.

We should consider to extend this test to a resume simulation for a later stage of planet evolution (again, for a couple of iterations). It  will not increase the coverage but check code behaviour in another range of parameters.

nichollsh:
Yes I definitely agree. I think particularly once we have also looked into #193.

nichollsh:
Thanks @lsoucasse!

PR #249, opened 2024-11-05
Add gas overlap option to config

### LINKED ISSUE #248: Pass gas overlap method via configuration file
At the moment, JANUS defaults to using random overlap to handle overlapping absorption of multiple gases with a given spectral band. AGNI uses equivalent extinction as its default. This is probably fine, but could easily lead to differences in some cases.

We could add functionality to JANUS to allow this to be stored in the atmosphere object, which is then passed to Cl_run_cdf with the `-g` flag. AGNI already allows this to be set by the `setup!()` function.

- [x] Update JANUS
- [x] Update AGNI
- [x] Update PROTEUS

nichollsh:
This PR adds the option of changing the gas overlap method via the configuration file. Closes #248.

Requires JANUS to be updated to 24.11.05 and AGNI to be on the latest version (as of 5 November).

### DIFF
--- a/input/aragog.toml
+++ b/input/aragog.toml
@@ -146,7 +146,7 @@ author  = "Harrison Nicholls, Tim Lichtenberg"
         surf_material   = "greybody"    # surface material file for scattering
         solution_atol   = 1e-3          # solver absolute tolerance
         solution_rtol   = 2e-2          # solver relative tolerance
-        overlap_type    = "ee"          # gas overlap method
+        overlap_method    = "ee"          # gas overlap method
 
     [atmos_clim.janus]
         p_top           = 1.0e-6        # bar, top of atmosphere grid pressure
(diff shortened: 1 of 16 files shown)


lsoucasse (changes requested):
Thanks @harrison. This looks good. I would suggest to add this new config option in the doc, also clarify what "ro", "ee" and "rorr" stand for. This gas overlap flag changes the way radiative properties are computed by Socrates ?

nichollsh:
The k-terms in the spectral files are computed with each gas treated separately. However, obviously with multiple gases in the atmosphere they might have overlapping opacities in shared spectral regions. There are various methods to handle this problem of combining k-terms. Three of these are random overlap (RO), equivalent extinction (EE), and RO with resorting+rebinning of k-terms (RORR). There are other methods supported by SOCRATES, but these other methods are fairly niche.

I've added docstrings to the files in `src/proteus/config/` so in theory the documentation should update automatically. However, I agree it's probably good to clarify what these abbreviations mean.

lsoucasse:
> The k-terms in the spectral files are computed with each gas treated separately. However, obviously with multiple gases in the atmosphere they might have overlapping opacities in shared spectral regions. There are various methods to handle this problem of combining k-terms. Three of these are random overlap (RO), equivalent extinction (EE), and RO with resorting+rebinning of k-terms (RORR). There are other methods supported by SOCRATES, but these other methods are fairly niche.
> 
> I've added docstrings to the files in `src/proteus/config/` so in theory the documentation should update automatically. However, I agree it's probably good to clarify what these abbreviations mean.

Thanks for clarifying this.

lsoucasse (approved):

PR #254, opened 2024-11-12
Update aragog wrapper with recent changes

nichollsh:
* Fixes issue with `core_radius` keyword
* Fixes issue with accessing rheological front as float (not as array)
* Updated documentation.

Still need to add aragog as a pyproject dependency - should be handled in another PR.


EmmaPostolec:
Nice for the update of the doc !

EmmaPostolec (approved):

PR #255, opened 2024-11-12
Add aragog as a package dependency.

lsoucasse:
Add the aragog package as a package dependency.

We were using some dummy data in the aragog data test folder that I replaced with actual data from the OSF. It changed the test data that I updated. First step towards using full lookup table data @planetmariana .


nichollsh:
@lsoucasse I have just made a release to PyPI for aragog which includes the fixes for the latest version - this is then the same version as used in Emma's escape branch.

src/proteus/interior/aragog.py
```diff
@@ -141,8 +145,8 @@ def SetupAragogSolver(config:Config, hf_row:dict):
             latent_heat_of_fusion = 4e6,
             rheological_transition_melt_fraction = 0.4,
             rheological_transition_width = 0.15,
-            solidus = "aragog/data/test/solidus_1d_lookup.dat",
-            liquidus = "aragog/data/test/liquidus_1d_lookup.dat",
+            solidus = FWL_DATA_DIR / "interior_lookup_tables/1TPa-dK09-elec-free/MgSiO3_Wolf_Bower_2018/solidus.dat",
```
nichollsh, line 148: Is this a different set of melting curves? Uploading them to OSF makes sense, but the interior evolution looks different as a result of this PR - maybe this is why?

nichollsh (changes requested):

src/proteus/interior/aragog.py
nichollsh: If we are switching to the Wolf & Bower curves, that's fine with me. I just want to confirm that this was intentional.

nichollsh: Ah - I see that you describe this in the body of the PR. Apologies for not reading this properly.

planetmariana: yes, the previous solidus_1d_lookup.dat was just a simple melting curve with two data points for testing. MgSiO3_Wolf_Bower_2018/solidus.dat is the new melting curve translated from entropy into temperature up to 1000GPa.

PR #257, opened 2024-11-12
Calculate and store orbital period

### LINKED ISSUE #252: Calculate and store orbital period
In order to implement a tidal heating model (issue #190 for now), we need to know the period of the planet's orbit. 

This can be easily achieved for an elliptical orbit using the parameters which are already set in the configuration file: star mass, semi-major axis, planet mass, and orbital eccentricity.

There's an equation for calculating this on Wikipedia: https://en.wikipedia.org/wiki/Elliptic_orbit

nichollsh:
The model now calculates orbital period from semimajor axis, star mass, and planet mass. This is then stored in the helpfile, although it doesn't currently evolve with time. Closes #252.

This will be useful when implementing tidal heating calculations, as the heat dissipated by tides depends on the frequency of the perturbation and the response of the planet. The period depends on the planet mass because we are accounting for orbital eccentricity, which could potentially evolve over time due to tides.

Recovers 'true' values well:
* Earth
   * Calculated 365.2565 days
   * Actual 365.2564 days
* Mercury
   * Calculated 87.9689 days
   * Actual 87.9691 days


lsoucasse (approved):
Great! Proteus is expanding!

nichollsh:
Thanks, Laurent!

PR #262, opened 2024-11-14
Self-consistent mass and radius for interior modules

### LINKED ISSUE #253: Self-consistent mantle and core mass with interior modules
I've realised that PROTEUS contains two different mantle masses, although we are treating them as one. There's the mantle mass derived  from the planet mass provided in the config file, and there's the mantle mass calculated by SPIDER (and Aragog?) which is derived using a hydrostatic calculation from the provided interior radius.

We should make sure that we use a consistent mantle/core mass. The best option in lieu of a true structure code is to use Aragog/SPIDER to do the hydrostatic integration, and use an input parameter for the Dummy interior module.

The question is then: do we prescribe an interior mass, or an interior radius? The latter is much easier since it's already an input to SPIDER/Aragog. However, the former is more physical since it is the mass that is a conserved quantity, not the radius. We need to think carefully about this.

Connected to #50.

nichollsh:
The current iteration of the code requires the user to specify interior mass and radius in the configuration file. However, interior mass is calculated as part of the Aragog/SPIDER/Dummy step anyway. This means that allowing the user to input the interior mass leads to two physically-inconsistent values. This is currently only physically consistent with using planetary masses and radii near Earth values, since then the SPIDER solution will happen to be identical to that provided by the user.

The simplest solution to this is to use the interior model to calculate the interior radius based on the inputted interior mass. This PR achieves this and closes #253.

Because Aragog currently uses constant physical properties, it does not make a very good structure model, but it's still more physically consistent this way. SPIDER (and even the Dummy module) does a good job in this regard.

Minor changes:
* Provide aragog with initial temperature specified in config file
* Enable aragog core cooling boundary condition
* Configuration file for K2-18 b
* All SPIDER options are passed to SPIDER via spider.py, not through the .opts file. This is to ensure consistency with Aragog (in that all parameters are set via the Python code).
* Reworked `plot_stacked` into `plot_structure` which includes the core temperature, `R_xuv`, and `R_obs`. This also (importantly) plots the entire profile on the same radius scale, so it's clearer what the structure of the planet looks like.
* Does not run atmosphere module when atmosphere has already escaped, and goes straight to checking the break criteria.
* Do not allow **all** gases to be set as condensible when running AGNI, as this can lead to sum(VMR)=0 in some regions, which causes the model to crash.

Requires https://github.com/ExPlanetology/aragog/pull/29 to be merged first.

Next steps:
* Consider more realistic structure model
* Allow user to input radius instead of mass, which avoids this structure calculation entirely.


nichollsh:
@timlichtenberg this is now ready for review I think. There are lots of changed files, but this is mostly a result of the changed tests. It makes the changes we discussed in the meeting earlier, plus a few bug fixes that I discovered in the last week or so.

input/default.toml
```diff
@@ -115,9 +115,8 @@ author  = "Harrison Nicholls, Tim Lichtenberg"
 # Planetary structure - physics table
 [struct]
 
-    # Enclosed mass and radius at the atmosphere-interior interface
+    # Enclosed mass at the atmosphere-interior interface
```
timlichtenberg, line 118: Should this be "enclosed mass at the top of atmosphere" according to our discussion earlier on?

nichollsh: Ah yes - I forgot to make this change! It should indeed. How about I make a new issue for this to treat it separately, since this PR is already too large? We can then also add the option of passing the radius instead of the mass.

src/proteus/interior/aragog.py
```diff
@@ -91,10 +90,10 @@ def SetupAragogSolver(config:Config, hf_row:dict):
     boundary_conditions = _BoundaryConditionsParameters(
             outer_boundary_condition = 4, # 4 = prescribed heat flux
             outer_boundary_value = hf_row["F_atm"], # first guess surface heat flux [W/m2]
-            inner_boundary_condition = 3, # 3 = prescribed temperature
+            inner_boundary_condition = 1, # 3 = prescribed temperature
             inner_boundary_value = 4000, # core temperature [K]
```
timlichtenberg, line 94:
```suggestion
            inner_boundary_value = 4000, # core temperature [K], if inner_boundary_condition = 3
```

src/proteus/interior/spider.py
```diff
@@ -258,11 +254,120 @@ def _try_spider( dirs:dict, config:Config,
...
+    call_sequence.extend(["-u238_half_life",       "4468E6"]) # years (Ruedas, 2017)
+
+
+
     # Runtime info
     flags = ""
     for flag in call_sequence:
         flags += " " + flag
-    log.debug("SPIDER call sequence: '%s'" % flags)
+    # log.debug("SPIDER call sequence: '%s'" % flags)
```
timlichtenberg, line 370:
```suggestion
    log.debug("SPIDER call sequence: '%s'" % flags)
```

timlichtenberg: If you comment this out, doesn't this make it very hard to debug?

src/proteus/interior/spider.py
```diff
@@ -376,13 +478,12 @@ def ReadSPIDER(dirs:dict, config:Config, R_int:float):
     output["RF_depth"]        = float(data_a[7])/R_int  # depth of rheological front
 
     # Manually calculate heat flux at near-surface from energy gradient
-    json_file   = MyJSON( dirs["output"]+'/data/{}.json'.format(sim_time) )
```
timlichtenberg, line 379: Where did this go?

src/proteus/proteus.py
```diff
@@ -229,12 +231,6 @@ def start(self, *, resume: bool = False):
             self.loops["init"] = self.loops["init_loops"] + 1
         log.info(" ")
 
-        # Create lockfile for keeping simulation running
-        self.lockfile = CreateLockFile(self.directories["output"])
```
timlichtenberg, line 233: Not necessary anymore?

src/proteus/proteus.py
```diff
@@ -229,12 +231,6 @@ def start(self, *, resume: bool = False):
...
-
-        # Download basic data
-        download_sufficient_data(self.config)
```
timlichtenberg, line 236: Same here?

src/proteus/utils/coupler.py
```diff
@@ -432,7 +433,7 @@ def UpdatePlots( hf_all:pd.DataFrame, dirs:dict, config:Config, end=False, num_s
         nc_times = [int(f.split("/")[-1].split("_atm")[0]) for f in ncs]
 
         # Check intersection of atmosphere and interior data
-        if not spider:
+        if dummy_int:
```
timlichtenberg, line 436: At next iteration I would suggest to make this maximally clear:
if dummy_int:
elif: spider
elif: aragog
else: throw error

tests/data/integration/physical/plot_global_lin.png
timlichtenberg, line 1: The planet is heating up and the melt fraction is increasing, this is a bit worrisome.

timlichtenberg (changes requested):
Looks very good overall, well done. I am a bit worried about the planet heating up and some omissions that are not clear to me where they went. Happy to have a quick later on again to merge asap or tomorrow morning otherwise.

input/default.toml
timlichtenberg: Good for me!

src/proteus/interior/spider.py
nichollsh: The code itself is still there - I just commented it out because we weren't actually using the values for anything. I left it in because it's useful for testing things sometimes, but there's no point using the extra resources to read the file if the values don't go anywhere.

src/proteus/proteus.py
nichollsh: This is still necessary, I've just moved it further up in the file. See response for download_sufficient_data.

src/proteus/proteus.py
nichollsh: This is still included in the code and is very much necessary. I had to move it further up in the file because it should be called before the initial structure calculation, because we need to make sure that the lookup tables are downloaded before trying to run Aragog.

tests/data/integration/physical/plot_global_lin.png
nichollsh: It's a little odd, but this is just the first 2000 years of evolution so it could be a numerical thing. This indicates that 2kyr isn't good enough for the test, so I'll extend it to a longer time-span.

nichollsh:
Thanks for the review Tim - this is really useful. I'll make these changes tonight/tomorrow, and if you could have a look again tomorrow then that would be great.

nichollsh:
@timlichtenberg please can you have another look at this again today? I have made the changes that you suggested and responded to the other comments above. I also had to update the test data.

I will then address this issue: https://github.com/FormingWorlds/PROTEUS/issues/267

timlichtenberg (approved):
Great!

PR #272, opened 2024-11-19
Set interior structure by: total planet mass, radius

### LINKED ISSUE #267: Allow user to set interior structure by radius as well as mass
With https://github.com/FormingWorlds/PROTEUS/pull/262, the user will be able to set the interior size by mass in a physically consistent manner. 

It would be useful if they could alternatively set it by radius too, such that mass is an output instead of an input.

### LINKED ISSUE #269: Offset input planet mass by volatile mass
With https://github.com/FormingWorlds/PROTEUS/pull/262, the user can specify a dry mantle mass as an input parameter, from which a mantle radius is calculated. It would be useful if, instead/alternatively, the user could specify the total planet mass.

This would mean subtracting-off the volatile inventory from the total input planet mass, and using the result (the dry interior mass) to solve for R_int as per the current formulation.

nichollsh:
Reworks the methods by which the user can set the interior structure. Previously, they would specify the dry interior mass, and the interior models would be used to inversely solve for the interior radius.

This is now offset by the total volatile inventory, such that the user only has to specify the total planet mass. This is what would be probed by RV or TTV measurements, so it is much easier to specify for the user. Closes #269.

Additionally, the user can avoid the whole mass -> radius conversion by setting the interior radius directly in the configuration file. Closes #267.

The physical test (which uses Aragog) now sets the structure using the planet radius (hence the test updates). The dummy test (which uses the Dummy interior) sets the structure using the total planet mass.

Minor things:
* AGNI has been updated to 0.11.0, which replaces the `reset_vmr` flag with the more obvious `rainout` flag. This PR includes a minor change to handle this accordingly.


input/aragog.toml
```diff
@@ -114,12 +114,11 @@ author  = "Harrison Nicholls, Tim Lichtenberg"
 
 # Planetary structure - physics table
 [struct]
-    mass        = 1.0       # M_earth
+    set_by      = 'radius_int'
```
timlichtenberg, line 117:
```suggestion
    set_by      = 'radius_int'    # Variable to set the interior structure: 'radius_int' or 'mass_tot'
```

input/dummy.toml
```diff
@@ -114,12 +114,11 @@ author  = "Harrison Nicholls, Tim Lichtenberg"
 
 # Planetary structure - physics table
 [struct]
-    mass        = 1.0       # M_earth
+    set_by      = 'radius_int'
```
timlichtenberg, line 117:
```suggestion
    set_by      = 'radius_int'    # Variable to set interior structure: 'radius_int' or 'mass_tot'
```

input/k218b.toml
```diff
@@ -114,12 +114,11 @@ author  = "Harrison Nicholls, Tim Lichtenberg"
 
 # Planetary structure - physics table
 [struct]
-    mass        = 8.63       # M_earth
+    set_by      = 'mass_tot'
```
timlichtenberg, line 117:
```suggestion
    set_by      = 'mass_tot'    # Variable to set interior structure: 'radius_int' or 'mass_tot'
```

input/trappist1c.toml
```diff
@@ -114,12 +114,11 @@ author  = "Harrison Nicholls, Tim Lichtenberg"
 
 # Planetary structure - physics table
 [struct]
-    mass        = 1.308       # M_earth
+    set_by      = 'mass_tot'
```
timlichtenberg, line 117:
```suggestion
    set_by      = 'mass_tot'    # Variable to set interior structure: 'radius_int' or 'mass_tot'
```

src/proteus/interior/wrapper.py
```diff
@@ -45,21 +43,23 @@ def determine_interior_radius(dirs:dict, config:Config, hf_all:pd.DataFrame, hf_
...
+    # We need to solve for the state hf_row[M_tot] = config.struct.mass_tot
     # This function takes R_int as the input value, and returns the mass residual
     def _resid(x):
         hf_row["R_int"] = x
 
         log.debug("Try R = %.2e m = %.3f R_earth"%(x,x/R_earth))
 
+        # Use interior model to get dry mass from radius
         run_interior(dirs, config, IC_INTERIOR, hf_all, hf_row, verbose=False)
         if solve_g:
```
timlichtenberg, line 57: Is there a case `solve_g` would be meaningfully turned off?

tests/data/integration/physical/plot_population_time_density.png
timlichtenberg, line 1: The simulation line is not visible anymore.

tests/data/integration/physical/plot_population_mass_radius.png
timlichtenberg, line 1: It's not clear where "simulation" is. Use a different color or thicker line style?

tests/integration/physical.toml
```diff
@@ -114,12 +114,11 @@ author  = "Harrison Nicholls, Tim Lichtenberg"
 
 # Planetary structure - physics table
 [struct]
-    mass        = 1.0       # M_earth
+    set_by      = 'radius_int'
```
timlichtenberg, line 117:
```suggestion
    set_by      = 'radius_int'    # Variable to set interior structure: 'radius_int' or 'mass_tot'
```

timlichtenberg (changes requested):
Some minor suggestions to improve readability, looks feasible otherwise.

src/proteus/interior/wrapper.py
nichollsh: No, I don't think so. Leaving it set to true works entirely fine for both Dummy and SPIDER.

The reason this flag is here is because Aragog will crash when it is turned on - I think related to this issue with density that @planetmariana is working on. We should try to remove this as soon as possible.

tests/data/integration/physical/plot_population_mass_radius.png
nichollsh: In this case the line is hidden behind the black dot because the model only evolves for a few kyr.

tests/data/integration/physical/plot_population_time_density.png
nichollsh: Ah yes. I'll fix this.

nichollsh:
Thanks a lot, @timlichtenberg! We should definitely look into this structure stuff more deeply, but I think that this is better than the previous implementation of over-specifying the mass and radius.

PR #275, opened 2024-11-19
Dummy tidal heating

### LINKED ISSUE #190: Dummy tidal heating calculation
For some of the planets of interest, tidal heating could be significant. [This paper](https://iopscience.iop.org/article/10.3847/1538-4357/ab0c21/pdf) calculates tidal heating on TRAPPIST-1 b to be up to several 100 W/m^2. For the planet L 98-59 d it could be up to 1000 W/m^2. This could be the difference between a solid surface and a deep magma ocean. The methods in the paper above are too complicated for a dummy tidal model.

A first step would be to implement a dummy tidal heating module. Ideally, this would be handled by Aragog (or SPIDER). However, a very simple approximation would be to use a fixed value of energy dissipation [W m-2 kg-1], which is then scaled by the mass of the mantle [kg], and then subtract this from the heat flux escaping the interior. This would essentially represent a global tidal heat dissipation that acts to slow down (or prevent) solidification of the planet.

We should explore better models for this in the future. Potentially this could be handled by Aragog directly?

nichollsh:
Adds a module for dummy tidal heating. This is implemented by simply allowing the user to set the tidal power density `H_tide` in the configuration file. Closes #190.

This required the following changes:
* Update radiative equilibrium criterion to account for the interior heat flux `F_tide` and `F_radio`. This is because global energy balance is now achieved when $F_{\text{atm}}=F_{\text{tide}}+F_{\text{radio}}$ not when $F_{\text{atm}}=0$, since $F_{\text{tide}}+F_{\text{radio}}$ effectively acts like an $F_{\text{eff}}$ (cf. hot jupiter atmosphere models).
* Add relative tolerance parameter to radeqm stop condition, rather than just an absolute tolerance (which was called `F_crit`). See https://web.mit.edu/10.001/Web/Tips/Converge.htm
* Update plots to include tidal and radiogenic heat fluxes. The radiogenic heating isn't new physics - PROTEUS is just now storing what SPIDER was already calculating.
* Update configuration files with new parameters.

This also lays the groundwork for implementing a new comprehensive tidal heating and orbital evolution calculation. It's not clear how these other models would operate, but they would at least have to communicate some information relating to tidal heating to the interior models.

Minor things:
* Add `apt-get update` to Ubuntu workflow, following this issue: https://github.com/actions/runner-images/issues/5470#issuecomment-1115949858.


timlichtenberg:
Do you want me or someone else to review this already?

input/aragog.toml
```diff
@@ -68,8 +68,9 @@ author  = "Harrison Nicholls, Tim Lichtenberg"
 
         # radiative equilibrium
         [params.stop.radeqm]
-            enabled = false
-            F_crit  = 0.2     # W m-2, model will terminate when |F_atm| < F_crit
+            enabled = true
```
timlichtenberg, line 71: Please add some qualifier here for what is "enabled" or "disabled".

timlichtenberg: Throughout all .toml files

input/default.toml
```diff
@@ -69,11 +69,12 @@ author  = "Harrison Nicholls, Tim Lichtenberg"
...
             enabled = true
-            F_crit  = 0.2     # W m-2, model will terminate when |F_atm| < F_crit
+            atol    = 0.2     # absolute tolerance [W m-2]
+            rtol    = 1e-3    # relative tolerance
 
         # steady state
         [params.stop.steady]
-            enabled  = true
+            enabled  = false
             F_crit   = 0.8        # Maximum absolute value of F_atm allowed for convergence
```
timlichtenberg, line 78: This is in W m-2 as well,right?

input/aragog.toml
nichollsh: Done. Added a qualifier at the start of the `[params.stop]` table.

timlichtenberg (approved):
Looks good, my only question is whether to include actual values also for radiogenic heating in the cfg files now or in a follow-up PR.

PR #278, opened 2024-11-21
Pass radionuclide parameters from config

### LINKED ISSUE #276: Set radiogenic heating from config file
SPIDER includes radiogenic heating, and Aragog will soon. This is currently hardcoded by `spider.py` to use specific abundances of specific radioisotopes.

This should be handled more generally, for the purposes of integrating it into Aragog, and could be exposed to the configuration file. Maybe the initial concentrations of these elements could be set in the `delivery` section of the table.

nichollsh:
Adds ability to pass radionuclide parameters from the configuration file, and to turn off radiogenic heating entirely (for both SPIDER and Aragog). This is done by specifying concentrations (ppmw) for naturally occurring Th, U, K, and the reference age to which these concentrations correspond. Other physical quantities (heat production, half life, isotopic abundance) are specified in `constants.py`.

Closes #276.

Requires https://github.com/ExPlanetology/aragog/pull/33 to be merged first.

I have compared the results against SPIDER, and we get the exact same heat production [W/kg] when using the same parameters. We also get reasonable values for the globally-averaged heat flux `F_radio` which is ~0.8 W/m^2.


timlichtenberg (approved):
Looks very reasonable, well done. Another set of assumptions to test thoroughly sometime. ;)

nichollsh:
Yes, definitely. Thanks Tim!

lsoucasse:
Oh... I was mistaken. It works fine to run with current version and radiogenic_heat turned to false. I had issues in the past running with empty radionucleide array that made me creating a dummy one, but it seems it has been fixed in the meantime.

nichollsh:
Fair enough. Thanks for pointing this out though - I had not considered this condition!

Issue #277, opened 2024-11-20
Structure calculation mass from radius does not work with Aragog


lsoucasse:
With the new structure scheme, the mantle calculation from radius does not work because the gravity is no longer initialized and we need the gravity to compute the mantle mass with Aragog. You can actually see in the physical test results that both mantle mass and gravity are increasing during the first iterations (IC_INTERIOR = 1) up to a not converged value while it should be correct from the beginning.

I think for now we have to have the EOS function that uses Aragog in Proteus and invert these relationships to set either the radius from mass or the mass from radius.

lsoucasse:
By the way, I don't understand why the structure calculation is within the time loop. As far as I understand all our interior models (dummy/Spider/Aragog) work with a fixed mesh/gravity/mantle_mass and we ignore mass variations due to the melting. Am I correct @nichollsh , @timlichtenberg ?

lsoucasse:
I fully understand the mid-term goal of changing the structure with time.

But for now, it solves it 2/3 times in the init and also solves it again when we resume a simulation. I am curious to know how Spider behaves actually. Does it give constant (and consistent) results from first iteration?

nichollsh:
I agree that it's definitely not optimal. For the default configuration (where structure is set by mass using SPIDER) I get this result:
![image](https://github.com/user-attachments/assets/61c51682-6015-43ba-a4a0-295c232a46b2)

lsoucasse:
And what about the other way around, when you set the radius? I am just afraid spider does not compute the mass correctly, because it might need the gravity constant which is set to zero at the beginning now.

I think we need a `determine_interior_mass` function in the `solve_structure` function to get the mantle mass and gravity constant for a given radius (calling iteratively the mantle mass calculation with updated gravity constant).

nichollsh:
Good point. I think there are actually two things here.

A) The Aragog case is obviously not right because of this `solve_g` flag being set to false. Setting it to true with Aragog causes the model to crash when I test it. We should definitely look into this.

B) Using SPIDER and setting the structure with the mass, `solve_g` is always true, so it will determine the gravity consistently with mass/radius. However, we should check that this is also fine when setting structure from radius, because as you say, the gravity needs to be solved self-consistently and this can only be done with a `determine_interior_mass` calculation.

lsoucasse:
Yes I agree.

I will take care of Aragog.

[...]

PR #279, opened 2024-11-21
Ls/resume aragog

lsoucasse:
This works allows Aragog to resume a simulation by reading init temperature file from output repository. It seems to behave well except for the mantle mass and gravity but it is a separate issue #277 . It also prompts the aragog logging. I hard coded the default level but we can easily make it configurable.


nichollsh:
Why is the resume functionality now included in the configuration file? Isn't it better to keep it as a flag, so that we don't have to edit the configuration file in order to resume a simulation.

lsoucasse:
> Why has the resume functionality now moved to the configuration file? Isn't it better to keep it as a flag, so that we don't have to edit the configuration file in order to resume a simulation?

I need Aragog to know whether we resume or not because the way to access the initial temperature is different. I found it was a nice way to pass the information via the config object. The resume argument is part of the config but you don't need to set it up in the config file (also it could be an option) as default is false and it is turned to true when you use the resume flag.

nichollsh:
Ah okay, I see. That's fine with me.

nichollsh (approved):

Issue #281, opened 2024-11-21
Pass tidal heating to Aragog and SPIDER


nichollsh:
Aragog and SPIDER both now support tidal heating as input variables. This can be set in PROTEUS using the dummy orbit module. 

The tidal heating provided by the dummy orbit module should be passed to Aragog and SPIDER when it is enabled. This can be done in a somewhat similar manner to radiogenic heating.

Future issues should handle physical calculations of the tidal heating rate. There are various models to explore to do this.

Issue #280, opened 2024-11-21
GitHub workflow sometimes fails on MacOS


nichollsh:
Sometimes the MacOS test workflow fails on the step where it installs the brew dependencies. This is the first step of the workflow, so it's not directly related to core PROTEUS code. See example here: https://github.com/FormingWorlds/PROTEUS/actions/runs/11953651244/job/33321985018

**To Reproduce**
Run the test workflow with a MacOS runner.

**Expected behavior**
For brew to execute without errors.

**Additional context**
These seem to be the same problem: https://github.com/Homebrew/brew/issues/17224, https://github.com/apache/arrow/pull/44798

nichollsh:
@stefsmeets do you have any suggestions for how to fix this?

nichollsh:
I just noticed that the second issue mentioned above (https://github.com/apache/arrow/pull/44798) is only 4 hours old, so maybe this will be fixed further up in the supply chain.

stefsmeets:
Could we mimick the apache changes:

https://github.com/raulcd/arrow/blob/e45627d759bb499ee58e72de42a7ea7319cc2204/.github/workflows/python.yml#L192-L195

Seems like pkgconf is already installed and brew fails to recognize this and somehow does not have rights to overwrite it.

Issue #192, opened 2024-09-26
Plotting tests fail because of different image sizes


lsoucasse:
When comparing reference plots with CI plots, the image sizes are slightly different.

I tried to resize the image before using the `compare_image` function of matplotlib in `tests/integration/test_dummy_integration.py`, but the resizing creates small shifts in axes and labels which make the comparison not relevant.

timlichtenberg:
It seems these reference plots are quite sensitive to minor changes in various things. Do we think it is a long-lasting solution to directly compare reference plots to one another or should we systematically resort to comparing only output numbers?

stefsmeets:
> It seems these reference plots are quite sensitive to minor changes in various things. Do we think it is a long-lasting solution to directly compare reference plots to one another or should we systematically resort to comparing only output numbers?

It's a good question. Comparing numbers is great, and testing the plots does not replace this. However, a visual comparison of the data can often help quickly make sense of the data.

In addition, we want to regression test the plots (a large part of the output from PROTEUS is in the plots after all) as well as improve code coverage (see #268).

About this being a long-lasting solution: the pattern itself is used by matplotlib for their internal testing. I have used the pattern in several projects, and I have not encountered issues, the plots are typically reproducible cross-platform and cross-version.

We should look into the difference however. The major difference between other implementations of this pattern is that they directly work with the `fig` object, stripping any styling, which reduces differences with e.g. font rendering (usually the biggest source of differences).

stefsmeets:
The current code only supports saving the file and returning the figure (e.g. for use in a notebook). We should split out the actual plotting bit from the io (e.g. saving vs. displaying). This will also help solve the issue with the tests.

nichollsh:
https://stackoverflow.com/questions/67515453/output-images-of-matplotlib-figures-have-different-sizes-across-operating-system

PR #282, opened 2024-11-21
Pass tidal heating to Aragog and SPIDER

nichollsh:
Passes tidal heating rate from PROTEUS to SPIDER and Aragog. Currently, this heating rate is just a constant value (W/kg) which is homogeneous throughout the mantle of the planet. 

This can only be specified through the dummy orbit module (in the config file) for now. 

Requires https://github.com/ExPlanetology/aragog/pull/34 to be merged first.

Closes https://github.com/FormingWorlds/PROTEUS/issues/281

Also includes patch for #280.


nichollsh:
This seems to work quite nicely. The plot below shows that for a non-zero tidal heat production, the planet settles into radiative equilibrium with non-zero `F_atm` and the surface temperature stops changing with time.
image

src/proteus/interior/aragog.py
```diff
@@ -64,6 +64,14 @@ def RunAragog(config:Config, dirs:dict, IC_INTERIOR:int, hf_row:dict, hf_all:pd.
...
         UpdateAragogSolver(dt, hf_row)
 
+    # Update tidal heating inside Aragog
+    tidal_value= 0.0
+    if config.interior.tidal_heat:
+        if config.orbit.dummy:
+            tidal_value = config.orbit.dummy.H_tide
+    tidal_value /= aragog_solver.parameters.scalings.power_per_mass
+    aragog_solver.parameters.energy.tidal_value = tidal_value
+
```
lsoucasse, line 74: These lines should go into the `SetupAragogSolver` (done once for all from the config).

tests/data/integration/dummy/runtime_helpfile.csv
lsoucasse, line 1: Why does the dummy test values change? The input file is the same.

nichollsh: Having two separate dummy configuration files is quite confusing. I'll remove the one in the input folder.

The image files from the physical test have changed in the sense of binary files, I guess due to some numerical thing or as a result of #192. However, if you look at the before/after images with the GitHub comparison tool,

lsoucasse: OK for the plots but what about the helpfile? If it is just tiny differences the test should pass anyway without having to push a new one?

src/proteus/interior/aragog.py
lsoucasse: > I have now moved this, although in the future we will also need to make sure it's updated in the UpdateAragogSolver function, since the tidal heat production will evolve over time.

Make sense, thanks!

tests/data/integration/dummy/runtime_helpfile.csv
nichollsh: @lsoucasse good point. It is not clear to me at the moment why exactly the values seem to have changed in the csv file. It seems to be a result of Aragog returning a different surface temperature, for some reason.

lsoucasse (approved):

Issue #258, opened 2024-11-12
Condition on Pxuv computation (only for Zephyrus escape) + check on Mors/Baraffe tracks


EmmaPostolec:
Add few lines in the atmos_clim wrapper to compute Pxuv only if zephyrus is activated as suggested by @lsoucasse in #227 

Quotting @lsoucasse here : 
" It would be worth adding an "if zephyrus activated" statement to the Pxuv computation, and if not fill the Pxuv entry with the surface pressure. I think it kind of goes with issue https://github.com/FormingWorlds/PROTEUS/issues/232. "

lsoucasse:
Could you also add in this work a check that Mors/Baraffe tracks are selected when running zephyrus. It should break if not. Then it will solve #232 .

timlichtenberg:
Why is Zephyrus incompatible with Spada tracks btw?

lsoucasse:
> Why is Zephyrus incompatible with Spada tracks btw?

It is the other way around. Zephyrus can only work if Spada tracks (Mors) are used.

timlichtenberg:
Ok, thanks for clarifying! This means Spada tracks should always be the default option in all config files.

lsoucasse:
Issue closed by PR #290.

Issue #232, opened 2024-10-25
Validation of config options compatibility


lsoucasse:
We currently have validators for each config entry which make sure that the value entered is correct.

But we also need to check the compatibility between config options. One instance: the Zephyrus model only works if you use Mors with Spada data. I guess there are (or will be) other requirements. Where/how should we implement these checks @stefsmeets ?

lsoucasse:
[Here](https://github.com/FormingWorlds/PROTEUS/pull/227#issuecomment-2470283349) is also an example of potential config compatibility issue.

lsoucasse:
On this topic we decided to implement appropriate check within the submodules when run.

PR #290, opened 2024-11-25
Validate escape config

### LINKED ISSUE #286: Don't set escape level when escape is disabled
The current escape implementation will store the escape radius (R_xuv) and pressure (`P_xuv`) based on the atmosphere profiles determined by JANUS/AGNI. 

This is fine when escape is turned on, but it should not happen when escape is disabled, and instead these values should be set to the surface.

nichollsh:
Set `p_xuv` and `z_xuv` to surface when ZEPHYRUS is disabled. Closes #286 and https://github.com/FormingWorlds/PROTEUS/issues/258

Ensure that configuration uses MORS+Spada when ZEPHYRUS is enabled. Related to #232.

Avoid spurious values for `period` in first iteration. Calculation of this quantity depends on `M_star`, which is undefined when `period` is first calculated, and can cause some strange behaviour. Fixed this by ensuring that `M_star` is always defined.


EmmaPostolec (approved):
Hi Harrison, very nice job ! The condition on Pxuv computation is really nice to save some computational time. Also, very nice idea to create the get_heigh_from_pressure function, very effective. Thanks for introducing error message also for the user. The computation of mu in the orbit/wrapper.py is not really linked to this issue ?

nichollsh:
Thanks Emma! The issue with mu is something that's slightly odd. I noticed that the tests *sometimes* fail with MacOS, even though they succeed on my PC. I think this is because of this `period` parameter being set by some undefined behaviour.

EmmaPostolec:
Also I think this PR is linked to https://github.com/FormingWorlds/PROTEUS/issues/258 ?

nichollsh:
Yes! It is. I didn't notice this issue, thanks Emma.

EmmaPostolec:
No worries! I assigned myself this task but you did it faster than me :)

lsoucasse:
I happy with these changes. You can merge them now Harrison so that I can update and merge my current PR afterwards.

nichollsh:
Ah okay, sure! Thanks Laurent.

PR #291, opened 2024-11-25
Ls/structure aragog

lsoucasse:
This work debugs the structure calculation with Aragog and refactors setting the structure of the planets. Main changes are:

- Using an updated version of Aragog with correct mantle mass calculation
- Setting consistently the core mass using a core density setup through the config (I provided with a default value is now consistly used through all interior modules)
- Removed the structure calculation out of the time loop, as we consider fixed mantle mass and core mass in the current state of the code.


lsoucasse:
Could you approve it again @nichollsh? Had to merge your changes in the meantime which outdated your previous review. Might be also worth for you to check the behaviour with Spider, although changes should be minor.

nichollsh:
This seems to work fine with SPIDER too. Thanks, Laurent

PR #292, opened 2024-11-26
Allow setting Hydrogen inventory relative to mantle mass

nichollsh:
Currently, the user can only set the hydrogen inventory in absolute terms (`H_oceans`). This becomes somewhat annoying when varying planet mass -- such as when comparing different planets in the same star system -- since it means that the relative amount of hydrogen is implicitly changing. 

This PR makes it possible to set the hydrogen inventory in relative terms (`H_ppmw`) in the same way as we do with nitrogen and sulfur. This is then summed with the absolute amount, although I can imagine that  the user will probably only set one of these parameters to be >0 at any one time.

### DIFF
--- a/input/aragog.toml
+++ b/input/aragog.toml
@@ -239,9 +239,10 @@ author  = "Harrison Nicholls, Tim Lichtenberg"
     # Set initial volatile inventory by planetary element abundances
     [delivery.elements]
         CH_ratio    = 1.0       # C/H ratio in mantle/atmosphere system
-        H_oceans    = 6.0       # Hydrogen inventory in units of equivalent Earth oceans, by mass
-        N_ppmw      = 2.0       # Nitrogen inventory in ppmw relative to mantle mass, by mass
-        S_ppmw      = 200.0     # Sulfur inventory in ppmw relative to mass of melt
+        H_oceans    = 6.0       # Hydrogen inventory in units of equivalent Earth oceans
+        H_ppmw      = 0.0       # Hydrogen inventory in ppmw relative to mantle mass
+        N_ppmw      = 2.0       # Nitrogen inventory in ppmw relative to mantle mass
+        S_ppmw      = 200.0     # Sulfur inventory in ppmw relative to mantle mass
 
     # Set initial volatile inventory by partial pressures in atmosphere
     [delivery.volatiles]
--- a/input/aragog.toml
+++ b/input/aragog.toml
@@ -238,7 +238,7 @@ author  = "Harrison Nicholls, Tim Lichtenberg"
 
     # Set initial volatile inventory by planetary element abundances
     [delivery.elements]
-        CH_ratio    = 1.0       # C/H ratio in mantle/atmosphere system
+        CH_ratio    = 1.0       # C/H mass ratio in mantle/atmosphere system
         H_oceans    = 0.0       # Hydrogen inventory in units of equivalent Earth oceans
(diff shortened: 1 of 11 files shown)


input/aragog.toml
```diff
@@ -239,9 +239,10 @@ author  = "Harrison Nicholls, Tim Lichtenberg"
     # Set initial volatile inventory by planetary element abundances
     [delivery.elements]
         CH_ratio    = 1.0       # C/H ratio in mantle/atmosphere system
-        H_oceans    = 6.0       # Hydrogen inventory in units of equivalent Earth oceans, by mass
-        N_ppmw      = 2.0       # Nitrogen inventory in ppmw relative to mantle mass, by mass
-        S_ppmw      = 200.0     # Sulfur inventory in ppmw relative to mass of melt
+        H_oceans    = 0.0       # Hydrogen inventory in units of equivalent Earth oceans
```
timlichtenberg, line 242: So both settings are possible then? Does this also translate to the CALLIOPE repository itself?

timlichtenberg: If so you need to make clear to set only one or the other somehow, via comments or if conditions.

input/hd63433d.toml
```diff
@@ -240,6 +240,7 @@ author  = "Harrison Nicholls, Tim Lichtenberg"
     [delivery.elements]
         CH_ratio    = 1.0       # C/H ratio in mantle/atmosphere system
```
timlichtenberg, line 241: The C content is then the only quantity that is set via different means than ppm. Change this as well? (Perhaps in a new PR).

In addition, I just realise it is not clear here whether this C/H ratio is by mass or by mole fraction. A comment would suffice.

src/proteus/config/_delivery.py
```diff
@@ -15,14 +15,17 @@ class Elements:
     CH_ratio: float
         Volatile C/H nass ratio in combined mantle+atmosphere system.
```
timlichtenberg, line 16:
```suggestion
        Volatile C/H mass ratio in combined mantle+atmosphere system.
```

input/aragog.toml
nichollsh: Yes, this makes both of them possible. In theory the user could use both at once, for whatever reason, but it probably makes sense for them to only set one of them to be >0 at any one time. 

Could add a warning for this I suppose.

timlichtenberg (approved):
Looks good, only a few clarifying statements in the comments needed. Additionally, the C/H setting could be reworked similarly.

input/hd63433d.toml
nichollsh: Yes, this means that C/H is the only quantity that cannot be specified in ppmw. We could probably make a new issue for this.

I'll fix the description.

input/aragog.toml
nichollsh: This doesn't require any changes to CALLIOPE, because it converts the H_ppmw to an equivalent amount of H_oceans before the value is passed to CALLIOPE.

timlichtenberg: How would setting both of them make sense? This gives two different H inventories, no?

nichollsh:
Thanks Tim! I'll make these changes.

input/aragog.toml
nichollsh: The current formulation sums them together, so the total hydrogen inventory is now:

H_total = H_oceans * ocean + H_ppmw * mantle.

This obviously reduces quite nicely if the user just sets one of the two parameters. And it allows them to set both (which will generate a warning) if they would want to do that too.

PR #297, opened 2024-11-27
Update termination criteria

### LINKED ISSUE #295: Model does not terminate when F_tidal > F_atm
The check for radiative equilibrium requires that `F_atm ~ F_tidal + F_radio`.

If `F_tidal + F_radio` is greater than `F_atm`, the planet will want to warm up from its initial state. However if the user has set `config.atmos_clim.prevent_warming = True` then the planet is prevented from doing so, and therefore it will not be able to increase `F_atm` in order to reach radiative equilibrium. 

This means that the model continues when it should probably terminate, and only stops when the maximum time is reached. This is a waste of resources.

In theory the steady state break condition can prevent this, but it's quite a poor metric for model termination. Steady state can physically only happen at radiative equilibrium in the absence of escape, so the steady state condition is effectively just checking for radiative equilibrium (but in a worse way than through the comparison above). 

It would be simpler to remove the steady-state break condition, use the radiative equilibrium one instead, and then also add a check for when `F_atm < F_tidal + F_radio` and `prevent_warming=True`.

nichollsh:
Removes old steady state check and improves radiative equilibrium / energy balance condition. Closes #295.

Tidies up this code significantly, and also prints information to log regarding which criteria are enabled.

Fixes bug where `config.params.stop.iter.minimum` wasn't actually being used.

### DIFF
--- a/input/aragog.toml
+++ b/input/aragog.toml
@@ -53,7 +53,7 @@ author  = "Harrison Nicholls, Tim Lichtenberg"
         # required number of iterations
         [params.stop.iters]
             enabled = true
-            minimum = 3
+            minimum = 5
             maximum = 9000
 
         # required time constraints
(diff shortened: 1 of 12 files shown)


lsoucasse (approved):
I really like these new changes. It lightens the main script and it is now very clear how the code goes through all the stopping conditions.

nichollsh:
Thanks!

PR #298, opened 2024-11-28
Allow setting volatile abundances as both metallicity or concentration

nichollsh:
Following #292, this change extends the methods for setting the planet's volatile inventory.

It is now possible to set the volatiles as follows:
* Hydrogen: equivalent Earth oceans (`H_oceans`) or relative to mantle mass (`H_ppmw`)
* Carbon: as metallicity (`CH_ratio`) or relative to mantle mass (`C_ppmw`)
* Nitrogen: as metallicity (`NH_ratio`) or relative to mantle mass (`N_ppmw`)
* Sulfur: as metallicity (`SH_ratio`) or relative to mantle mass (`S_ppmw`)

### DIFF
--- a/input/aragog.toml
+++ b/input/aragog.toml
@@ -73,12 +73,6 @@ author  = "Harrison Nicholls, Tim Lichtenberg"
             atol    = 0.2     # absolute tolerance [W m-2]
             rtol    = 1e-3    # relative tolerance
 
-        # steady state
-        [params.stop.steady]
-            enabled  = false
-            F_crit   = 0.8        # Maximum absolute value of F_atm allowed for convergence
-            dprel    = 1.0e-9     # Percentage change in melt fraction over time (dp/p)/dt*100
-
         [params.stop.escape]
             enabled   = true
             mass_frac = 3e-4      # Stop when atm_mass < this frac of initial mass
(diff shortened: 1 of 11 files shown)


nichollsh:
Fair point - it could be confusing until they run the code and get an error. However, I do think that we should include in the configuration files all the possible parameters so that the user knows what is available. As we discussed before, it's likely that they will not immediately check the documentation*.

We could add default values of zero as you describe and then comment-out the unused parameters in the configuration. That way the user knows that they are available, but they aren't expected to actually provide both X_ppmw and XH_ratio at the same time. @lsoucasse what do you think of this?

* maybe we could add something in the code that prompts the user to check the docs if something fails?

lsoucasse:
> Fair point - it could be confusing until they run the code and get an error. However, I do think that we should include in the configuration files all the possible parameters so that the user knows what is available. As we discussed before, it's likely that they will not immediately check the documentation*.
> 
> We could add default values of zero as you describe and then comment-out (but leave in the file) the unused parameters in the configuration. That way the user knows that they are available, but they aren't expected to actually provide both X_ppmw and XH_ratio at the same time. @lsoucasse what do you think of this?
> 
> * maybe we could add something in the code that prompts the user to check the docs if something fails?

Commented the unused options is a good idea so that indeed the user know there is an alternative without looking at the doc. I am happy with this solution.

lsoucasse (approved):

PR #300, opened 2024-12-05
Organize input folder 

EmmaPostolec:
Hi all,
I discussed with @nichollsh about the issue of escape.toml being accidentally removed from the input folder previously. To address this and to avoid having too many initialization files in the input directory, we thought about organizing them into subfolders.

In this PR, I have re-added dummy.toml and escape.toml and organized the input folder into three subfolders: demos, planets, and trappist1.

Please let us know your thoughts on this approach and if you have any suggestions, ideas, or improvements.


nichollsh (approved):

nichollsh:
The tests seem to be failing due to inconsistencies in the value of `CH4_mol_atm`.

> DataFrame.iloc[:, 71] (column name="CH4_mol_atm") values are different (0.31447 %)

I have noticed this error prior to your PR, @EmmaPostolec, so it's most likely not a result of these changes. This quantity is very sensitive to temperature at the surface, so differences on the order of 0.3% are not unexpected. If you run the tests again, they will probably pass. We need to think about this some more.

EmmaPostolec:
Hi @nichollsh,

Thanks a lot for mentioning this! At first, I thought it might be related to the change in the path for the input files we are testing, but after digging deeper into the test script, I am not so sure.

I am re-running the test on GitHub right now let's hope it works! Alternatively, would it make sense to adjust the test criteria to 0.4% instead of 0.3% to ensure it passes?

Also, I was wondering if the tests are failing for the same reason in PR 301: https://github.com/FormingWorlds/PROTEUS/pull/301.

Let me know what you think is the best approach!

timlichtenberg (approved):
Looks good! Once the test passes can be merged in my view. Adjusting the criteria a bit higher would be ok from my point of view.

nichollsh:
> Let me know what you think is the best approach!

I think the best thing to do here is to re-run the tests until they pass, and then merge these PRs, since we know they are not causing the problem.

We should revisit the failing tests in a different issue. We could increase the tolerance, but if it's only CH4_mol_atm that ever causes problems then maybe we could exclude that particular  column from the tests. It is testing the same physics as CH4_bar, H2_mol_atm, etc, anyway.

nichollsh:
Made an issue here: #302

EmmaPostolec:
Nice thanks for the comments @timlichtenberg and @nichollsh ! Can I merge it now ? :)

timlichtenberg:
Yes!

PR #304, opened 2024-12-09
Neglect checking for CH4 molar abundance

### LINKED ISSUE #302: Tests sometimes fail due to `CH4_mol_atm`
**Describe the bug**
The physical integration test sometimes fails due to mismatching values of `CH4_atm_mol`. This is seemingly random, but it isn't surprising. See #300 for example.

**To Reproduce**
Run the physical integration test. Happens on Ubuntu and MacOS.

**Additional context**
The abundance of CH4 is highly temperature dependent, which is a physical result of its equilibrium thermochemistry. However, this causes problems when the total amount of atmosphere is tenuous (due to escape) as the small relative errors on CH4 abundance from earlier in the simulation can become amplified.

For example: 
At `time=0` arbitrary units there are 1e30 moles of CH4, at `time=100` there are 2e30 moles, and at `time=300` there are 1e10 moles. This means that any small errors arising at `time=0` become problematic at `time=300`. 

Since this problem only seems to happen for CH4, we could consider excluding it from the test suite. This can be justified by the fact that the same physics and code are tested by the abundances of H2 and CO2, as well as the partial pressure of CH4 itself.

nichollsh:
As discussed this morning.
Closes #302.

### DIFF
--- a/tests/integration/test_integration_dummy.py
+++ b/tests/integration/test_integration_dummy.py
@@ -23,10 +23,15 @@ def test_dummy_run(dummy_run):
     ref_dir = PROTEUS_ROOT / 'tests' / 'data' / 'integration' / 'dummy'
 
     hf_all = ReadHelpfileFromCSV(out_dir)
-    hf_all_ref = ReadHelpfileFromCSV(ref_dir)
+    hf_ref = ReadHelpfileFromCSV(ref_dir)
+
+    # Neglect these columns
+    neglect = ["CH4_mol_atm"]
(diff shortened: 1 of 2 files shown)


nichollsh:
This seems to have fixed the issue with the tests.

lsoucasse (approved):
Good to have this neglecting option. Might consider adding other columns in the future. Thanks @nichollsh !

nichollsh:
Thanks @lsoucasse!

PR #305, opened 2024-12-09
Set `p_obs` from CFF, update postprocessing scripts

### LINKED ISSUE #303: Calculate transit pressure from peak of contribution function
The pressure level probed in transmission is what sets the observed radius, and therefore the bulk density estimate. It is important that this is estimated in a reasonable manner.

At the moment, this pressure level is hardcoded to 100 Pa (1 mbar) but it should be set based on the optical properties of the atmosphere. AGNI has functionality to calculate this based on where the contribution function peaks. For JANUS, we can make this a config parameter, which is better than hiding it in the code.

nichollsh:
* Update postprocessing script, which was broken following recent AGNI updates.
* Store observed level `p_obs` in helpfile. Calculated by AGNI using contribution function, set by JANUS in config. Closes #303.
* AGNI makes CFF plot when called from PROTEUS.
* Allow disabling condensation in AGNI config block. Check for when all gases are allowed to be condensible, since this was causing the model to crash in some cases.
* Update initial temperature guess for AGNI first iteration. Previous guess was causing problems for low-gravity planets.
* Update configuration files with new options.
* Update tests with new `p_obs` column.
* Moved `download_sufficient_data` call into `start()` so that it isn't called when Proteus is imported. This was making some of the plotting CLI very slow, since it needed to import all of the OSF library dependencies just to make a plot.


lsoucasse (approved):
From my side the changes look good!

PR #306, opened 2024-12-09
Ls/aragog lookup

lsoucasse:
This work makes the use of OSF look up tables the default behaviour in Aragog. Thanks @planetmariana for debugging this!

I also slightly changed the structure calculation such that it takes the value in the config file as an initial guess, that way we can speed up the solution search.

The physical test is much slower when using the lookup tables. So I strongly reduced the time step to keep a reasonable computing time.


lsoucasse:
Hi @planetmariana, I have an issue with the Github CI to go further with this PR. But could you already download this version, run the test locally and tell me if you are happy with this implementation?

planetmariana:
Hi @lsoucasse @timlichtenberg @nichollsh,

I tried the new version of PROTEUS with the implementation of the lookup tables from Aragog running the config file Aragog.toml. It didn't crashed, which is good from one side, it lasted 12hr in total. But the solution looks like the rheological front is fluctuating weirdly, but I guess it's because the config file was for a dummy case, and it can be a better set on initial conditions. I'm gonna test now with the exact same conditions for SPIDER. 

plot_interior
plot_global_log

timlichtenberg:
This looks like the behaviour in this issue a bit: https://github.com/FormingWorlds/PROTEUS/issues/193. Could be related to timestepping instead of aragog itself?

lsoucasse:
@tim and @planetmariana. I fixed the CI. I suggest we merge this as the runs do not crash and give reasonable output. We should create a separate issue addressing the Spider/Aragog cross validation.

planetmariana (approved):
@lsoucasse looks good to me !

PR #308, opened 2024-12-10
Minor update of escape.toml

EmmaPostolec:
Hi all, here is a very minor update in input/demos/escape.toml. I just changed the output folder name to 'escape' instead of 'default'. I also change the stellar tracks to 'spada' instead of 'baraffe' because this is required for the escape computation. Those 2 mistakes aredue to a bad use of copy/paste in a previous PR.


timlichtenberg (approved):
👍

lsoucasse (approved):

PR #311, opened 2024-12-16
Creating folder of PETSc 

planetmariana:
This PR is for creating the PETSc folder before executing installation. Previously, when running ./tools/get_petsc.sh it was not  installing this dependency as it was not finding the folder to execute it and therefore get_spider.sh was also crashing. With this change, the installation continues smoothly as pointed in the instructions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,6 +55,7 @@ Zephyrus
 vulcan
 VULCAN
 Vulcan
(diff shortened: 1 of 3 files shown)


lsoucasse (approved):
Good you spot this Mariana. Thanks!

PR #317, opened 2025-01-14
Restrict time-steps to be smaller than the predicted time until model termination

### LINKED ISSUE #193: Oscillatory behaviour of fluxes and surface temperature

nichollsh:
In order to address the oscillations identified in #193, this PR introduces a mechanism by which the interior time-step is limited to be less than the expected time remaining in the simulation.

This is done by estimating the rates of change ($d/dt$) in melt fraction ($\Phi$) and flux residual ($F_{atm}-F_{radio}-F_{tidal}$) as a simple finite difference calculation. Given the current values for melt fraction and flux, we can then approximate how long the model has 'left' until the planet solidifies or reaches radiative equilibrium. The time-step algorithm (`timestep.py`) then restricts the next model time-step to be less than expected time left in the simulation. 

Since the time left will be short near model termination, this seems to be effective at preventing oscillations. See image below showing the difference in behaviour with the dummy configuration (no escape). Closes #193. This likely also addresses https://github.com/ExPlanetology/aragog/issues/21 but I have not checked this. I have run the code with the default (JANUS+SPIDER) configuration, and it seems to work fine.

Also added a new flag `save_mod` for setting how frequently data are saved to the CSV helpfile. This improves performance substantially in some cases. Also fixed some minor formatting inconsistencies in the logger.

untitled


nichollsh:
@timlichtenberg the tests on this are now passing fine, so please can you review it whenever you are ready?

timlichtenberg:
On it, will do in the next hour.

nichollsh:
Thanks! Added Laurent in case he also has any suggestions.

lsoucasse (approved):
Looks good Harrison.

timlichtenberg (approved):
Agree, good work!

nichollsh:
Thanks both!

Issue #283, opened 2024-11-22
Physical orbit/tides module


nichollsh:
PROTEUS now includes a dummy Orbit module which simply fixes the semi-major axis, eccentricity, and tidal heat flux to the values provided in the configuration file.

We should explore more realistic models for tidal heating and orbital evolution, as it could potentially have a large impact on the energy budget of these planets.

This will probably need to be addressed in conjunction with #63.

timlichtenberg:
Since the links are now available, the tidal model should encapsulate analytic solutions for both mantle fluid-driven (https://arxiv.org/abs/2412.07285) and thermal atmospheric (https://arxiv.org/abs/2410.16200) tides.

PR #321, opened 2025-01-17
Support for depth-dependent tidal heating

nichollsh:
- Extract melt fraction arrays from Aragog/SPIDER. Stored in Proteus object.
- Melt fraction passed to `orbit` module, and used to calculate tidal heating. Currently only dummy orbit module supported but expecting a physical module soon.
- Tidal heating array (1D) from orbit module provides heating density (W/kg) at each interior level. This is passed to Aragog/SPIDER/dummy interior modules.
- Added tidal heating to interior plot `cpl_interior.py`.
- Update Aragog dependency version.

Requires:
- [x] Update SPIDER on OSF.
- [x] https://github.com/ExPlanetology/aragog/pull/41 merged.

After this has been merged we can directly look at tackling #283.


input/default.toml
```diff
@@ -109,6 +109,7 @@ author  = "Harrison Nicholls, Tim Lichtenberg"
 
     [orbit.dummy]
         H_tide  = 1e-11  # Fixed tidal power density [W kg-1]
+        Phi_tide = 0.3    # Maximum local melt fraction to which tides can be applied.
```
timlichtenberg, line 112: Not clear to me why the dummy module needs such a cutoff.

nichollsh: I agree that the dummy module should ideally be kept quite simple. 

However, Hamish's tidal model will have a melt-fraction (and viscosity) dependent calculation of the heating rate in the interior. It will also only be applicable to regions in which Phi < 30%. In order to validate the passing of Phi and H_tide between the interior and tidal heating models, it's useful if the dummy module also has an option for the tidal heating rate to depend on Phi. This will make implementing more realistic tidal heating models (such as Hamish's) much easier to do and to debug, since we now already have a easy to test mechanism by which to pass these data around.

We can also set `Phi_tide >= 1.0` to reproduce the previous behaviour in which tidal heating is applied homogeneously throughout the mantle.

src/proteus/proteus.py
```diff
@@ -73,8 +73,10 @@ def __init__(self, *, config_path: Path | str) -> None:
         self.loops = None
 
         # Interior
-        self.IC_INTERIOR = -1   # initial condition (1: new simulation, 2: mid-simulation)
-        self.dt = 1.0           # Time step variable [yr]
+        self.IC_INTERIOR = -1       # Initial condition (1: start, 2: mid-simulation)
```
timlichtenberg, line 76: This is confusing: the value is -1 but the comments only indicate 1 or 2 as possible values. What is -1?

nichollsh: This is a default value for which the model has not yet been initialised. The only valid values are 1 and 2, so -1 will cause the model to (intentionally) crash if it attempts to do an interior calculation before it is initialised. This should be better documented, I agree.

timlichtenberg (approved):

nichollsh:
Thank you Tim!

PR #326, opened 2025-01-20
Dummy tides inequality, convergence robustness, docs

nichollsh:
- Allow setting dummy tidal heating with an inequality so that we can represent dummy solid or liquid phase tides. E.g. `phi < 0.3`, `phi > 0.8`.
- Moved rheological transition location and width variables into config file.
- Update citation info
- Update termination criteria. They must be satisfied for two sequential iterations in order for the model to exit, otherwise the model can spuriously exit if it briefly satisfies the criteria while moving past them, which is not a physically justifiable point at which to terminate. Exception to this is atmospheric escape, which only needs to be satisfied for a single iteration.
- Updated physical test. New convergence criteria mean that it does 1 extra iteration.


planetmariana (approved):
The condition for the tides looks nice, thanks @nichollsh

Issue #309, opened 2024-12-11
Cross validation Spider / Aragog


lsoucasse:
We have now Aragog running, using look up table and producing reasonable output.

But we now need a comparison with Spider to validate the implementation, see first results from @planetmariana [here](https://github.com/FormingWorlds/PROTEUS/pull/306#issuecomment-2535521679).

planetmariana:
Hi @timlichtenberg @lsoucasse @nichollsh,

I run the exact same simulation as [here](https://github.com/FormingWorlds/PROTEUS/pull/306#issuecomment-2535521679), but with interior model=SPIDER. With those conditions the planet reached the global energy balance. Besides the time-stepping factor mentioned [#193](https://github.com/FormingWorlds/PROTEUS/issues/193), the rheological front when using Aragog starts above the CMB, so I'm a bit confused about it, as we are using the same melting curves, just in the P-T space. On the other side, both of them finish around the same time. We can discuss further about this, in the meantime I will check from the Aragog which parameter is causing the behaviour in the first 80 GPa.



![plot_interior](https://github.com/user-attachments/assets/7a789993-228d-4ee8-a44f-bb7c29a0f6c5)
![plot_global_log](https://github.com/user-attachments/assets/dc645747-0d9f-413a-bf01-febfb62add73)

timlichtenberg:
I copy the images from the Aragog run from https://github.com/FormingWorlds/PROTEUS/pull/306#issuecomment-2535521679 to here in this thread again.
![image](https://github.com/user-attachments/assets/83f86ea3-b5d6-4cad-8e45-9468b7697f65)
![image](https://github.com/user-attachments/assets/c939563f-a94a-4e9a-8b6f-b4de59ecf628)

It looks to me as if the Aragog simulation is not initialised on an adiabat. From 80 GPa downwards the temperature profile is isothermal already at the first timestep. We need to understand the reason for this, possibly the initialisation is not yet working correctly?

lsoucasse:
Could you send us the corresponding Aragog input file @planetmariana ?

planetmariana:
@lsoucasse is the input file Aragog.toml in main, I just change the interior model as SPIDER or Aragog, using both with JANUS. Both simulations took around 12hr, so for now I'm checking from Aragog which parameter can be changed. As discussed before @timlichtenberg  if I change rheological_transition_melt_fraction there's a difference, but I don't know how unphysical that can be. See below:


[...]

PR #329, opened 2025-01-23
Support for physical tidal heating module 'lovepy'

### LINKED ISSUE #283: Physical orbit/tides module

### LINKED ISSUE #327: OSF too many requests
When running large grids of models, sometimes they will crash because of a HTTP 429 (too many requests) response from OSF. This is because each grid point is individually querying OSF for the presence of data files.

We could mitigate this easily by having grid_proteus run each instance in offline mode, using the `--offline` flag.

### LINKED ISSUE #328: Dummy atmosphere `gamma` is ignored
The `gamma` opacity parameter for the dummy atmosphere is being ignored. The value is overwritten in the `RunDummyAtm` function.

To fix this, we simply need to make sure that the correct `gamma` value is used (set by `gamma = config.atmos_clim.dummy.gamma`).

nichollsh:
This PR comes with some significant changes to the code. These are all with the aim of supporting depth-dependent tidal heating, which is handled through the "lovepy" tidal heating model. This model only supports tides within rheologically-solid materials. Under the assumption of bottom-up solidification, this means that tidal heating is initially only applied at the CMB. The calculation then extends upwards in the mantle as it solidifies.

See this branch of Hamish's repo: https://github.com/hamishHay/lovepy/blob/julia_solid_body.

Major changes:
* Stores the current state of the interior inside an object `Interior_t` defined inside PROTEUS. This allows us to pass viscosity, melt-fraction, and density data from SPIDER to lovepy. It also allows heating arrays to be passed back to SPIDER.
* Support for calling lovepy as an orbit/tides module. Closes #283.
* Added new configuration options for lovepy. This includes `visc_thresh`, which is the minimum local viscosity required for a tidal heating calculation to be performed.
* Adapt dummy tidal heating model such that tides scale linearly with phi.

Minor changes:
* Updated print statements for consistent formatting
* Update interior plots
* Add lovepy instructions to PROTEUS documentation
* Updated configuration files with new options for shear and bulk moduli. These were previously hard-coded in the interior wrapper(s), but we also need them for tide modelling.
* Resolves #328.
* Store Im(k2) love number, semimajor axis, and eccentricity in helpfile.
* `grid_proteus.py` dispatches cases in offline mode by default. Mitigates and closes #327.
* Made core density and heat capacity non-optional parameters for the config.

Next steps:
* Validate tidal heating compatibility with aragog, although this should probably wait until #309 has been completed.
* Add tests for lovepy
* Add license and readme for lovepy repo


timlichtenberg:
Let us know when you want us to review this PR. :)

input/default.toml
```diff
@@ -111,13 +111,17 @@ author  = "Harrison Nicholls, Tim Lichtenberg"
...
+        visc_thresh = 1e9   # Minimum viscosity required for heating [Pa s]
 
 # Planetary structure - physics table
 [struct]
     set_by      = 'mass_tot'    # Variable to set interior structure: 'radius_int' or 'mass_tot'
     mass_tot    = 1.0           # Total planet mass [M_earth]
     radius_int  = 0.9           # Radius at mantle-atmosphere boundary [R_earth]
     corefrac    = 0.55          # non-dim., radius fraction
+    core_density = 10738.33      # Core density [kg m-3]
+    core_heatcap = 880.0         # Core specific heat capacity [J K-1 kg-1]
```
timlichtenberg, line 124: Are `core_density` and `core_heatcap` consistent with the standard settings in SPIDER; are these variables here used to define these quantities in all sub-modules?

src/proteus/interior/aragog.py
```diff
@@ -274,9 +272,13 @@ def GetAragogOutput(hf_row:dict):
     Htidal_s = aragog_output.heating_tidal[:,-1]  # [W kg-1]
     output["F_tidal"] = np.dot(Htidal_s, mass_s)/area
 
-    # Melt fraction array
-    Phi_array = aragog_output.melt_fraction_staggered[:,-1]
-    output["Phi_array"] = Phi_array
```
timlichtenberg, line 279: Is the phi_array not update anymore, or is this done somewhere else then?

src/proteus/interior/common.py
```diff
@@ -0,0 +1,143 @@
...
+    numer = 1.0 + _bigphi(phi, par)**par.delta
+    denom = (1.0-_bigf(phi, par)) ** (B_ein*(1-par.phist))
+    return par.dotl * numer / denom
+
+# Path to location at which to save tidal heating array
+def get_file_tides(outdir:str):
+    return os.path.join(outdir, "data", "tides_recent.dat")
+
+# Structure for holding interior variables at the current time-step
+class Interior_t():
```
timlichtenberg, line 67: This is great! For completion, this struct should also hold the `pressure` and `temperature` quantities at basic and staggered nodes. At the moment this is a duplication, but it will be super useful for the integration of a more detailed structural model.

src/proteus/interior/aragog.py
nichollsh: This is now stored in the class `Interior_t` as a numpy array.

src/proteus/interior/common.py
nichollsh: I agree - this is a good suggestion. We could eventually do a similar thing for the atmosphere too. I'll add the arrays you suggest to this PR.

src/proteus/orbit/lovepy.jl
```diff
@@ -0,0 +1,63 @@
...
+                            radius::Array{prec,1},
+                            visc::Array{prec,1},
+                            shear::Array{prec,1},
+                            bulk::Array{prec,1};
+                            ncalc::Int=2000
+                              )::Tuple{Array{Float64,1},Float64,Float64}
+
+    # Internal structure arrays.
+    # First element is the innermost layer, last element is the outermost layer
+    ρ = convert(Vector{prec}, rho)
```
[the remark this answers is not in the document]
nichollsh: It's not ideal to have this Julia code floating around inside PROTEUS. I am hoping in a future PR to tidy up the interface with lovepy.

tests/data/integration/physical/plot_escape.png
timlichtenberg, line 1: This has seemingly affected the mass loss rats by 23 orders of magnitude! Please double check where this comes from, there is perhaps an inconsistency in how the stellar flux is propagated. There are some changes related to that in the escape wrapper.

timlichtenberg (changes requested):
Awesome! Two requested changes in inline comments.

input/default.toml
nichollsh: Yeah, this is consistent with the standard settings. I've just made them explicitly appear in the configuration file so that they don't get lost in the code.

tests/data/integration/physical/plot_escape.png
nichollsh: Good catch! I didn't notice this and will look into it.

src/proteus/interior/common.py
nichollsh: Done

tests/data/integration/physical/plot_escape.png
nichollsh: I think this comes from a unit conversion issue. We previously stored the semimajor axis in AU, via the config. The new changes store it in the helpfile in SI units.

nichollsh:
@timlichtenberg I have resolved the issue with the escape flux as described above in my in-line reply. GitHub no longer allows us to compare the before/after images for this figure, because the files are now identical. I think this is ready to be merged.

lsoucasse (approved):
Changes look good. Why did you need to create an interior object to store the interior output? It seems they were already all included in hf_row variable.

EmmaPostolec:
Hi all, great work @nichollsh ! 👍 

I am a bit late on this PR (sorry) but I was trying to run a grid using grid_proteus.py, and I got an error about the implementation of this cpu_count function at line 268 of the script ( see https://github.com/FormingWorlds/PROTEUS/pull/329/files#diff-496853ffdb18d3a657ed1a6875d29c7086182e02fbd04ab722afbbf2a003a32c )
``` [2025-02-03 17:14:22] Running PROTEUS across parameter grid 'LOV_b'
[2025-02-03 17:14:22] Output path: '/dataserver/users/formingworlds/postolec/PROTEUS/output/LOV_b'
[2025-02-03 17:14:22] Uncaught exception
  File "/dataserver/users/formingworlds/postolec/PROTEUS/tools/grid_proteus.py", line 479, in <module>
    pg.run(90, test_run=False)
  File "/dataserver/users/formingworlds/postolec/PROTEUS/tools/grid_proteus.py", line 268, in run
    num_threads = min(num_threads, os.process_cpu_count())
                                   ^^^^^^^^^^^^^^^^^^^^
AttributeError: module 'os' has no attribute 'process_cpu_count' ``` 

I updated this line with ```os.cpu_count()```instead of ```os.process_cpu_count``` in my grid_proteus.py file. I am wondering if anyone also have this issue (then I can quickly open a branch and a PR to correct it) or if it's just an issue for me  (on the cluster)? Let me know if you also encounter this while running grid_proteus.py :)

PR #333, opened 2025-02-04
Fix os.cpu_count() function in grid_proteus.py for users with Python<3.13

EmmaPostolec:
Hi all, this is related to this https://github.com/FormingWorlds/PROTEUS/pull/329#issuecomment-2634093669 

I updated the grid_proteus.py file to be able to run a grid without having the error about os library while using Pyhton 3.12.2 with my proteus environment as suggested by @nichollsh here https://github.com/FormingWorlds/PROTEUS/pull/329#issuecomment-2634129762

### DIFF
--- a/tools/grid_proteus.py
+++ b/tools/grid_proteus.py
@@ -265,7 +265,7 @@ class Grid():
         num_threads = min(num_threads, self.size)
 
         # do not use more threads than are available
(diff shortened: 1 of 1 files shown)


nichollsh:
Thanks for catching this!

nichollsh (approved):

PR #334, opened 2025-02-11
AGNI no-solve configuration and minor changes

### LINKED ISSUE #331: Store model wall-clock runtime in helpfile
It would be useful for the model to store measures of the performance in the helpfile. These are not physically interesting, but are important for informing the Bayesian Optimisation methods of @Ben-Riegler.

### LINKED ISSUE #332: AGNI no-solve configuration
Add option for running AGNI without solving for T(p). Just prescribe T(p) and get fluxes, for speed.

nichollsh:
- Add option for running AGNI without solver, just prescribing a TP profile. This is very fast and could easily be put into a test or an example for the JOSS paper. Closes #332.
- As part of the above, removes the global `atm` variable into the Proteus class.
- Store runtime info in helpfile. Closes #331.
- Minor fixes to plots and docs.
- Option for disabling the 'strict' check on the convergence criteria, which requires them to be met for two consecutive iterations.
- Time stepping optimisations. The model checks if it is close to approaching solidification / radeqm in order to make sure that the time-step does not exceed the expected point of model convergence. However, this only matters when these criteria are enabled. The model now does not check for this if the criteria are disabled.


input/demos/spider.toml
timlichtenberg, line 1: Why no demo SPIDER file anymore?

input/planets/k218b.toml
timlichtenberg, line 1: I feel we have to pick up the discussion again how to handle the config files. Since the number of parameters keeps rising, it seems to be really annoying to edit all files everytime... :> So the solution may be to instead have one "default" file and store only the parameters that are changed from the default in the specific demo cases. This will require some tweaking of how config files are read. This may actually make a good issue for @HannoSpreeuw to consider sometime in early March. What do you think @lsoucasse?

EmmaPostolec (approved):
Looks good to me, nice job @nichollsh !

input/planets/k218b.toml
EmmaPostolec: Good point @timlichtenberg !

src/proteus/proteus.py
```diff
@@ -107,11 +110,19 @@ def start(self, *, resume: bool = False, offline: bool = False):
             Run in offline mode; do not try to connect to the internet.
         """
 
-        # Import
-        from proteus.atmos_clim import RunAtmosphere
```
timlichtenberg, line 111: Define some common nomenclature in Docs regarding function names etc?

timlichtenberg: They keep changing from CapitalLetter to under_score_naming.

src/proteus/utils/terminate.py
```diff
@@ -149,6 +149,14 @@ def check_termination(handler: Proteus) -> bool:
 
     # Quantify model state at this iteration
     finished = False
+    handler.finished2 = False
```
timlichtenberg, line 152: "finished2" and "finished1" are difficult to parse and understand variable names. Something more descriptive?

tests/helpers/helpers.py
```diff
@@ -5,6 +5,7 @@
 from PIL import Image
 
 PROTEUS_ROOT = Path(__file__).parents[2]
+NEGLECT = ["CH4_mol_atm", "CH4_mol_total", "CH4_kg_atm", "CH4_kg_total", "CH4_bar", "runtime"]
```
timlichtenberg, line 8: Why only these?

input/demos/spider.toml
nichollsh: I created it for demo-ing the tidal heating stuff, but that works well now and can be tested through the other configs if necessary.

input/planets/k218b.toml
nichollsh: I agree. We decided before to not have a default one, because it could be confusing, but it feels like we have to go through and edit a bunch every time someone changes the code.

nichollsh: I think it should be quite easy to do though

src/proteus/proteus.py
nichollsh: The code was originally written with CapitalLetter format, but we later decided that the standard should be lower_case_with_underscores. I think the latter is what Python is written in.

nichollsh: But yes, I agree we should add this in the "contributing" section of the docs.

src/proteus/utils/terminate.py
nichollsh: I'll improve this.

tests/helpers/helpers.py
nichollsh: In this PR I have just moved this functionality to a new file. We should try to keep the list of neglected keys as short as possible. CH4 is included because it's *very* sensitive to the numerics. Runtime is included because it depends on the performance of the computer used.

nichollsh:
Thanks both! I'll make some minor changes and re-request a review.

input/planets/k218b.toml
timlichtenberg: So the idea would be to call a specific .toml file, which only contains the parameters that are different from default.toml, correct? In this case, both the "case.toml" and "default.toml" file should be copy-pasted into the simulation directory, so that afterwards it is crystal clear which parameters were used.

nichollsh: Yeah I think that would be a good solution. As long as it's all compiled into a file that the user can check, it should hopefully avoid issues.

timlichtenberg (approved):
Great, improves the readability and contribution instructions a lot!

PR #336, opened 2025-02-14
Enable/disable real gas EOS in AGNI

### LINKED ISSUE #335: Enable/disable real gas EOS
AGNI now has support for real gas equations of state. These are AQUA for water, Van der Waals for many, and will fall back to the ideal gas EOS in exotic cases. These are combined using Amagat's additive volume law.

PROTEUS would benefit from the ability to turn this on/off.

nichollsh:
AGNI now supports real gas equations of state. It's useful to be able to toggle this via PROTEUS. Closes #335.

Also adds a config validator which disallows `surf_state=skin` when `solve_energy=false`. This combination is invalid, as it would completely decouple the atmosphere and interior temperatures.

This requires the user to update AGNI to version 1.1.0

### DIFF
--- a/input/default.toml
+++ b/input/default.toml
@@ -154,6 +154,7 @@ author  = "Harrison Nicholls, Tim Lichtenberg"
         solution_rtol   = 2e-2          # solver relative tolerance
         overlap_method    = "ee"          # gas overlap method
         condensation    = true          # volatile condensation
+        real_gas        = true          # use real-gas equations of state
 
     [atmos_clim.janus]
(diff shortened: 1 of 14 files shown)


timlichtenberg (approved):
Very cool stuff, do you already have an idea how this changes evolution?

nichollsh:
Nothing clear yet. I think it impacts the convection a bit - generally making it more likely as the critical lapse rate is shallower with these EOS vs ideal gas. It should also change the height too, and therefore the escape rates. I am running some tests on that now.

PR #337, opened 2025-02-17
Atmospheric escape time-step and other changes

nichollsh:
- Do not allow large timesteps as P_surf -> 0
- Update AGNI API with recent changes - set initial T(p) to adiabat
- Update docs with tip for AGNI on Snellius. Also added AGNI install instructions.
- Rework escape/elements plot into a single figure with 2 panels, which is more useful
- Update default config with AGNI no_solve configuration. This is based on L 98-59 b. We can swap out SPIDER for Aragog in this config soon.
- Aragog only logs to file when logging = "DEBUG"


lsoucasse (approved):
Thanks Harrison,
So now default works with parameters of planet L 98-59 b. Would it be worth saving Earth parameters (previous default) into something like planets/earth.toml ?

timlichtenberg:
Yes, please let's have a `earth.toml` in the example config files.

nichollsh:
Okay - sure! I'll bring back an Earth one.

lsoucasse (approved):

nichollsh:
Thanks, Laurent!

PR #340, opened 2025-02-20
Updates to Julia packages

nichollsh:
* The LovePy tidal heating model is now available on my GitHub page, structured as a Julia package. This PR imports it in the 'proper' way, rather than the hack that was used previously.
* AGNI has been updated to _1.2.0_, which is not compatible with the current PROTEUS version. This PR fixes the AGNI interface. This is a result of AGNI now using a RK4 hydrostatic integrator, which requires heights to be replaced with radii.
* Updated the PROTEUS docs to make things a bit clearer for new users.


lsoucasse (approved):
Thanks for these useful updates/simplifications.

PR #342, opened 2025-02-25
Support for H2S and NH3 outgassing

nichollsh:
H2S is an important sulphur-bearing product of outgassing at moderate temperatures and reducing conditions. It can be the main bearer of S atoms in some cases. This PR adds support for H2S, which is calculated by CALLIOPE through `S2 + 2 H2 = 2 H2S`. It should also allow us to better reproduce the putative 'hybrid' H2+H2S atmospheres for L 98-59 d.

NH3 is important for setting the opacity of the atmosphere, and also contains some of the nitrogen atoms. This PR doesn't include NH3 solubility, but does allow it to form through the reaction `N2 + 3 H2 = 2 NH3` through CALLIOPE.

Requires CALLIOPE version `25.02.26`.


EmmaPostolec (approved):
Great work, Harrison ! It looks good to me for merging :)

PR #344, opened 2025-02-28
Diagnostic parameters and minor changes

### LINKED ISSUE #261: Calculate and store WTG parameter, and other diagnostic variables
We model these planets as a 1D column. However, this cannot easily capture dynamical processes such as Coriolis effects and equatorial jets, which would redistribute heat from the substellar point elsewhere.

This could be partially handled by #241. 

However, we can first estimate how reasonable this 1D column approach is by calculating the weak temperature gradient (WTG) parameter (see [1] and [2]). This can be easily calculated and stored in the helpfile. It would allow estimates of the validity of the single-column assumption for different planets, parameters, and points in time. 

This does require the user to provide the planet's rotation rate in the configuration file. In theory this rotation rate would evolve over time due to tides, but we can keep it fixed for now. This can be easily done by allowing the user to choose between various day-orbit resonance ratios (e.g. 1:1 for a synchronous rotator) or by inputting a day length directly.

Other useful diagnostic variables would be: the Roche limit (for comparison with tides) if we include orbital evolution, and the Hill radius (for determining if extended atmospheres around small "cores" become unbound).

1: https://doi.org/10.1146/annurev-fluid-010518-040516
2: https://doi.org/10.1098/rspa.2016.0107

nichollsh:
I have noticed that -- for inflated atmospheres -- the radius of the escaping level can become a large fraction of the Hill radius. This indicates that these atmospheres are close to becoming unbound. It is worth tracking the Hill and Roche radii as diagnostic variables. Similarly, it would be useful to know the WTG parameter in order to estimate the potential for day-night heat redistribution. This PR implements these and closes #261.

The planet's axial period (day length) is now stored as a separate variable to the orbital period. At the moment we set them to be equal, but in the future this might not necessarily be the case. This change comes about because the WTG parameter depends on the axial period of the planet.

This PR also includes a few minor fixes to the plotting functions. In particular, it turns out that some of them were previously failing when executed via the CLI. I have also improved the readability of some of the plots.


timlichtenberg (approved):

nichollsh:
Thank you!

PR #349, opened 2025-03-21
Ls/aragog adiabat

lsoucasse:
This PR changes the default initial temperature field used in Aragog to an adiabatic profile. It requires Aragog version 0.1.11 which features this adiabatic initialisation plus a correction in the core cooling boundary condition.

Test data for the physical integration test have been updated accordingly. You can check in the interior plot that the behaviour is consistent.


timlichtenberg:
@lsoucasse @planetmariana What are the next steps for this PR? Is it ready to be reviewed?

lsoucasse:
> @lsoucasse @planetmariana What are the next steps for this PR? Is it ready to be reviewed?

Hi @timlichtenberg, there is a problem with the test that for some reason fails on the CI vor some OS/python version while it works fine on my laptop. I have no clue at the moment, I need to investigate. You can still anyway review the change, especially look at the interior plots for the integration test which looks good I think.

planetmariana (commented):
thanks @lsoucasse for this change, the interior plots at least for the initial adiabatic profile looks the same as in my simulations. Is there a way to see the final output from the test? Just to check if the simulation solidified completely or if it stops at certain point !

planetmariana (approved):
Looks good @lsoucasse thanks ! at that order of magnitude 1e-8Kg is really small, so setting fully liquid and fully solid within that range it's still valid !

lsoucasse:
> Looks good @lsoucasse thanks ! at that order of magnitude 1e-8Kg is really small, so setting fully liquid and fully solid within that range it's still valid !

It was actually 1e+8 which is big in itself and explains why the test was failing but it is still an almost zero mass compared to the total mantle mass of 1e+24 kg. It arose from the way the solid mantle mass was calculated. I added flag to ensure correct calculation when the total melt fraction gets close to zero or close to one.

nichollsh (approved):

PR #352, opened 2025-03-24
Rework configuration and docs for simplicity

### LINKED ISSUE #350: Simplify configuration files by adding 'default' options
As discussed this morning at the PROTEUS meeting...

It would significantly improve the user experience if we relied on default options for many of the configuration variables. This will constrain the complexity of the code in the future, as we add more physics. It will also be easier for new users to get started with editing the configuration files.

One argument against this is that using defaults would act to 'hide' certain options from the user. However, we can still compile the complete `Config` object into a single file for each run, saved into the model output folder as `init_coupler.toml`.

nichollsh:
The configuration file now uses default variables, which means that the user does not need to supply all variables in for all configurations. Closes #350.

This is better for new users. For example if you want to use AGNI but aren't sure how to set the tolerances, then it will use the default tolerances when solving for T(p).

It's better for maintaining lots of configs. For example, if you only care about using aragog, you will not need to set variables in the SPIDER section. These will remain as None (and not accessed) or as default values.

It's possible to leave out some sections entirely. For example, if you want to use only dummy modules, you won't need to provide sections for AGNI, JANUS, SPIDER, etc.

I have also added more validators into the config parser and updated the docs. There are also some minor performance improvements when using the CLI. Updated dependencies.

Test results were updated because the 'default' parameters are slightly different to the previous parameters. However, you can see from the plots that the changes are very small.

Bumped version to 25.03.28 for release, since this is a fairly major change.


lsoucasse (approved):
Very nice work @nichollsh! It is nice to see smaller toml config files, though keeping the master default config showing all possible options of the code. Thanks.

timlichtenberg:
I have started to review, but before finishing I think we should discuss here what a `default` should be. I think this should not be a `dummy`configuration, but rather the current standard module configuration that provides physical results. For testing a full dummy run, I would offer an example `.toml` file that sets all modules to `dummy` instead. The reason for this is that the typical use case for any user is running a physical simulation. The dummy configuration will only be relevant for testing if the installation works overall, and for debugging. The majority of users will want to run physical simulations, and then have to search for the proper settings for each module individually. What do others think about this?

timlichtenberg:
Ok, then I am not sure how the default is set for the modules. In `src/proteus/config/_atmos_clim.py` for example, line 190:
`module: str = field(validator=in_(('dummy', 'agni', 'janus')))`. Shouldn't there be a default for that?

nichollsh:
Within this new setup for parsing and handling config files there are optional and required variables. All of the `module` variables are required - meaning that there is no default. The user always has to specify which module configuration they want to use. 

We could consider reworking this. Maybe we could make aragog, AGNI, and MORS the default modules?

timlichtenberg:
Ok, that seems fine to me then. We need make double sure that ´default-toml`runs out of the box without errors. This will be a good startup for the JOSS tests as well. Is the aragog-AGNI pair now working for everybody then?

docs/usage.md
```diff
@@ -11,25 +14,18 @@ proteus start --config [cfgfile]
 Where `[cfgfile]` is the path to the required configuration file.
 Pass the flag `--resume` in to resume the simulation from the disk.
 
-You can also run PROTEUS inside a Screen session using:
+See the [config guide](https://fwl-proteus.readthedocs.io/en/latest/config/) for information
+on how to edit the configurations files, and an explanation of their structure.
 
-```console
```
timlichtenberg, line 16: I don't think this should go. Yes, it is explained in `--help`, but these few lines provide an easy entry to running the code. Instead, I would add the equivalent explanation here for `tmux` for the Kapteyn cluster, which most are currently using I guess.

src/proteus/interior/wrapper.py
```diff
@@ -120,7 +120,7 @@ def solve_structure(dirs:dict, config:Config, hf_all:pd.DataFrame, hf_row:dict):
         hf_row["R_int"] = config.struct.radius_int * R_earth
         calculate_core_mass(hf_row, config)
         # initial guess for mass, which will be updated by the interior model
-        hf_row["M_int"] = config.struct.mass_tot * M_earth
+        hf_row["M_int"] = 1.2 * M_earth
```
timlichtenberg, line 123: Why 1.2?

timlichtenberg (approved):
Great many changes! Let's see how this works in practice. :)

docs/usage.md
nichollsh: The `RunPROTEUS.sh` script doesn't actually exist anymore, since we added the command line interface. We could add sections giving brief tutorials on tmux?

src/proteus/interior/wrapper.py
nichollsh: This is an initial guess, so it doesn't really matter too much. It used to use the `mass_tot` parameter as the initial guess, but that's now set to `None` when the structure is defined by the radius, so it cannot be used.

PR #357, opened 2025-04-01
Adding melting curves as a key on toml file

planetmariana:
Adds the melting curves as a key, therefore add a config object. I created a new folder on the 1TPa-dK09-elec-free folder in the OSF repository with three different sets of melting curves: 

1) Wolf_Bower+2018
2)Monteaux+2016 (shifted +600 from solidus to test liquidus and shifted liquidus -600 to test solidus)

In that way, it's possible to run grids using different melting curves that the user can define.

### DIFF
--- a/input/demos/aragog.toml
+++ b/input/demos/aragog.toml
@@ -1,4 +1,22 @@
 # PROTEUS configuration file (version 2.0)
+
+# Root tables should be physical, with the exception of "params"
+# Software related options should go within the appropriate physical table
+
+# The general structure is:
+#   [root]     metadata
+#   [params]   parameters for code execution, output files, time-stepping, convergence
+#   [star]     stellar parameters, model selection
+#   [orbit]    planetary orbital parameters
+#   [struct]   planetary structure (mass, radius)
+#   [atmos]    atmosphere parameters, model selection
+#   [escape]   escape parameters, model selection
+#   [interior] magma ocean model selection and parameters
+#   [outgas]   outgassing parameters (fO2) and included volatiles
+#   [delivery] initial volatile inventory, and delivery model selection
+
+# ----------------------------------------------------
+# Metadata
 version = "2.0"
 author  = "Harrison Nicholls, Tim Lichtenberg"
 
@@ -8,7 +26,7 @@ author  = "Harrison Nicholls, Tim Lichtenberg"
     # output files
     [params.out]
         path        = "dummy_aragog"
-        logging     = "DEBUG"
+        logging     = "INFO"
         plot_mod    = 1      # Plotting frequency, 0: wait until completion | n: every n iterations
         plot_fmt    = "png"  # Plotting image file format, "png" or "pdf" recommended
         write_mod   = 5      # Write CSV frequency, 0: wait until completion | n: every n iterations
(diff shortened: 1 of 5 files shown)


input/demos/aragog.toml
```diff
@@ -92,7 +94,7 @@ author  = "Harrison Nicholls, Tim Lichtenberg"
     rheo_phi_loc    = 0.4    # Centre of rheological transition
     rheo_phi_wid    = 0.15   # Width of rheological transition
     bulk_modulus    = 260e9   # Bulk modulus [Pa]
-
+    melting_dir     ="Monteaux-600"
```
timlichtenberg, line 97:
```suggestion
    melting_dir     = "Monteux-600"
```
Also can you please add a comment in this line.

src/proteus/interior/aragog.py
```diff
@@ -137,6 +137,7 @@ def SetupAragogSolver(config:Config, hf_row:dict, interior_o:Interior_t):
 
     # Get look up data directory, will be configurable in the future
     LOOK_UP_DIR = FWL_DATA_DIR / "interior_lookup_tables/1TPa-dK09-elec-free/MgSiO3_Wolf_Bower_2018/"
+    MELTING_DIR=  FWL_DATA_DIR / "interior_lookup_tables/Melting_curves/"
```
timlichtenberg, line 140: This looks good! It's not clear to me how the user gets these melting curves. Are they automatically downloaded somewhere or do they have to be added by hand? If the latter this would break the `main` branch, as only you so far have these files I believe?

tests/integration/dummy.toml
```diff
@@ -139,6 +139,7 @@ author  = "Harrison Nicholls, Tim Lichtenberg"
     rheo_phi_loc    = 0.4    # Centre of rheological transition
     rheo_phi_wid    = 0.15   # Width of rheological transition
     bulk_modulus    = 260e9   # Bulk modulus [Pa]
+    melting_dir     ="Monteaux-600"
```
timlichtenberg, line 142:
```suggestion
    melting_dir     = "Monteux-600"
```

tests/integration/physical.toml
```diff
@@ -188,7 +188,7 @@ author  = "Harrison Nicholls, Tim Lichtenberg"
     rheo_phi_loc    = 0.4    # Centre of rheological transition
     rheo_phi_wid    = 0.15   # Width of rheological transition
     bulk_modulus    = 260e9   # Bulk modulus [Pa]
-
+    melting_dir     ="Monteaux-600"
```
timlichtenberg, line 191:
```suggestion
    melting_dir     = "Monteux-600"
```

timlichtenberg (changes requested):
Nice, looks good! A couple of potential issues:

- It's not clear to me how the user gets the melting files. Are they already included in the OSF and are automatically downloaded? If that is not the case, they need to be added in this branch before we can merge into main, otherwise the code will break for everyone who does not have these files.
- Can you please add descriptor to the documentation, where it is described what the new parameter does. I presume it should also have a default value, which is simply the "standard" melting curve, whatever you think is the most reasonable standard assumption for now.

nichollsh (changes requested):
Thanks for this Mariana! I think adding the melting curves into the config as a new parameter is a great idea - that way the user can change it much more easily.

It would be useful to add a 'default' value for the curves too.

Similar to Tim's question: I am wondering what happens when the user does not have the folder? We could obtain it from OSF in the same way as the current lookup data.

planetmariana:
@timlichtenberg @nichollsh 

Yes, I updated the OSF repository and data.py (It requires a PR from the aragog side to download everything inside interior_lookup_tables). For now, I would say using liquidus from Monteux+2016 and solidus =liquidusMonteux-600 shows a more reasonable behaviour on the melt fraction. That could be the default for now. What do you think?

nichollsh:
Very nice; thanks for adding this. I can't comment on the validity of the melting curves, but if it works better with Monteux+16 then maybe using that one would be best.

If this requires a PR on the Aragog side, should the Aragog version in the `pyproject.toml` also be updated then? So that the user and the CI are obtaining the right version from PyPI.

lsoucasse (approved):

lsoucasse:
@nichollsh and @timlichtenberg , we have just reviewed the PR with Mariana.

The metling curves are now downloaded separately from the other thermophysics properties and the melting curve data are automatically downloaded from the OSF, thanks to the function Mariana added. There was just a typo in the OSF key idea that prevents the tests from passing. Everything works fine now and there is no need to modify Aragog.

@nichollsh, Let us know whether you want us to merge this PR first or yours first. Whatever option is fine.

nichollsh (approved):

nichollsh:
@planetmariana @lsoucasse nice work. Feel free to merge this today. My branch has a few tweaks left before it's ready.

Issue #358, opened 2025-04-02
Add additional config keys for Aragog


lsoucasse:
For now most Aragog config options are hardcoded. We might think of adding some more keys configurable, following Mariana's work with the melting directory.

Now I think of:
- the directory for lookup thermophysics properties
- the inner boundary condition (core cooling or fixed value)
- the heat fluxes to account for (conduction, convection...)

Do you agree @planetmariana and @timlichtenberg ?

planetmariana:
@lsoucasse Yes, actually I'm working right now on the keys for the  inner/outer boundary conditions, as I'm comparing core cooling with prescribed heat flux. I can also do it for the heat fluxes :) . I guess for the different phases properties should remain in aragog, as they are already linked to the look_up_dir or constant values.

timlichtenberg:
Sounds great! I agree indeed that the core settings and the heat flux on/off conditions should be there (all with defaults).

Let's keep the phase properties for now hardcoded; at some point this will become relevant, but let's do one step after the other.

nichollsh:
@planetmariana is this issue now closed?

lsoucasse:
I guess we can close this issue.

PR #359, opened 2025-04-03
New config keys on toml file from aragog interior model

planetmariana:
Adding new keys on toml file from aragog interior model, as described in [#358](https://github.com/FormingWorlds/PROTEUS/issues/358) . This includes:

- Inner boundary condition (1 = core cooling model, 2=prescribed heat flux, 3 = prescribed temperature)
- Inner boundary value (core temperature [K], if inner_boundary_condition = 3. CMB heat flux [W/m^2], if inner_boundary_condition = 2)
- conduction, convection, gravitational separation and mixing (whether to include or ignore each of them)

Therefore, the user can choose the different configurations on the .toml file

### DIFF
--- a/input/demos/aragog.toml
+++ b/input/demos/aragog.toml
@@ -87,14 +87,14 @@ author  = "Harrison Nicholls, Tim Lichtenberg"
         rate        = 2.0e4         # Bulk unfractionated escape rate [kg s-1]
 
 [interior]
-    grain_size      = 0.1   # crystal settling grain size [m]
-    F_initial       = 1e6   # Initial heat flux guess [W m-2]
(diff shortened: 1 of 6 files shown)


nichollsh (approved):
Very nice. It's good to have these parameters exposed to the user via the config, than hidden in the code.

I'll do the same for SPIDER soon.

lsoucasse (approved):
Good job @planetmariana, I guess we will need to clean the toml files to only expose keys that do not take default values (except in all_options.toml) when we merge with Harrison's branch.

nichollsh:
> when we merge with Harrison's branch.

Yep - definitely. My branch is almost ready, but I am just waiting for the example simulations to be completed with the new default configs.

We can merge this PR first.

PR #361, opened 2025-04-03
Software test for running PROTEUS with AGNI atmosphere model

### LINKED ISSUE #360: AGNI unit test
AGNI is used as the 'default' atmosphere model in the new baseline configuration files.

We should therefore add a unit test which includes the AGNI wrapper functions. The current tests use JANUS and the dummy atmosphere model. The AGNI test would be in addition to these.

Related to #268 and also #324

nichollsh:
Adds a `physical_agni` test for running PROTEUS with AGNI as the atmosphere model with `solve_energy=False`. This complements the use of JANUS in the existing `physical` test config. Includes steps in the GitHub workflow to install Julia and AGNI, with some caching.

Code coverage before: 63%
Code coverage now: 68%

Test speed before: 14 minutes.
Test speed now: 18 minutes.

Closes #360


nichollsh:
This is now ready for review. We can iterate on it in the future, but I think this current formulation serves as a decent baseline for a test which includes AGNI.

timlichtenberg (approved):
Very useful!

nichollsh:
Thanks both!

PR #363, opened 2025-04-04
Store Fxuv and use correct distance scaling

### LINKED ISSUE #362: Store Fxuv in helpfile
PROTEUS obtains Fxuv from MORS for calculating escape rates. However, this value is not stored in the helpfile. It's important to have these for analysis of simulation outputs.

Also, the current formulation calculates Fxuv from Lxuv using the semimajor axis. However, it should probably use the `separation` variable to do this, which accounts for the orbital eccentricity.

nichollsh:
* Moves Fxuv calculation into star wrapper
* Uses correct distance for calculating Fxuv from Lxuv
* Stores Fxuv in helpfile for later analysis
Closes #362 

The new distance scaling accounts for the eccentricity, which means that the planet is on-average slightly further away from the star. This results in slightly lower escape rates.

Before: `1254793.79, 1254793.79, 1254792.59, 1254784.36, 1253588.73, 1252449.58, ...`
After: `1242829.01, 1242829.01, 1242827.82, 1242819.67, 1241635.44, 1240507.15, ...`


timlichtenberg (approved):
Good for me, let's wait for @EmmaPostolec's review as well.

EmmaPostolec (approved):
Hi Harrison, great changes ! Indeed, I think it makes more sense to have Fxuv computation in the star wrapper and also compute it with separation instead of semi-major axis

Issue #206, opened 2024-10-16
Robust file dependency download process


nichollsh:
Sometimes the tests fail because the model is not able to get data from OSF. This can usually be fixed by requesting the test to be run again, in which case the data usually downloads fine.

Example of it failing: https://github.com/FormingWorlds/PROTEUS/actions/runs/11366750494/job/31617828810

We could include a loop in the downloader functions that re-attempts the download if an error is thrown.

timlichtenberg:
Issue splits into two issues
- [ ] Download sometimes fails completely and throws an exception because of server-side issues.
- [ ] Sometimes it just doesn't download everything, and just a few files or one file is missing. Need some sort of registry or file inventory.

First issue is probably easier to solve than second one.

nichollsh:
Regarding issue B, we could potentially query OSF for the files we expect to download. This can be done fairly easily with curl.

For example, the node ID for the spectra files is `vehxg`, so we can obtain a list of all the files with:
`curl -X GET https://api.osf.io/v2/nodes/vehxg/files/osfstorage/`

This provides a JSON dataset which lists all of the folders and files within the node. We can the recursively follow the folders to make a tree of files. For each file, OSF also provides an MD5 hash that can be used to verify that all the files downloaded fine.

In the same JSON dataset, we also get download links for each file. We could also consider using this as an alternative to the current `osfclient` library which hasn't been updated in over 4 years.

nichollsh:
Working example in the notebook here:

[request.ipynb.zip](https://github.com/user-attachments/files/18478163/request.ipynb.zip)

PR #366, opened 2025-04-05
Integrate offline chemistry into main codebase

### LINKED ISSUE #365: Move offline chemistry into main codebase
The current offline VULCAN chemistry interface is in `tools/`. It is a 'hack' onto the main part of the code. This means we cannot include validator, specify parameters in the config files, or add tests..

We should move the `offline_chemistry.py` functionality into the `atmos_chem` module of PROTEUS. At the same time, we can add a config section for handling this. 

Instead of editing and then running `tools/offline_chemistry.py`, the user could run the command line argument `proteus offchem -c [cfgfile]`.

nichollsh:
## Summary
The offline chemistry functionality aims to postprocess PROTEUS output data with the VULCAN chemical kinetics model. In this sense, the atmospheric chemistry is calculated 'offline' from the main simulations. It provides insight into what a more realistic composition might be, when shaped by disequilibrium processes. These chemistry calculations can be quite slow, and therefore currently impractical for integrating into the main loop.

However, an interface to this offline formulation is useful. This is can currently be done by editing and then running the script `tools/offline_chemistry.py`. 

This PR integrates the offline chemistry functionality into PROTEUS more directly, by controlling it via the `Config` class. This ensures easy configuration, validation of options, parameter transparency, and also allows unit tests. Closes #365.

To use the offline chemistry functionality:
1. Run PROTEUS as usual: `proteus start -c path/to/config.toml`
2. Run the chemistry via the CLI: `proteus offchem -c `path/to/config.toml`

Test coverage before: 72%
Test coverage now: 72% (unchanged)

## Warnings
This adds a new config section `atmos_chem`, meaning that all existing config files will need to be updated with a section stub for compatibility. However, we should not have to back-update config files when adding other chemical models in the future.

## Next steps
In the future we could consider having the chemistry run automatically at the end of a PROTEUS simulation. This would occur after the final time step, and before synthetic observations are performed.


input/all_options.toml
```diff
@@ -284,6 +284,31 @@ author  = "Harrison Nicholls, Tim Lichtenberg, Mariana Sastre"
...
+    updraft_const   = 0.0       # Set constant updraft velocity
+
+    # Vulcan-specific atmospheric chemistry parameters
+    [atmos_chem.vulcan]
+        clip_fl = 1e-20     # Floor on stellar spectrum [erg s-1 cm-2 nm-1]
+        clip_vmr = 1e-10    # Neglect species with vmr < clip_vmr
+        make_funs = true    # Generate reaction network functions
+        ini_mix = "profile" # Initial mixing ratios (profile, outgas)
+        fix_surf = false    # Fixed surface mixing ratios
+        network = "CHO"     # Class of chemical network to use (CHO, NCHO, SNCHO)
```
timlichtenberg, line 307: Should the generic network be SNCHO, as a "typical" simulation may include all these species? Or are these networks substantially slower?

timlichtenberg (approved):
Looks good to me, nice work. Indeed, adding a setting to automatically postprocess a simulation at the end to then generate the observables, and batch processing this important next developments.

input/all_options.toml
nichollsh: Yeah, it probably should use the SNCHO network as the default. 
I copied this line from the `physical_agni` test, which uses CHO for performance reasons.

nichollsh:
Thanks Tim! I've changed the default network in `all_options.toml` so I think this PR is ready to be merged.

nichollsh:
Although the tests just failed when downloading data from OSF...
Relates to issue #206.

timlichtenberg (approved):
👍

lsoucasse (approved):
Looks good, do you understand the issue with the doc build?

lsoucasse:
> Looks good, do you understand the issue with the doc build?

OK, it was just a download issue that a rebuilt fixed. Ready to merge @nichollsh

Issue #348, opened 2025-03-19
Crash when element inventory is small as a result of escape


EmmaPostolec:
This issue is trying to solve the following error when PROTEUS is computing atmospheric escape :

``` [ INFO  ] Running EL escape (ZEPHYRUS) ...
[ INFO  ]     age_star = 1.0e+02 Myr
[ INFO  ]     F_xuv    = 1.5e+03 W m-2
[ INFO  ] Bulk escape rate: 7.03e+16 kg yr-1 = 2.23e+09 kg s-1
Uncaught exception
Traceback (most recent call last):
  File "/Users/users/postolec/data/miniconda3/envs/proteus/bin/proteus", line 8, in <module>
    sys.exit(cli())
             ^^^^^
  File "/Users/users/postolec/data/miniconda3/envs/proteus/lib/python3.12/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/users/postolec/data/miniconda3/envs/proteus/lib/python3.12/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/Users/users/postolec/data/miniconda3/envs/proteus/lib/python3.12/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/users/postolec/data/miniconda3/envs/proteus/lib/python3.12/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/users/postolec/data/miniconda3/envs/proteus/lib/python3.12/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/dataserver/users/formingworlds/postolec/PROTEUS/src/proteus/cli.py", line 88, in start
    runner.start(resume=resume, offline=offline)
  File "/dataserver/users/formingworlds/postolec/PROTEUS/src/proteus/proteus.py", line 371, in start
    RunEscape(self.config, self.hf_row, self.interior_o.dt, self.stellar_track)
[...]

PR #375, opened 2025-04-08
Allow bare-rock simulation mode after atmosphere has escaped

EmmaPostolec:
This PR aimed to have a warning when M_vols is small to avoid getting the following error : 
``` 
[ INFO  ] ---------------------------------------------------
[ INFO  ] Running EL escape (ZEPHYRUS) ...
[ INFO  ] Bulk escape rate: 6.96e+16 kg yr-1 = 2.20e+09 kg s-1
[ CRITICAL ] Uncaught exception
  File "/Users/users/postolec/data/miniconda3/envs/proteus/bin/proteus", line 8, in <module>
    sys.exit(cli())
             ^^^^^
  File "/Users/users/postolec/data/miniconda3/envs/proteus/lib/python3.12/site-packages/click/core.py", line 1157, in __call__
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/users/postolec/data/miniconda3/envs/proteus/lib/python3.12/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/Users/users/postolec/data/miniconda3/envs/proteus/lib/python3.12/site-packages/click/core.py", line 1688, in invoke
  File "/Users/users/postolec/data/miniconda3/envs/proteus/lib/python3.12/site-packages/click/core.py", line 1434, in invoke
  File "/Users/users/postolec/data/miniconda3/envs/proteus/lib/python3.12/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/dataserver/users/formingworlds/postolec/PROTEUS/src/proteus/cli.py", line 89, in start
    runner.start(resume=resume, offline=offline)
  File "/dataserver/users/formingworlds/postolec/PROTEUS/src/proteus/proteus.py", line 375, in start
    RunEscape(self.config, self.hf_row, self.interior_o.dt, self.stellar_track)
  File "/dataserver/users/formingworlds/postolec/PROTEUS/src/proteus/escape/wrapper.py", line 50, in RunEscape
    solvevol_target = calc_new_elements(hf_row, dt, config.escape.reservoir)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/dataserver/users/formingworlds/postolec/PROTEUS/src/proteus/escape/wrapper.py", line 134, in calc_new_elements
ZeroDivisionError: float division by zero
```
This error was obtained trying to model the TOI561b planet using similar parameters as the earth simulation. Neverthelesss, this is an extreme case with a = 0.0106 AU orbiting Sun-like star.

This PR closes the issue #348.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -24,7 +24,7 @@ jobs:
         python-version: ['3.11', '3.13']
         include:
           - os: ubuntu-latest
-            INSTALL_DEPS: sudo apt-get update; sudo apt-get install libnetcdff-dev netcdf-bin
+            INSTALL_DEPS: sudo apt-get update; sudo apt-get install libnetcdff-dev netcdf-bin libssl-dev
             CC: gcc
             CXX: g++
             FC: gfortran
@@ -137,7 +137,8 @@ jobs:
           key: ${{ env.pythonLocation }}-${{ hashFiles('pyproject.toml') }}
 
       - name: Install Python dependencies
-        run: python -m pip install -e .[develop]
+        run:
+          python -m pip install -e .[develop]
 
       # Get lookup data for PROTEUS
       - name: Get FWL data from cache
@@ -165,7 +166,7 @@ jobs:
         with:
           name: integration-physical
           path: |
-            output/physical/plot_*.png
+            output/physical/plot/plot_*.png
             output/physical/runtime_helpfile.csv
 
       - name: Report coverage
--- a/input/planets/l9859d.toml
+++ b/input/planets/l9859d.toml
@@ -9,7 +9,7 @@ author  = "Harrison Nicholls, Tim Lichtenberg"
         logging     = "INFO"
         plot_mod    = 40      # Plotting frequency, 0: wait until completion | n: every n iterations
         write_mod   = 2
-        archive_mod = 5
+        archive_mod = 0
         remove_sf   = true
 
     # time-stepping
--- a/src/proteus/observe/platon.py
+++ b/src/proteus/observe/platon.py
@@ -12,6 +12,7 @@ if TYPE_CHECKING:
 
 log = logging.getLogger("fwl."+__name__)
 
+PLATON_TLIMS = (100.0+0.5, 4000.0-0.5)
 PLATON_METHOD="xsec"
 PLATON_GASES=("H2", "H", "He", "H2O", "CH4", "CO", "CO2", "O", "C", "N", "NH3", "N2",
                 "O2", "O3", "H2S", "HCN", "NO", "NO2", "OH", "PH3", "SiO", "SO2", "TiO",
@@ -151,7 +152,7 @@ def _get_ptr(atm:dict):
     rad = np.array(atm["rl"]) # m
     if prs[1] < prs[0]:
         prs = prs[::-1]
-        tmp = tmp[::-1]
+        tmp = np.clip(tmp[::-1], a_min=PLATON_TLIMS[0], a_max=PLATON_TLIMS[1])
         rad = rad[::-1]
     return prs, tmp, rad
 
--- a/src/proteus/proteus.py
+++ b/src/proteus/proteus.py
@@ -484,6 +484,10 @@ class Proteus:
 
             ############### / HOUSEKEEPING AND CONVERGENCE CHECK
 
+        # Write conditions at the end of simulation
+        log.info("Writing data")
+        WriteHelpfileToCSV(self.directories["output"], self.hf_all)
+
         # Run offline chemistry
         if self.config.atmos_chem.when == "offline":
             log.info(" ")
@@ -501,10 +505,6 @@ class Proteus:
                 log.warning("Cannot observe planet after desiccation")
             else:
                 run_observe(self.hf_row, self.directories["output"], self.config)
-
-        # Write conditions at the end of simulation
-        log.info("Writing data and plots")
-        WriteHelpfileToCSV(self.directories["output"], self.hf_all)
 
         # Make final plots
         if self.config.params.out.plot_mod is not None:
--- a/src/proteus/utils/coupler.py
+++ b/src/proteus/utils/coupler.py
@@ -503,7 +503,7 @@ def WriteHelpfileToCSV(output_dir:str, current_hf:pd.DataFrame):
         os.remove(fpath)
 
     # write new file
-    current_hf.to_csv(fpath, index=False, sep="\t", float_format="%.8e")
+    current_hf.to_csv(fpath, index=False, sep="\t", float_format="%.10e")
     return fpath
 
 def ReadHelpfileFromCSV(output_dir:str):
--- a/src/proteus/utils/data.py
+++ b/src/proteus/utils/data.py
@@ -5,6 +5,7 @@ import logging
 import os
 import subprocess as sp
 from pathlib import Path
+from time import sleep
 from typing import TYPE_CHECKING
 
 import platformdirs
@@ -62,9 +63,10 @@ def download(
     folder: str,
     target: str,
     osf_id: str,
-    desc: str
-
-):
+    desc: str,
+    max_tries: int = 3,
+    wait_time: float = 5,
+) -> bool:
     """
     Generic download function.
 
@@ -78,6 +80,15 @@ def download(
         OSF project id
     desc: str
         Description for logging
+    max_tries: int
+        Number of tries to download the file
+    wait_time: float
+        Time to wait between tries
+
+    Returns
+    -------
+    bool
+        True if the file was downloaded successfully, False otherwise
     """
     log.debug(f"Get {desc}?")
 
@@ -87,9 +98,22 @@ def download(
     if not (data_dir / folder).exists():
         storage = get_osf(osf_id)
         log.info(f"Downloading {desc} to {data_dir}")
-        download_folder(storage=storage, folders=[folder], data_dir=data_dir)
+        for i in range(max_tries):
+            log.debug(f"    attempt {i+1}")
+            try:
+                download_folder(storage=storage, folders=[folder], data_dir=data_dir)
+                break
+            except RuntimeError as e:
+                log.warning(f"    {desc} download failed: {e}")
+                if i < max_tries - 1:
+                    log.info(f"    Retrying in {wait_time} seconds...")
+                    sleep(wait_time)
+                else:
+                    log.error(f"    Failed to download {desc} after {max_tries} attempts")
+                    return False
     else:
         log.debug(f"    {desc} already exists")
+    return True
 
 
 def download_surface_albedos():
--- a/tools/grid_proteus.py
+++ b/tools/grid_proteus.py
@@ -536,7 +536,7 @@ if __name__=='__main__':
     use_slurm = True
 
     # Execution limits
-    max_jobs = 190       # maximum number of concurrent tasks
+    max_jobs = 300       # maximum number of concurrent tasks
     max_days = 1         # maximum number of days to run
     max_mem  = 3         # maximum memory per CPU in GB
 


nichollsh:
It might be related to this 10 Gyr estimated age, yes. Do you get the same error with a smaller current age (such as 7 Gyr)?

EmmaPostolec:
> It might be related to this 10 Gyr estimated age, yes. Do you get the same error with a smaller current age (such as 7 Gyr)?

Thanks for the hint, Harrison ! Good point I am testing this right now let see how it goes :)

nichollsh:
> Does it means all volatiles have escape from the atmosphere?

It looks like it. Are you using Zephyrus?

EmmaPostolec:
> > Does it means all volatiles have escape from the atmosphere?
> 
> It looks like it. Are you using Zephyrus?

Yes I can test it also with dummy escape or no escape at all ?

nichollsh:
That's interesting. It does seem like a very high escape rate. What is R_xuv in this case?

nichollsh:
I think related to this, I sometimes seem to get an error where CALLIOPE is unable to find a solution for the outgassing:

```text 
...
  File "/home2/p321409/Projects/CALLIOPE/src/calliope/solve.py", line 449, in equilibrium_atmosphere
    raise Exception("Could not find solution for volatile abundances (max attempts reached)")
Exception: Could not find solution for volatile abundances (max attempts reached)
```

This occurs when the atmosphere is almost entirely escaped, although in this case it is still as large as 90 bar. I suppose the small volatile inventories make finding a solution for the partial pressures tricky.

timlichtenberg:
What's the status on this @EmmaPostolec? :)

nichollsh:
Also related to this, I have noticed some strange systematic behaviour in PROTEUS for escaped cases, which makes them 'appear' to have low densities inconsistent with their lack of atmosphere.

This is because P_surf can go from relatively large (~10s of bar) to <1 bar within the final time-step. However, PROTEUS does not recalculate the atmosphere structure in this case because there is presumably no atmosphere to model.

This means that the final row of the helpfile contains the *previous* value for R_obs and rho_obs, which are for P>1 bar.

EmmaPostolec:
Hi both, sorry I am coming back on this after a long time. 

@nichollsh : 

> It might be related to this 10 Gyr estimated age, yes. Do you get the same error with a smaller current age (such as 7 Gyr)?
> 
> The graph seems to indicate that the model is only evolving up to ~2 Gyr, so it's not clear to me why this error is being encountered.

- I tested simulations with Zephyrus escape and epsilon = 1.0 with star.mors.age_no = 1.5 and 7 Gyr to see if I would still get the ``` ZeroDivisionError: float division by zero``` error. 


> That's interesting. It does seem like a very high escape rate. What is R_xuv and F_xuv in this case?

- F_xuv      = 1.47e+03   W m-2 and R_xuv = 8.74e+06 m, both quite high !

> ```
> 
> ```
> 
> 
>     
>   
> 
> This occurs when the atmosphere is almost entirely escaped, although in this case it is still as large as 90 bar. I suppose the small volatile inventories make finding a solution for the partial pressures tricky.

- I am a bit confused : so in this case the atmosphere is supposed to be entirely escaped but you still have it at 90 bar ? Is this a numerical error due to the small volatile inventories ? I see in proteus.py lines 400-409 : 

```             # Check for when atmosphere has completely escaped (P_surf < 0.01)
            #    This will mean that the mixing ratios become undefined, so use value of 0.
            if self.hf_row["P_surf"] < 0.01:
                self.has_escaped = True
                for gas in gas_list:
                    self.hf_row[gas+"_vmr"] = 0.0
                self.hf_row["atm_kg_per_mol"] = 0.0

            ############### / OUTGASSING
``` So if I understand correctly, those lines set the quantity of all atmospheric species to 0. But those species can still be present in the magma ocean in small quantities leading to small values of partial pressures ? In this case should I include a check with if volatiles species abundances < threshold then set all volatiles abundances to 0 to avoid this error in CALLIOPE ? Maybe this is not the good way to go but do you have any idea/suggestions on this ? I am not sure how to fix this error. 

> Also related to this, I have noticed some strange systematic behaviour in PROTEUS for escaped cases, which makes them 'appear' to have low densities inconsistent with their lack of atmosphere.
> 
> This is because P_surf can go from relatively large (~10s of bar) to <1 bar within the final time-step. However, PROTEUS does not recalculate the atmosphere structure in the final time-step because there is presumably no atmosphere to model.
> 
> This means that the final row of the helpfile contains the _previous_ value for R_obs and rho_obs, which are for P>1 bar.

- In the simulations I run I had this setting : 
```
        [params.stop.escape]
            enabled   = false
            p_stop    = 1.0     # bar, model will terminate with p_surf < p_stop

```
So the simulation was still running after all the atmosphere escape. Maybe I shoud turn this to true to avoid this problem ? 
For this problem should I set a piece of code that set the density to 0 if the atmosphere is gone ?

**Comment by timlichtenberg** [comment]
>            # Check for when atmosphere has completely escaped (P_surf < 0.01)
>             #    This will mean that the mixing ratios become undefined, so use value of 0.
>             if self.hf_row["P_surf"] < 0.01:
>                 self.has_escaped = True
>                 for gas in gas_list:
>                     self.hf_row[gas+"_vmr"] = 0.0
>                 self.hf_row["atm_kg_per_mol"] = 0.0
> 
>             ############### / OUTGASSING
> ```
> 
> So if I understand correctly, those lines set the quantity of all atmospheric species to 0. But those species can still be present in the magma ocean in small quantities leading to small values of partial pressures ? In this case should I include a check with if volatiles species abundances < threshold then set all volatiles abundances to 0 to avoid this error in CALLIOPE ? Maybe this is not the good way to go but do you have any idea/suggestions on this ? I am not sure how to fix this error.
> 

My suggestion for this would possibly be to completely disable CALLIOPE when the atmosphere has escaped, there may still be value in keeping the simulation running, even if `[params.stop.escape] enabled   = false`. This may, however, invite other errors with AGNI or JANUS.

> > Also related to this, I have noticed some strange systematic behaviour in PROTEUS for escaped cases, which makes them 'appear' to have low densities inconsistent with their lack of atmosphere.
> > This is because P_surf can go from relatively large (~10s of bar) to <1 bar within the final time-step. However, PROTEUS does not recalculate the atmosphere structure in the final time-step because there is presumably no atmosphere to model.
> > This means that the final row of the helpfile contains the _previous_ value for R_obs and rho_obs, which are for P>1 bar.
> 
> * In the simulations I run I had this setting :
> 
> ```
>         [params.stop.escape]
>             enabled   = false
>             p_stop    = 1.0     # bar, model will terminate with p_surf < p_stop
> ```
> 
> So the simulation was still running after all the atmosphere escape. Maybe I shoud turn this to true to avoid this problem ? For this problem should I set a piece of code that set the density to 0 if the atmosphere is gone ?

Yes, this is probably the simplest solution for now. Can you test if this works?

**Comment by EmmaPostolec** [comment]
> My suggestion for this would possibly be to completely disable CALLIOPE when the atmosphere has escaped, there may still be value in keeping the simulation running, even if `[params.stop.escape] enabled = false`. This may, however, invite other errors with AGNI or JANUS.

Maybe I can wait for @nichollsh idea before including this in proteus.py or the outgas wrapper.py ? ? To see how it cill behave with AGNI/JANUS ?

> Yes, this is probably the simplest solution for now. Can you test if this works?

Sure, I launched toi561b.toml with `[params.stop.escape] enabled = true` and star.mors.age_now = 10 Gyr and here is what I got :  
Screenshot 2025-04-18 at 5 36 27 PM
Screenshot 2025-04-18 at 5 36 58 PM

The last loop of the simulation before final plots : 
```
[ INFO  ] Loop counters
[ INFO  ] init      total
[ INFO  ] 3/3     0062/9000 
[ INFO  ] ---------------------------------------------------
[ INFO  ] Evolve interior...
[ DEBUG ] Using spider module to evolve interior
[ DEBUG ] Attempt 1
[ DEBUG ] Previous step size: 1.94e+02 yr
[ INFO  ] Time-stepping intent: slow down
[ DEBUG ] Escape expected in 3.297e+02 yrs
[ INFO  ] New time-step target is 3.01e+02 years
[ DEBUG ] SPIDER iteration: dt=3.01e+02 yrs in 1 steps (at i=60)
[ DEBUG ] SPIDER output suppressed
[ DEBUG ] Attempt 1 succeeded
[ INFO  ]     T_magma    = 2507.111 K
[ INFO  ]     Phi_global = 0.993  
[ INFO  ]     RF_depth   = 0.448  
[ INFO  ]     F_int      = 2.23e+05 W m-2
[ INFO  ]     F_tidal    = 0.00e+00 W m-2
[ INFO  ]     F_radio    = 3.83e-01 W m-2
[ INFO  ] ---------------------------------------------------
[ INFO  ] Evolve orbit and tides...
[ INFO  ]     period = 0.444 days
[ INFO  ] ---------------------------------------------------
[ INFO  ] Stellar flux management...
[ DEBUG ] Update stellar radius and mass
[ DEBUG ] Update stellar fluxes and temperature
[ INFO  ]     F_ins      = 3.34e+06   W m-2
[ INFO  ]     F_xuv      = 1.47e+03   W m-2
[ INFO  ]     T_star     = 4970.45    K
[ DEBUG ] Update equilibrium temperature
[ INFO  ] Updated spectrum not required
[ INFO  ] ---------------------------------------------------
[ INFO  ] Running EL escape (ZEPHYRUS) ...
[ INFO  ] Bulk escape rate: 9.26e+16 kg yr-1 = 2.94e+09 kg s-1
[ DEBUG ]      H (bulk) mass ratio = 2.30e-01 
[ DEBUG ]      C (bulk) mass ratio = 2.30e-01 
[ DEBUG ]      N (bulk) mass ratio = 4.25e-02 
[ DEBUG ]      S (bulk) mass ratio = 4.97e-01 
[ DEBUG ]     Si (bulk) mass ratio = 0.00e+00 
[ DEBUG ]     Mg (bulk) mass ratio = 0.00e+00 
[ DEBUG ]     Fe (bulk) mass ratio = 0.00e+00 
[ DEBUG ]     Na (bulk) mass ratio = 0.00e+00 
[ INFO  ] Elemental escape fluxes:
[ INFO  ]      H = 2.13e+16 kg yr-1
[ INFO  ]      C = 2.13e+16 kg yr-1
[ INFO  ]      N = 3.94e+15 kg yr-1
[ INFO  ]      S = 4.60e+16 kg yr-1
[ INFO  ]     Si = 0.00e+00 kg yr-1
[ INFO  ]     Mg = 0.00e+00 kg yr-1
[ INFO  ]     Fe = 0.00e+00 kg yr-1
[ INFO  ]     Na = 0.00e+00 kg yr-1
[ INFO  ] ---------------------------------------------------
[ INFO  ] Solving for equilibrium partial pressures at surface
[ DEBUG ]     target masses: {'H': 3.7815804378067487e+18, 'C': 3.781580402400333e+18, 'N': 6.973372243823004e+17, 'S': 8.152948618158595e+18, 'Si': 0.0, 'Mg': 0.0, 'Fe': 0.0, 'Na': 0.0}
[ DEBUG ]     Initial guess attempt number = 31
[ DEBUG ]     Residuals: [-33010176.0, 22186496.0, -4516992.0, 3919872.0]
[ INFO  ]     H2O    : 0.00     bar (5.85e-05 VMR)
[ INFO  ]     CO2    : 0.04     bar (6.89e-02 VMR)
[ INFO  ]     H2     : 0.00     bar (1.19e-04 VMR)
[ INFO  ]     CH4    : 0.00     bar (3.71e-17 VMR)
[ INFO  ]     CO     : 0.49     bar (8.52e-01 VMR)
[ INFO  ]     N2     : 0.05     bar (7.90e-02 VMR)
[ INFO  ]     S2     : 0.00     bar (1.03e-09 VMR)
[ INFO  ]     SO2    : 0.00     bar (9.47e-06 VMR)
[ INFO  ]     H2S    : 0.00     bar (9.82e-10 VMR)
[ INFO  ]     NH3    : 0.00     bar (7.16e-10 VMR)
[ INFO  ]     total  : 0.58     bar
[ INFO  ]     mmw    : 29.1100  g mol-1
[ INFO  ] ---------------------------------------------------
[ INFO  ] Running JANUS...
[ DEBUG ] Setting stratosphere to 0.50 K
[ DEBUG ] Will run socrates inside '/tmp/proteus_5833280075366/'
[ INFO  ] Water clouds have formed = False
[ INFO  ] SOCRATES fluxes (net@surf, net@TOA, OLR): 2.01035e+05, 2.79259e+05, 9.78852e+05 W m-2
[ INFO  ] ---------------------------------------------------
[ DEBUG ] Extending helpfile with new row
[ DEBUG ] Writing helpfile to CSV file
[ INFO  ] Runtime info...
[ INFO  ]     Wall time  = 2025-04-18 17:12:44   
[ INFO  ]     Model time = 5.50e+04   yr
[ INFO  ]     T_surf     = 2507.111   K
[ INFO  ]     T_magma    = 2507.111   K
[ INFO  ]     P_surf     = 5.75e-01   bar
[ INFO  ]     Phi_global = 9.93e-01   
[ INFO  ]     F_atm      = 2.79e+05   W m-2
[ INFO  ]     F_int      = 2.23e+05   W m-2
[ INFO  ] Checking convergence criteria
[ DEBUG ] Check keepalive file
[ DEBUG ] Check escape
[ DEBUG ]     val, req = 5.755e-01, 1.000e+00  bar
[ INFO  ] 
[ INFO  ] ===> Atmosphere has escaped! <===
``` 
So indeed setting `[params.stop.escape] enabled = true` stops the situation after the entire atmosphere has escaped. But I am thinking maybe we want the simulation to keep going without the atmosphere in the future ?

timlichtenberg:
Yes, we would want to keep the simulation going without the atmosphere.

nichollsh:
I agree with both of you that it would be ideal to run simulations post-escape. The termination criterion on P_surf is the most physical/intuitive, so maybe we should leave it in.

When the escape termination criterion is turned off, so the simulation continues post-escape, we could check for when the volatile inventory (kg) drops below some critical value and then switch to some post-escape simulation mode (as I think Emma suggests above). Maybe a critical value of 1e8 kg would work by default, but should be included in config file. 

In this post-escape mode we could set the atmosphere module to the `dummy` with `gamma=0`, which should be equivalent to bare rock thermal emission, and also disable the outgassing check. It would also be consistent to set `R_obs = R_int` and the `*_total_kg` values to zero at that point. 

We will need to rethink this when volcanic degassing is included.

EmmaPostolec:
> I agree with both of you that it would be ideal to run simulations post-escape. The escape termination criterion based on P_surf < P_stop is the most physical/intuitive, so maybe we should leave it in.
> 
> When the escape termination criterion is turned off, so the simulation continues post-escape, we could check for when the volatile inventory (kg) drops below some critical value and then switch to some post-escape simulation mode (as I think Emma suggests above). Maybe a critical value of 1e8 kg would work by default, but should be included in config file.
> 
> In this post-escape mode we could set the atmosphere module to the `dummy` with `gamma=0`, which should be equivalent to bare rock thermal emission, and also disable the outgassing step. It would also be consistent to set `R_obs = R_int` and the `*_total_kg` values to zero at that point.
> 
> We will need to rethink this when volcanic degassing is included. It also won't capture the behaviour of retaining a tenuous atmosphere, since the escape rate would presumably tend to zero as the atmosphere is stripped. Probably not too important since we cannot measure thin atmospheres anyway.

Thanks a lot for the greaat suggetions @nichollsh it helps a lot ! Now I see what I should change to make it work. We can discuss this during the PROTEUS meeting today. Thanks for opening the issue @timlichtenberg

PR #377, opened 2025-04-09
Calculate synthetic observations based on offline chemistry output

### LINKED ISSUE #376: Generate synthetic spectra from offline chemistry output
At the moment we can do the following things to postprocess PROTEUS outputs:
1. Calculate atmospheric composition with chemical kinetics (VULCAN)
2. Calculate transmission/emission spectra (platon)

However, it's not possible to pass the results from (1) into the calculation of (2). This would be very useful in general, as well as for the current paper, as the spectra would then show features associated with the physics in VULCAN.

nichollsh:
* Options for running synthetic observations on the results of offline chemistry, as well as using the AGNI NetCDF profiles, or just based on the outgassing calculation.
* Support for accessing synthetic observations functionality via CLI.
* Generalise (and tidy up) synthetic observations wrapper
  - Groundwork for future support of telescope simulators
* Add new plot for atmospheric composition
* Add offline chemistry output into tests (compares content of DataFrames)

Closes #376.

These changes are motivated by the current paper in-prep, however they could be extended further in the future and be useful for other projects. For example, we might want to add a telescope simulator into the pipeline.

It's still highly discouraged to run VULCAN in parallel on the same machine, as VULCAN edits its own source code during execution. We should probably address that in a new Issue soon (e.g. https://github.com/nichollsh/VULCAN/issues/1)

Test coverage before: 72%
Test coverage now: 71%
We should look into adding the synthetic observations into the tests.


tests/data/integration/physical/plot_interior.png
timlichtenberg, line 1: There is some weird behaviour going on with the lower mantle viscosity. Potentially not important, but shouldn't be different in principle?

timlichtenberg (approved):
Looks good overall, though there is a change in the standard plot lower mantle viscosity; unclear where this should come from.

EmmaPostolec (approved):
Looks great, this is really nice work Harrison ! This will be super helpful when working with observations :)

tests/data/integration/physical/plot_interior.png
planetmariana: That comes from the lookup tables, the current tables are up to 135GPa, after that it performs an interpolation, it should go when extending the tables or changing the core fraction.

input/all_options.toml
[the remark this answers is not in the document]
nichollsh: This would be an option for the future, yes. Currently not implemented.

nichollsh:
Thanks all :)

Issue #355, opened 2025-03-31
Expand PROTEUS grid functionality for queueing manager and usage on Habrok & Snellius


timlichtenberg:
In order to perform larger simulation grids, we need a method to execute them using a queueing manager, such as Slurm. Specifically, we want to execute the code on Habrok (the RUG clsuter), and Snellius.

Currently, [`grid_proteus.py`](https://github.com/FormingWorlds/PROTEUS/blob/45b361fac78b1a85805eb006297c5fa82a196d1e/tools/grid_proteus.py#L4) handles this manually, but is set up specifically for being executed by hand. This runs reasonably well on the Kapteyn machines norma1/2, but it is limited in usage with queueing managers and also limited to one node.

This is an urgent tissue, as it is currently one of the main limiting factors of deploying the code on larger scale.

Isses to tackle are:
- [x] Get guest accounts for Harrison, Laurent, and Hanno on Habrok
- [x] Make sure all accounts are set up for Habrok
- [x] Update [`grid_proteus.py`](https://github.com/FormingWorlds/PROTEUS/blob/45b361fac78b1a85805eb006297c5fa82a196d1e/tools/grid_proteus.py#L4) or create a new script to be used with Slurm specifically (#382)
- [x] Create run scripts for submitting grids via this new script to Snellius and Habrok
- [x] Test the performance over larger grids (>1000 simulations)
- [x] Update the documentation to provide instructions for performing these simulation grids 

@HannoSpreeuw @lsoucasse Please can we have a discussion here how we can achieve this as efficiently as possible. The consistent issues we have with the local Kapteyn machine are holding the current project back.

lsoucasse:
I don't think it is necessary to provide me access to Habrok. I can access to Snellius and experiment there.

Both Habrok and Snellius uses SLURM as queuing manager.

HannoSpreeuw:
I also have a Snellius account.

nichollsh:
Creating a method for dispatching PROTEUS with SLURM would also be beneficial for me, as then I could run/test the code on the Oxford [ARC](https://www.arc.ox.ac.uk/overview) cluster. This is analogous to the Groningen Habrok cluster. I don't have access to this cluster, but I could request access to it in order to test the framework on a wider range of platforms.

timlichtenberg:
That sounds like a good idea. I am still working on getting Habrok accounts for all of you,

lsoucasse:
[...]

PR #382, opened 2025-04-14
Generate slurm config from grid_proteus

stefsmeets:
This PR adds a way to generate a slurm batch file for `grid_proteus.py`.

Submit using: `sbatch proteus_slurm_array.sh`

### Todo

- [x] Add docstring
- [x] Add back `run` method and refactor common parts out
- [x] Test it

Addresses #355

### DIFF
--- a/tools/grid_proteus.py
+++ b/tools/grid_proteus.py
@@ -4,13 +4,17 @@
 
 from __future__ import annotations
 
+import gc
 import itertools
 import logging
+import multiprocessing
 import os
 import shutil
+import subprocess
 import sys
 import time
 from copy import deepcopy
+from datetime import datetime
 from pathlib import Path
 
 import numpy as np
@@ -246,13 +250,7 @@ class Grid():
         log.info(" ")
 
 
-    def slurm_config(self, max_jobs:int=10, test_run:bool=False):
-        log.info("Running PROTEUS across parameter grid '%s'" % self.name)
-
-        log.info("Output path: '%s'" % self.outdir)
-        if self.using_symlink:
-            log.info("Symlink target: '%s'" % self.symlink_dir)
-
+    def write_config_files(self):
         # Read base config file
         base_config = read_config_object(self.conf)
 
@@ -264,8 +262,7 @@ class Grid():
             thisconf:Config = deepcopy(base_config)
 
             # Set case dir relative to PROTEUS/output/
-            path = f"{self.name}/case_{i:05d}"
-            thisconf.params.out.path = path
+            thisconf.params.out.path = self.name+"/case_%05d/"%i
 
             # Set other parameters
             for key in gp.keys():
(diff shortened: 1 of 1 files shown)


stefsmeets:
@lsoucasse @nichollsh do you want to test this? I don't have access to snellius or habrok, but let me know how the generated script works for you.

nichollsh:
Thanks for this Stef. Seems like a great way to handle the problem.
When I try your script, it generates the `proteus_slurm_array.sh` file as you described.

When I submit it, the tasks appear in `squeue`:
```
             JOBID PARTITION     NAME     USER ST       TIME  NODES NODELIST(REASON)
16587663_[10-431%1 regularsh proteus.  p321409 PD       0:00      1 (JobArrayTaskLimit)
        16587663_0 regularsh proteus.  p321409  R       0:00      1 node112
        16587663_1 regularsh proteus.  p321409  R       0:00      1 node112
        16587663_2 regularsh proteus.  p321409  R       0:00      1 node112
        16587663_3 regularsh proteus.  p321409  R       0:00      1 node112
        16587663_4 regularsh proteus.  p321409  R       0:00      1 node112
        16587663_5 regularsh proteus.  p321409  R       0:00      1 node112
        16587663_6 regularsh proteus.  p321409  R       0:00      1 node111
        16587663_7 regularsh proteus.  p321409  R       0:00      1 node111
        16587663_8 regularsh proteus.  p321409  R       0:00      1 node111
        16587663_9 regularsh proteus.  p321409  R       0:00      1 node111
```
Does this mean we are limited to 10 tasks at a time?

Each task within the array also seems to give the following error:
`/var/spool/slurmd/job16587349/slurm_script: line 445: --offline: command not found`

The stdout file shows:
```
executing proteus
Usage: proteus [OPTIONS] COMMAND [ARGS]...

Options:
  --version  Show the version and exit.
  --help     Show this message and exit.

Commands:
  doctor   Diagnose your PROTEUS installation
  get      Get data and modules
  observe  Run synthetic observations pipeline
  offchem  Run offline chemistry on PROTEUS output files
  plot     (Re-)generate plots from completed run
  start    Start proteus run
executing --offline
```

stefsmeets:
In `grid_proteus.py` you can set `max_jobs`. I arbitrarily put 10 there, but you can put whatever you want. Indeed that means it is limited to 10 at a time, but you can tune your code to the number of nodes you have available.

`/var/spool/slurmd/job16587349/slurm_script: line 445: --offline: command not found`

Seems like it is not executing the commands correctly and instead considering every argument as a command, I will have a look.

stefsmeets:
Not sure tbh, the script looks fine to me. I don't have access to slurm, so I cannot debug it.

nichollsh:
This seems to work! Thanks @stefsmeets.

I had to make a small change:
* Set `self.tmpdir` to use a directory inside the grid's output folder. This is because the `/tmp` directory it currently writes to is on a different machine to the one that SLURM executes the simulation on.

nichollsh:
Two suggestions:
1. We could have grid_proteus submit the job automatically, rather than having the user do it. 
2. We should update the information printed to the display. "Do not close this program! It must stay alive to manage each of the subproceses." is not true anymore. Similarly, it might be handy to print information on how to cancel the jobs.

stefsmeets:
> 1. We could have grid_proteus submit the job automatically, rather than having the user do it.

I'd advice against this. Submitting is a single command and imo it is better to learn how to deal with slurm than to try and automate it.

> 2\. The `proteus_slurm_array.sh` script should probably be generated in the output folder, since it's grid specific.

Makes sense, I will update the code.

> 3\. We should update the information printed to the display. It might be handy to print information on how to cancel the jobs.

Agree that it could be helpful. I can add scancel / squeue, any other suggestions here?

nichollsh:
> Is that what the `os.sync` and odd sleep timings are for?

Yes. I agree that it's probably system specific, so there's no neat solution, but I had issues along these lines before which were resolved by forcing buffer synchronisation and a sleep period.

nichollsh:
> I can add scancel / squeue, any other suggestions here?

Both of these sound good.

stefsmeets:
> > Is that what the `os.sync` and odd sleep timings are for?
> 
> Yes. I agree that it's probably system specific, so there's no neat solution, but I had issues along these lines before (on a single machine, not on a cluster) which were resolved by forcing buffer synchronisation and a sleep period. I can imagine that the current formulation isn't sufficient when the files are synchronised across multiple machines, rather than just from the buffer to the disk on a single machine.
> 
> I think the duration for this sleep period probably varies a lot. We could play it safe and include a 30 seconds wait period? This would also give the user time to cancel if a mistake was made.

Yeah, not sure either. Adding a sleep is also annoying, especially if you don't need it. If you can detect which system you are running on, you could do something like:

```python
if HABROK:
    log.info('sleeping for 30 to allow file synchroniozation')
    time.sleep(30)
```

For a general solution, I think this should be done by the code that writes the config files, i.e. that it makes sure the files exist. But the above would work also. Let's tackle that in a follow-up PR, since I don't have access to any of these systems.

stefsmeets:
This PR is ready, @nichollsh if you are happy with it, I think it can be merged.

nichollsh:
Makes sense. Thanks a lot for this. I am happy for it to be merged.

nichollsh (approved):

timlichtenberg:
Awesome, thanks a lot @stefsmeets for incorporating these addtions! :-)

PR #385, opened 2025-04-17
Archive data files from simulations

### LINKED ISSUE #384: Limit number of files in large grids
## The problem

Running PROTEUS creates a lot of files. We save 1D profiles for the atmosphere and interior, for each point in time. There are also plots and other misc files.

For example: a simulation for 1 Gyr of evolution created 708 files in its output folder. If we have a grid of 800 points this would be >566k files. 

The hard file limit on Habrok is 537600. **I was able to exceed this limit with a single grid**. There's advice regarding file/space limits on the [Habrok wiki](https://wiki.hpc.rug.nl/habrok/data_management/storage_areas). They suggest compressing files into archives, and also to use local disk storage for short term usage. We could consider both of these, but the former seems easier. 

If we want to fully utilise the number of CPUs available, we need to get around this file limitation. It will probably exist on other servers too.

## Suggestion

Add an option in the config file to compress the PROTEUS output into archives. For a given output folder `sim/`, we could compress `sim/data/` into `sim/data.tar`. In the above example this would reduce the number of files from 708 to 32. 

This step could be performed once the simulation has completed, or done on-the-fly where all but the most recent files are placed into a running archive. This might require disabling some of the plotting routines (e.g. cpl_atmosphere), as they need access to the files, and only running these at the end.

We could also save disk space by removing the spectral files (`sim/runtime.sf`, `sim/runtime.sf_k`) when each simulation terminates. These tend to be quite large files.

An extreme option is to simply not save the full data from the model, limiting ourselves to only the most recent atmosphere and interior profiles, plus the helpfile.

nichollsh:
Closes #384

### DIFF
--- a/input/demos/dummy.toml
+++ b/input/demos/dummy.toml
@@ -8,6 +8,7 @@ author  = "Harrison Nicholls, Tim Lichtenberg"
         logging     = "ERROR"
         plot_mod    = 0      # Plotting frequency, 0: wait until completion | n: every n iterations
         write_mod   = 50      # Write CSV frequency, 0: wait until completion | n: every n iterations
+        archive_mod = 'none'
 
     [params.dt]
         minimum      = 1e2    # yr, minimum time-step
--- a/src/proteus/config/_params.py
+++ b/src/proteus/config/_params.py
@@ -6,11 +6,21 @@ from __future__ import annotations
 from attrs import define, field
 from attrs.validators import ge, gt, in_, lt
 
+from ._converters import none_if_none
+
+def valid_path(instance, attribute, value):
+    if not isinstance(value, str) or not value.strip():
+        raise ValueError(f"'{attribute.name}' must be a non-empty string")
 
 def max_bigger_than_min(instance, attribute, value):
     if value <= instance.minimum:
         raise ValueError("'maximum' has to be bigger than 'minimum'.")
 
+def valid_archive_mod(instance, attribute, value):
+    if value is None:
+        return
+    if value < 0:
+        raise ValueError("'archive_mod' must be None or greater than 0.")
 
 @define
 class OutputParams:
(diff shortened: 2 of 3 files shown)


planetmariana (approved):
Thanks @nichollsh, this is amazing! Reducing the number of files two order of magnitude :D , maybe we can add some clarification in the docs saying that if the user would like to extract x simulation, would have to download and un-tar the file :)

docs/usage.md
```diff
@@ -2,28 +2,55 @@
...
+
+Simulations with PROTEUS create several types of output files. For the `all_options` example,
+the results are located at `output/all_options/`. The tree below outlines the purposes of
+some files and subfolders contained within the simulation's output folder.
+
+```
+all_options/
+ ├─runtime_helpfile.csv         <---- table containing the main simulation results
+ ├─proteus_00.log               <---- the log file from the simulation
+ ├─init_coupler.toml            <---- a completed copy of the configuration file
```
timlichtenberg, line 38: Not a priority in this PR, but this file should be renamed, as "init_coupler" is legacy file naming convention by now.

timlichtenberg (commented):
Looks very good overall! 

Something I couldn't immediately see: is the `/plots`subfolder also archived within the simulation directory or is kept separate? If the former, I would suggest to keep it out of the archive, as the number of files within that folder are only a few, but it enables a quick visual inspection by eye without having to go through the unzip process.

docs/usage.md
nichollsh: Agreed. Maybe something like `config.toml`?

timlichtenberg: `config.toml` seems good to me.
For replacement name for `runtime_helpfule.csv`, should we highlight that these variables are mostly global? For example: `output_global.csv`?

nichollsh:
> is the `/plots`subfolder also archived within the simulation directory or is kept separate?

The plots stored inside `plots/` are kept as loose files at the moment. I wrote the archiving functionality to be generalisable to any subfolder, so we could archive them if we wanted to in the future. Same for synthetic observations and VULCAN output.

However, I decided to keep them loose for the same reason that you specified.

timlichtenberg:
Great!

nichollsh:
> maybe we can add some clarification in the docs saying that if the user would like to extract x simulation, would have to download and un-tar the file :)

There's a short section here: https://fwl-proteus--385.org.readthedocs.build/en/385/usage/#archiving-output-files

Do you think I should expand it or move it?

timlichtenberg (approved):

planetmariana:
> 
> There's a short section here: https://fwl-proteus--385.org.readthedocs.build/en/385/usage/#archiving-output-files @planetmariana, do you think I should expand it or move it?

Oh, I didn't saw that section. I think it's perfect then ! Thanks @nichollsh

PR #393, opened 2025-04-22
Limit platon temperature range, and a mitigation for data-download failures

### LINKED ISSUE #206: Robust file dependency download process

### LINKED ISSUE #392: Platon error: temperature out of range
Platon's opacity lookup table is valid for temperatures 100 < T < 1000 K, which is a reasonable range for the upper atmosphere. The model throws an error if providing temperatures outside this range, which can sometimes occur on the top-of-atmosphere temperature node with AGNI.

We can resolve this by clipping the temperature profile provided to Platon to this range.

(Might be worth considering pRT as an alternative to Platon, as I think @timlichtenberg suggested previously. Installing pRT on Habrok breaks my netCDF installation so we should look into this another time.)

nichollsh:
* Limit platon temperature profile to valid range. Closes #392 
* Change helpfile.csv precision to %.10e from %.8e, to avoid roundoff errors in Time column
* Write helpfile on final iteration before synthetic observations, so that they use the correct data.
* Re-attempt download from OSF up to three times. Closes #206.

Resolves issue #395 by manually adding libssl to the Ubuntu deps, and forcing AGNI to use OpenSSL v3.0 rather than v3.5 (changes made in the AGNI repo).

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -21,10 +21,10 @@ jobs:
     strategy:
       matrix:
         os: ['ubuntu-latest', 'macos-latest']
-        python-version: ['3.11', '3.12']
+        python-version: ['3.11', '3.13']
         include:
           - os: ubuntu-latest
-            INSTALL_DEPS: sudo apt-get update; sudo apt-get install libnetcdff-dev netcdf-bin
+            INSTALL_DEPS: sudo apt-get update; sudo apt-get install libnetcdff-dev netcdf-bin libssl-dev
             CC: gcc
             CXX: g++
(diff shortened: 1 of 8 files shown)


EmmaPostolec (approved):
Looks good to me Harrison :)

PR #401, opened 2025-05-01
Integrating Zalmoxis 

flaviacris03:
This PR integrates the Zalmoxis interior modeling module into PROTEUS. The main additions include:

1. New core functions for Zalmoxis:

- `get_density_from_eos()` calculates density for given pressure by reading in from EOS files
- `interior_structure_odes()` calculates the derivatives of mass, gravity, and pressure wrt radius
- `zalmoxis_solver()` solves the interior structure of the planet

2. Integration in `solve_structure()`:

- If case = self, the interior structure is solved with previously existing function `determine_interior_radius()`
- If case = zalmoxis, the interior structure is solved with the new `determine_interior_radius_with_zalmoxis()` function. Within `determine_interior_radius_with_zalmoxis()`, the Zalmoxis solver (`zalmoxis_solver()`) is first called to compute the planet’s interior structure when the total mass (`mass_tot`) and a core mass fraction (`coremassfrac`) are specified. Subsequently, `run_interior()` is called to further evolve or update the planet's interior state based on the Zalmoxis output.

4. Reading in the Seager EOS
`get_Seager_EOS()` (in utils.data) builds the material properties dictionary for the Seager et al. (2007) EOS data used by Zalmoxis. It reads EOS files from FWL_DATA/EOS_material_properties/EOS_Seager2007

Usage:
- Zalmoxis can be enabled by setting module = "zalmoxis" under the [struct] configuration block in the configuration file.
- Parameters to configure Zalmoxis include:
```
coremassfrac = 0.325           # non-dimensional core mass fraction
inner_mantle_mass_fraction = 0 # non-dimensional inner mantle mass fraction (for more than 2 interior layers)
weight_iron_frac = 0.325       # non-dimensional iron fraction in the planet
num_levels = 100               # number of Zalmoxis radial layers
EOSchoice = "Tabulated:iron/silicate"  # EOS choice for Zalmoxis
```


lsoucasse:
Hi @flaviacris03, I turned your PR as ready for review otherwise it does not run the tests. Here there are some problem. You need to first make the tests pass on your machine running `pytest` at the root of the proteus directory. For fixing the code style (formatting), you need to run `ruff check` (again at the root of the repository) and it will tell you what it does not like in your changes.

lsoucasse (changes requested):
Thanks for the work @flaviacris03.
This needs to be fixed for the test to pass. Also I am wondering if Zalmoxis should be included in the physical test (or in a new test), what do you think @timlichtenberg ? Maybe in a following PR?
Also, I see you updated the `minimal.toml` but you also have to modify the `all_options.toml` such that all the options you created for Zalmoxis are listed, even if the input file does not make use of it, just for exposing completeness of all possible options to the user.

timlichtenberg:
@lsoucasse I agree there should be a Zalmoxis-related test. For now I suggested to Flavia that we first want to build Zalmoxis-internal tests and get the general coupling to work without errors. Once that is the case we can think of integrating a test the speaks to the numerical integration of Zalmoxis.

flaviacris03:
I will make the changes you suggested @timlichtenberg  and set up a new PR once the gravity branch is merged into Aragog so I can include those changes into Zalmoxis as well.

timlichtenberg:
Sounds good, thanks.

PR #402, opened 2025-05-06
Update VULCAN wrapper with new Config object

nichollsh:


lsoucasse (approved):
Looks good to me, though I never use Vulcan so far.

PR #403, opened 2025-05-06
Fix 353

### LINKED ISSUE #353: Use of global makes debugging harder
I very rarely run into the use of `global` in Python code.

This is an example, in [aragog.py](https://github.com/FormingWorlds/PROTEUS/blob/f9cb8c6ecc494a6b5b027627ba6599efaab0ecfc/src/proteus/interior/aragog.py#L44C5-L44C25):
```
    global aragog_solver
```

I am not sure we really need `global` here.
If not, I would like to replace it by a function argument.

`global` objects are hard to debug, since they introduce hidden states.

HannoSpreeuw:
Fixes #353 
i.e. `aragog_solver` is no longer a `global`.
but the [Aragog computational efficiency problem](https://github.com/ExPlanetology/aragog/issues/47) remains.

This fix:

1. Introduction of `AragogRunner` class.
2. `aragog_solver` is an attribute of `interior_o` i.e. an attibute of an instance of `Interior_t`.
3. `AragogRunner.__init__` has a line `interior_o.aragog_solver.initialize()`.  This will trigger a [Solver().initialize in Aragog](https://github.com/ExPlanetology/aragog/blob/b86e217838c128b5709e9948990acc0ba949c2b9/aragog/solver.py#L436). Most likely there is a lot there, [like reading in files](https://github.com/ExPlanetology/aragog/blob/b86e217838c128b5709e9948990acc0ba949c2b9/aragog/phase.py#L202), that is, in fact, solver independent.
5. Last line of `AragogRunner.__init__`: `self.aragog_solver = interior_o.aragog_solver`

### DIFF
--- a/tests/integration/test_integration_physical.py
+++ b/tests/integration/test_integration_physical.py
@@ -33,6 +33,7 @@ IMAGE_LIST = (
         "plot_population_mass_radius.png"
         )
 
+
 @pytest.fixture(scope="module")
 def physical_run():
     runner = Proteus(config_path=config_path)
--- a/src/proteus/interior/aragog.py
+++ b/src/proteus/interior/aragog.py
@@ -10,6 +10,7 @@ from typing import TYPE_CHECKING
 import netCDF4 as nc
 import numpy as np
 import pandas as pd
+import hashlib
 import platformdirs
 from aragog import Output, Solver, aragog_file_logger
 from aragog.parser import (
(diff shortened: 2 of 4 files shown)


tests/integration/test_integration_physical.py
lsoucasse, line 1: Is this additional line really necessary?

HannoSpreeuw: Sorry, which additional line?
Ah, you mean `dirs)` on the next line?
Yes, otherwise it would exceed 80 characters.

nichollsh:
> It seems to be related to agni, maybe @nichollsh can help?

It's probably related to the VULCAN update that was just merged. Because VULCAN is not a package, we cannot specify the version that PROTEUS should pull. This means that PROTEUS uses the latest version, which is now only working again with my latest PR merged.

@HannoSpreeuw, if you merge `main` into this branch then that should fix the problem with the tests.

HannoSpreeuw:
Thanks @nichollsh 

All right, that would be equivalent to rebasing against a fresh main, right?

nichollsh:
Possibly. I am not an expert on git, so it is not clear to me if rebasing is the right word.
I would do:
```bash
git checkout main
git pull
git checkout Fix_353
git merge main
git push
```

nichollsh (approved):
Looks good to me. Thanks for fixing this, Hanno

lsoucasse (approved):

PR #406, opened 2025-05-07
New plot to visualise planet evolution

### LINKED ISSUE #404: Atmosphere visualiser tool/plot
It would be useful to include an artistic visualiser within PROTEUS, which shows how a given a simulated planet might look 'to the eye' at a given point in time.

These visualisations would be powerful in research proposals as eye catching figures aimed at reviewers. They could also be used for outreach, advertising PROTEUS, and in the documentation.

nichollsh:
This PR introduces a new plot, which is generated automatically at the end of each simulation. See example below. The planet's atmosphere is plotted as a series of concentric circles, with the colour of each circle set by the radiation fluxes calculated during the simulation. These fluxes are mapped to colours using [CIE colour matching functions](https://en.wikipedia.org/wiki/CIE_1931_color_space?useskin=vector) with some tweaks. Closes #404.

It also includes the star in the background, scaled based on its distance and relative size. The inset shows the emission spectrum calculated by SOCRATES.

This is useful for visualising how the planet cools and contracts over time. These plots could be used in presentations, proposals, etc.

I have externally made a notebook which generates frames of these plots and animates them, but did not include it in this PR

https://github.com/user-attachments/assets/8c56e19b-396f-4a2d-b693-abfe75efa774

### DIFF
--- a/src/proteus/atmos_clim/common.py
+++ b/src/proteus/atmos_clim/common.py
@@ -129,8 +129,12 @@ def read_atmosphere_data(output_dir:str, times:list, extra_keys=[]):
     """
     profiles = [read_ncdf_profile(os.path.join(output_dir, "data", "%.0f_atm.nc"%t),
                                     extra_keys=extra_keys) for t in times]
-    if len(profiles) < 1:
-        log.warning("No atmosphere NetCDF files found in output folder")
+    if None in profiles:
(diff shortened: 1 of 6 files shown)


EmmaPostolec (approved):
This is a really cool plot to have a better idea of the evolution of the planet in time, I really like it !

PR #410, opened 2025-05-27
Ls/aragog_reset

lsoucasse:
This PR makes use of a new Aragog functionality that is the ability to reset the solver while keeping in memory the previous state of the phase object. This avoids loading multiple times the lookup data and makes the execution of Proteus much faster.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -52,8 +52,10 @@ jobs:
       - name: Symlink gfortran
         if: runner.os == 'macOS'
         run: |
-          sudo ln -s /opt/homebrew/bin/gfortran-13 /opt/homebrew/bin/gfortran
-          sudo ln -s /opt/homebrew/Cellar/gcc/12.*/lib/gcc/12/*.dylib /opt/homebrew/lib/
+          if [ ! -L /opt/homebrew/bin/gfortran ]; then
+            sudo ln -s /opt/homebrew/bin/gfortran-13 /opt/homebrew/bin/gfortran
+          fi
+          sudo ln -s /opt/homebrew/Cellar/gcc/12.*/lib/gcc/12/*.dylib /opt/homebrew/lib/ || true
           which gfortran
 
       # Setup Julia


src/proteus/interior/aragog.py
```diff
@@ -76,15 +75,14 @@ def setup_or_update_solver(config: Config, hf_row: dict,
         else:
             if interior_o.ic == 1:
                 AragogRunner.update_structure(config, hf_row, interior_o)
-                interior_o.aragog_solver.initialize()
+                interior_o.aragog_solver.reset()
```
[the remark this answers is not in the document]
lsoucasse: True! Good catch, I made several attempts beforehead but now the reset method is always called.

src/proteus/interior/aragog.py
```diff
@@ -73,8 +72,10 @@ def setup_or_update_solver(config: Config, hf_row: dict,
             if config.params.resume:
                 AragogRunner.update_solver(dt, hf_row, interior_o,
                                    output_dir=dirs["output"])
+            interior_o.aragog_solver.initialize()
```
HannoSpreeuw, line 74: Now we have
```
interior_o.aragog_solver.initialize()
```
after 
```
AragogRunner.setup_solver(config, hf_row, interior_o, dirs["output"])
```
or even after
```
AragogRunner.update_solver(dt, hf_row, interior_o, output_dir=dirs["output"])
```

So that means that the initialize method (of the Solver class in Aragog) should perhaps be renamed, since initialization is always a first step, at least that is the association I have.

src/proteus/interior/aragog.py
```diff
@@ -293,6 +299,47 @@ def update_solver(dt:float, hf_row:dict, interior_o:Interior_t,
             interior_o.tides /
             interior_o.aragog_solver.parameters.scalings.power_per_mass)
 
+    @staticmethod
+    def update_structure(config: Config, hf_row:dict, interior_o:Interior_t):
+
+        print("update structure")
+        interior_o.aragog_solver.parameters.mesh.outer_radius = hf_row["R_int"] / interior_o.aragog_solver.parameters.scalings.radius
```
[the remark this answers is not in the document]
lsoucasse: OK, but if you really want to enforce this you should update the code style requirements?

src/proteus/interior/aragog.py
lsoucasse: `initialize` is only called after the setup and `reset` is called after `update_solver` or `update_structure`.

src/proteus/interior/aragog.py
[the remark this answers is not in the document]
lsoucasse: Stef set this up. It uses ruff and there are some settings in the pre-commit yaml file at the root of the repo.

src/proteus/interior/aragog.py
HannoSpreeuw: All right, I can turn this into an Aragog issue - if you agree.
The Solver class obviously has an [\_\_init\_\_ method](https://github.com/ExPlanetology/aragog/blob/f601a8f89792088ee2abe67080125f5e2b5c083d/aragog/solver.py#L412)
and an [initialize method](https://github.com/ExPlanetology/aragog/blob/f601a8f89792088ee2abe67080125f5e2b5c083d/aragog/solver.py#L436).

The name of the initialize method should be turned into something more meaningful, currently I find it confusing.

HannoSpreeuw:
```
            if config.params.resume:
                AragogRunner.update_solver(dt, hf_row, interior_o,
                                   output_dir=dirs["output"])
            interior_o.aragog_solver.initialize()
```
So if `config.params.resume==True` then the initialize method is applied after updating the solver, which seems awkward.

HannoSpreeuw: I mean, I can find out what initialize does, by tracking the calls, but think of a someone new who tries to understand the logical order of the subsequent steps; it may not be so clear why you first have to setup a solver and then initialize it, unless you rename the initialize and/or setup methods to something more informative.

src/proteus/interior/aragog.py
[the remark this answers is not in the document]
HannoSpreeuw: Got [it](https://github.com/FormingWorlds/PROTEUS/blob/main/.pre-commit-config.yaml).

Launched from [here](https://github.com/FormingWorlds/PROTEUS/blob/main/.github/workflows/code-style.yaml)?

timlichtenberg:
This seems to have fixed it for now...?

lsoucasse:
The tests now pass after removing the cache and reverting the lookup data.
Also tests in the main branch are passing, see [here](https://github.com/FormingWorlds/PROTEUS/actions/runs/15389332177).
I think we can merge this now.
@planetmariana we now need to figure out what changes with your new data and updates the test data if needed.

timlichtenberg (approved):
Seems good to me.

timlichtenberg:
Or @lsoucasse is it better to do this locally first?

PR #411, opened 2025-05-27
Validate module versions

### LINKED ISSUE #398: Check submodule version compatability
## Problem 
When updating PROTEUS with `git pull`, the code will not check that the installed versions of submodules are compatible with the updates to PROTEUS.

For example, recent changes to PROTEUS require CALLIOPE and AGNI to be updated to their latest versions. A user would not be warned of this, as pointed out by @EmmaPostolec.

## Solutions
1. We could consider adding a command like `proteus update` which pulls updates for PROTEUS and its submodules. This would enhance the usability of the code, but would be quite complicated and potentially annoying if the user wants to use a specific version of each submodule.
2. Instead, we could simply add a check to be performed when the code is started. This would only check that the submodules are compatible and warn the user, requesting that they update them manually.

Option (2) seems like the better one. Open to other suggestions and thoughts.

nichollsh:
Ensures that a user is using up-to-date versions of modules. Raises an error if one of the modules is out of date. Adds documentation explaining what to do if this is the case. Closes #398.

Also tidied up docs page: Troubleshooting.md

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -14,6 +14,31 @@ Have you added your SSH key to GitHub? See these pages for guidance:
 * [Adding a new SSH key to your Github account](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/adding-a-new-ssh-key-to-your-github-account)
 * [Testing your SSH connection](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/testing-your-ssh-connection)
 
+## All: Out-of-date modules detected
+
+You will have an error that looks like this:
+```console
+[ INFO  ] Validating module versions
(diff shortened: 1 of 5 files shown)


lsoucasse (approved):
Looks good Harrison. So you do have version numbers in Agni now? Also, is it only checking the modules you actually need for the run?

nichollsh:
Thanks, Laurent.

Yes - it only checks the modules needed for the loaded configuration. So if you're not using JANUS, it won't check that JANUS is up to date (or even installed at all).

AGNI uses semantic versioning. The versions used to be prefaced with 'v' but I dropped this last year. You can see the versions here: https://github.com/nichollsh/AGNI/releases

PR #435, opened 2025-06-23
Ls/zenodo

lsoucasse:
This PR adds Zenodo as another remote host for forming world data. The download function has been modified such that it tries to download the data from Zenodo first and then from the OSF if it fails.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -58,6 +58,7 @@ dependencies = [
     "pandas",
     "scipy",
     "sympy",
+    "zenodo-get",
     "astropy"
 ]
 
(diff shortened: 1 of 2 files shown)


lsoucasse:
> So shall we review this here first now, and then subsequently you will add the PRs for the other modules, or shall we wait with the reviews for the other modules first?

I would like to merge it now, what is happening in the other modules is independent of Proteus and will not change the PROTEUS code.

nichollsh:
@lsoucasse what's the plan with mapping the other spectral files' Zenodo record numbers?

lsoucasse:
> @lsoucasse what's the plan with mapping the other spectral files' Zenodo record numbers?

I want to create another issue to list the data we have to upload to Zenodo. I can do some of it but I would like some else to add short description of the records to briefly state where the data come from (which paper) to credit authorship. I am a bit uncomfortable being the "creator" of all these Zenodo records.

PR #456, opened 2025-07-19
Fixes 440 partially and 455

### LINKED ISSUE #440: Make PROTEUS as a whole pip-installable
- Whole code needs to work with just `pip install fwl-proteus`
- This includes that Aragog becomes the default interior module
- SPIDER does not be part of the whole pip-installable package
- Combination should Aragog-AGNI-CALLIOPE-MORS
- If not doable as a whole, but in 2-3 steps, this is also ok
- Add to documentation to make this installation route crystal clear for the user

### LINKED ISSUE #455: Remove backticks from line 39 in get_socrates.sh
```
echo "Now add the following to your `~/.bashrc` file:"
````
from [line 39 in `tools/get_socrates.sh`](https://github.com/FormingWorlds/PROTEUS/blob/fc354289e026b735afdab413e0b8419224d957f3/tools/get_socrates.sh#L39)  yields
```
-bash: /home4/username/.bashrc: Permission denied
```
on Habrok.

**To Reproduce**
Log into Habrok.
Run from bash:
```
echo "Now add the following to your `~/.bashrc` file:"
```

Easy fix, change to
```
echo "Now add the following to your ~/.bashrc file:"
```
i.e. remove backticks.

### PR DESCRIPTION
**Pull request by HannoSpreeuw**
Fixes #440 partially, i.e. more than 2 steps are required:

    git clone https://github.com/FormingWorlds/PROTEUS.git
    cd PROTEUS
    git switch Fix_440
    conda env create -f environment.yml
    conda activate proteus
    pip install -e .
    proteus install-all --export-env

It also requires the git and (mini)conda executables.

I do think it is an improvement in terms of convenience and robust, although it always install Julia in a default folder.
Also,
`coverage run -m pytest`
yields
```
E           OSError: Out-of-date modules detected. Refer to the Troubleshooting guide on the wiki:
E           https://fwl-proteus.readthedocs.io/en/latest/troubleshooting/
```
So I need to fix that as well as a number of warnings.

Additionally, SOCRATES install errors are reported:
```
🌤️ Installing SOCRATES...
git@github.com: Permission denied (publickey).
Cloning from GitHub
    https://github.com/nichollsh/SOCRATES.git -> socrates
Cloning into 'socrates'...
remote: Enumerating objects: 2599, done.
remote: Counting objects: 100% (166/166), done.
remote: Compressing objects: 100% (134/134), done.
remote: Total 2599 (delta 58), reused 64 (delta 29), pack-reused 2433 (from 1)
Receiving objects: 100% (2599/2599), 63.54 MiB | 32.32 MiB/s, done.
Resolving deltas: 100% (1203/1203), done.
Updating files: 100% (1562/1562), done.
Building the radiation code in bin/ and utilities in sbin/
Sat Jul 19 11:00:32 CEST 2025
cc -o Cgen_null Cgen_null.c -lm
Cgen_null.c: In function 'main':
Cgen_null.c:14:5: warning: old-style function definition [-Wold-style-definition]
   14 | int main(argc, argv)
      |     ^~~~
Cgen_null.c:52:13: error: too many arguments to function 'fill'; expected 0, have 6
   52 |             fill(&np, p, p1, p2, division, I_LINEAR);
      |             ^~~~ ~~~
Cgen_null.c:34:9: note: declared here
   34 |    void fill();
...
...
```
which, however, do not seem fatal.
@nichollsh Any clue wrt to fixing this?
`git@github.com: Permission denied (publickey)` is a nuisance.

Fixes #455

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,0 +1,14 @@
+name: proteus
+channels:
+  - conda-forge
+  - defaults
+dependencies:
+  # Core Python
+  - python=3.13
+  - pip
+
+  # Non-Python dependencies (system-level)
+  - make
+  - gfortran
+  - netcdf-fortran
+  - git  # required for AGNI cloning if Git is not present
--- a/environment.yml
+++ b/environment.yml
@@ -1,7 +1,6 @@
 name: proteus
 channels:
   - conda-forge
-  - defaults
 dependencies:
   # Core Python
   - python=3.13
--- a/environment.yml
+++ b/environment.yml
@@ -5,6 +5,7 @@ dependencies:
   # Core Python
   - python=3.13
   - pip
+  - pytest
 
   # Non-Python dependencies (system-level)
   - make
--- a/src/proteus/cli.py
+++ b/src/proteus/cli.py
@@ -322,7 +322,11 @@ def install_julia_and_get_bin_path() -> Path | None:
 
     try:
         proc = subprocess.run(
-            ["bash", "-c", "curl -fsSL https://install.julialang.org | sh"],
+            [
+                "bash",
+                "-c",
+                "curl -fsSL https://install.julialang.org | sh  -s -- -y",
+            ],
             check=True,
             capture_output=True,
             text=True,
@@ -335,12 +339,15 @@ def install_julia_and_get_bin_path() -> Path | None:
         )
         if match:
             julia_bin = Path(match.group(1))
+            click.secho(
+                f"📦 Detected Julia install path: {julia_bin}", fg="cyan"
+            )
             return julia_bin
         else:
             click.secho(
                 "⚠️  Could not determine Julia install path from output."
             )
-            click.secho(proc.stdout)
+            click.secho(proc.stdout, fg="white")
             return None
     except subprocess.CalledProcessError as e:
         click.secho(
--- a/src/proteus/cli.py
+++ b/src/proteus/cli.py
@@ -356,6 +356,9 @@ def install_all(export_env: bool):
     rad_dir = socrates_dir.resolve()
     os.environ["RAD_DIR"] = str(rad_dir)
 
+    # Pick up the current shell PATH (which includes the Julia path)
+    env = os.environ.copy()
+
     # --- Step 3: Install AGNI ---
     agni_dir = root / "AGNI"
     if not agni_dir.exists():
@@ -364,7 +367,9 @@ def install_all(export_env: bool):
             subprocess.run(
                 ["git", "clone", "git@github.com:nichollsh/AGNI.git"], check=True
             )
-            subprocess.run(["bash", "src/get_agni.sh"], cwd=agni_dir, check=True)
+            subprocess.run(
+                ["bash", "src/get_agni.sh"], cwd=agni_dir, env=env, check=True
+            )
         except subprocess.CalledProcessError as e:
             click.secho("❌ Failed to install AGNI", fg="red")
             click.echo(e)
--- a/src/proteus/cli.py
+++ b/src/proteus/cli.py
@@ -379,7 +379,9 @@ def install_all(export_env: bool):
                 ["git", "clone", "https://github.com/nichollsh/AGNI.git"],
                 check=True,
             )
-            subprocess.run(["bash", 'echo "PATH seen by get_agni.sh: $PATH"'])
+            subprocess.run(
+                ["bash", "-c", 'echo "PATH seen by get_agni.sh: $PATH"']
+            )
             subprocess.run(
                 ["bash", "src/get_agni.sh"], cwd=agni_dir, env=env, check=True
             )
--- a/src/proteus/cli.py
+++ b/src/proteus/cli.py
@@ -379,6 +379,7 @@ def install_all(export_env: bool):
                 ["git", "clone", "https://github.com/nichollsh/AGNI.git"],
                 check=True,
             )
+            subprocess.run(["bash", 'echo "PATH seen by get_agni.sh: $PATH"'])
             subprocess.run(
                 ["bash", "src/get_agni.sh"], cwd=agni_dir, env=env, check=True
             )
--- a/src/proteus/cli.py
+++ b/src/proteus/cli.py
@@ -376,7 +376,7 @@ def install_all(export_env: bool):
         click.secho("🧪 Installing AGNI...", fg="blue")
         try:
             subprocess.run(
-                ["git", "clone", "git@github.com:nichollsh/AGNI.git"],
+                ["git", "clone", "https://github.com/nichollsh/AGNI.git"],
                 check=True,
             )
             subprocess.run(
--- a/src/proteus/cli.py
+++ b/src/proteus/cli.py
@@ -269,7 +269,8 @@ def observe(config_path: Path):
 cli.add_command(offchem)
 cli.add_command(observe)
 
-DEFAULT_DATA_DIR = Path.home() / ".proteus" / "FWL_DATA"
+DEFAULT_DATA_DIR = Path(__file__).resolve().parents[3] / "FWL_DATA"
+print(f"[DEBUG] DEFAULT_DATA_DIR resolves to: {DEFAULT_DATA_DIR}")
 
 
 def resolve_fwl_data_dir() -> Path:
--- a/src/proteus/cli.py
+++ b/src/proteus/cli.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import os
 import platform
+import re
 import shutil
 import subprocess
 import sys
@@ -307,7 +308,7 @@ def is_julia_installed() -> bool:
     return shutil.which("julia") is not None
 
 
-def install_julia():
+def install_julia_and_get_bin_path() -> Path | None:
     click.secho("🐹 Julia is not installed. Installing Julia...", fg="blue")
     if platform.system() == "Windows":
         click.secho(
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by nichollsh** [comment]
Thanks, @HannoSpreeuw.

The error "Out-of-date modules detected..." will probably be resolved when you merge `main` back into this branch.

I think these compilation errors are from some auxiliary SOCRATES tools which are written in C rather than F90. Probably not an issue but maybe worth looking into.

The warning about publickey authentication can be easily suppressed. Doesn't matter for non-developers.

I do not think that Julia should be installed by this script. Just as we do not include installing gfortran, python, or any other 'system level' packages.

**Comment by HannoSpreeuw** [comment]
> 4\. I do not think that Julia should be installed by this script. Just as we do not include installing gfortran, python, or any other 'system level' packages.

Okay. We can install Julia as a non-system-level package:
```
conda install -c conda-forge julia
```
Btw, `proteus install-all` always checks if Julia is already installed:
```
def is_julia_installed() -> bool:
    return shutil.which("julia") is not None
```

**Comment by nichollsh** [comment]
I mean that the script should not install Julia at all, just as it would not install Python from-scratch either.

**Comment by nichollsh** [comment]
@HannoSpreeuw I also suggest making Python 3.12 the preferred version in `environment.yml`, since SPIDER will not install under Python 3.13. Obviously we are moving away from SPIDER anyway, but it would be good to make it easy for now.

**Comment by HannoSpreeuw** [comment]
All right, I have addressed the most important issues. It seems to work apart from some Deprecation Warnings from "pip install -e .":
```
Building wheels for collected packages: fwl-proteus, f90nml, wget
  Building editable for fwl-proteus (pyproject.toml) ... done
  Created wheel for fwl-proteus: filename=fwl_proteus-25.5.20-0.editable-py3-none-any.whl size=8929 sha256=d54a87ab043123d63de4e261bab21fa141ba14d5df4158698099dca9d3253d1d
  Stored in directory: /tmp/pip-ephem-wheel-cache-lee040gh/wheels/6d/cd/54/656735642e28f22681237da0dc9fd788d1fc18c19bef28563a
  DEPRECATION: Building 'f90nml' using the legacy setup.py bdist_wheel mechanism, which will be removed in a future version. pip 25.3 will enforce this behaviour change. A possible replacement is to use the standardized build interface by setting the `--use-pep517` option, (possibly combined with `--no-build-isolation`), or adding a `pyproject.toml` file to the source tree of 'f90nml'. Discussion can be found at https://github.com/pypa/pip/issues/6334
  Building wheel for f90nml (setup.py) ... done
  Created wheel for f90nml: filename=f90nml-1.4.5-py2.py3-none-any.whl size=33196 sha256=9cebeedfb922ad84a350cf43765153873dc7d203e3fff90c76bc0b45fb14fd7b
  Stored in directory: /home/user/.cache/pip/wheels/6c/c1/9a/10123be1e773c22deef65a047b8068bef5de3015c6cb6754bb
  DEPRECATION: Building 'wget' using the legacy setup.py bdist_wheel mechanism, which will be removed in a future version. pip 25.3 will enforce this behaviour change. A possible replacement is to use the standardized build interface by setting the `--use-pep517` option, (possibly combined with `--no-build-isolation`), or adding a `pyproject.toml` file to the source tree of 'wget'. Discussion can be found at https://github.com/pypa/pip/issues/6334
  Building wheel for wget (setup.py) ... done
  Created wheel for wget: filename=wget-3.2-py3-none-any.whl size=9685 sha256=15423cb1e044b854fa6dc0d6de54a09673cf2328f1fe3403615684b40b0a725c

```

**Comment by nichollsh** [comment]
Seems to work for me, too. Thanks Hanno. We can discuss this at the meeting.

**Comment by timlichtenberg** [comment]
Nice! My suggestion would be to reroute deprecation warnings to a log file if possible.

**Comment by timlichtenberg** [comment]
Great, will try it out asap

**Comment by nichollsh** [comment]
I've tried this on a completely 'fresh' Ubuntu PC which has never had Python/Julia/PROTEUS installed before. Works well.

The PROTEUS repo is around 500 MB, which is probably because of large files in the Git history. We should remove old files from the Git history of PROTEUS where possible. ZEPHYRUS also suffers from this problem. Can be done in another issue.

My main suggestions relate to the documentation:
* Troubleshooting page should include a section for when git is not installed
* Some of the conda packages (notably torch and nvidia stuff) are very large. Alongside the FWL_DATA files and code repos, we should probably warn the user beforehand how much disk space is required. Based on my network usage during the process, I think this comes to ~6 GB.
* Could provide an estimate for how long the install process takes (~30 minutes from scratch)
* User is prompted to `source ~/.bashrc` at end of install process. They should then be reminded to `conda activate `proteus` too.

Thanks for this Hanno!

**Review by nichollsh — APPROVED** [review]

**Comment by nichollsh** [comment]
I haven't seen this error before, but I don't think that it's directly due to this PR and would probably exist if you installed PROTEUS/AGNI in the 'usual' way without Hanno's script.

The error seems to be related to a conflict between the Julia and Conda installations of the HDF5 library:
```
juliacall.JuliaError: InitError: could not load library "/Users/timlichtenberg/.julia/artifacts/81e9ed542529a6847c4e5eca0c0f523592a7bb09/lib/libhdf5_fortran.310.3.2.dylib"
dlopen(/Users/timlichtenberg/.julia/artifacts/81e9ed542529a6847c4e5eca0c0f523592a7bb09/lib/libhdf5_fortran.310.3.2.dylib, 0x0001): Symbol not found: _H5FD_mpio_init
  Referenced from: <60B75FDB-18EC-368D-9593-64B0BC952353> /Users/timlichtenberg/.julia/artifacts/81e9ed542529a6847c4e5eca0c0f523592a7bb09/lib/libhdf5_f90cstub.310.3.2.dylib
  Expected in:     <88236FA7-0343-39DE-8A42-61913A33F1E4> /Users/timlichtenberg/miniforge3/envs/proteus/lib/libhdf5.310.dylib
```

I have seen similar conflicts between the Julia/Conda netcdf libraries too, which is related to HDF5. They can usually be avoided by installing AGNI outside of a conda environment, but doing that would subvert the whole purpose of having a single Python-based install script.

Not sure what the solution is, but I'll keep thinking. Tim, do you have an Intel-based Mac?

**Comment by timlichtenberg** [comment]
Apple M1 or M2 I think (not at home). If you think this has nothing to do with the PR I am happy to have this merged, so we can proceed. We can check out what the issue is later on.

**Comment by HannoSpreeuw** [comment]
> I have seen similar conflicts between the Julia/Conda netcdf libraries too, which is related to HDF5

Makes me think of [the discussion we had about this topic](https://github.com/FormingWorlds/PROTEUS/issues/440#issuecomment-3037050990).

@timlichtenberg Looking at your log, you seem to have miniforge installed. Curious if the error also occurs if you use [miniconda](https://www.anaconda.com/docs/getting-started/miniconda/install#macos-terminal-installer) instead.

**Comment by nichollsh** [comment]
One solution might be to pass the environment variable `JULIA_HDF5_PATH=/some/path` and/or `LD_LIBRARY_PATH=""` when calling the subprocess for the AGNI setup scripts. This should avoid Julia picking up the HDF5 library installed by conda.

**Comment by nichollsh** [comment]
@HannoSpreeuw I suggest that in the meantime we make the above mentioned documentation changes, and then merge this PR. We can look into this library issue as a separate PR and attempt to replicate it on various machines.

**Comment by HannoSpreeuw** [comment]
> * AGNI installation aborted because I had not `wget` installed, then I relaunched using the install-all script, but that did not do the trick

I was thinking whether I should add `wget`  as a requirement, but it does not seem to be applied in `AGNI/src/get_agni.sh`.

**Comment by nichollsh** [comment]
Should probably add wget as requirement, yeah. It's used by AGNI to get data from Zenodo.

**Comment by HannoSpreeuw** [comment]
Thanks.
The error comes from a missing "version" file in the SOCRATES directory.
I don't understand why it should be missing; a local "proteus install-all" puts this file where it should be.
What "proteus install-all" does not yet do is to pick up the FWL_DATA and RAD_DIR env vars when these have been set at the start of the installation, it uses default paths instead. 
Perhaps this invokes the error.

**Comment by nichollsh** [comment]
@HannoSpreeuw I think the reason it is missing is that the script installs socrates to `PROTEUS/PROTEUS/socrates` but then the code looks for it at `PROTEUS/PROTEUS/SOCRATES`. The difference seems to be in lower- vs upper-case.

**Comment by HannoSpreeuw** [comment]
Wow, that could be it!

**Comment by HannoSpreeuw** [comment]
[commit 1189b4b](https://github.com/FormingWorlds/PROTEUS/pull/456/commits/1189b4bdfb4f6796ea2f6392ffdfc5dd611774de) has taken me back to square one.

The problem is that in [get_socrates.sh](https://github.com/FormingWorlds/PROTEUS/blob/1189b4bdfb4f6796ea2f6392ffdfc5dd611774de/tools/get_socrates.sh#L15), we have 
```
socpath="socrates"
```
hardcoded as lower case.

So I guess I should revert that commit. 
But then we will again hit [this problem](https://github.com/FormingWorlds/PROTEUS/pull/456/commits/861972860d67f86addca0f46bf5c5a15e9a7bf79), i.e. that in our CI we have RAD_DIR [set to `.../ SOCRATES` (upper case)](https://github.com/FormingWorlds/PROTEUS/blob/1189b4bdfb4f6796ea2f6392ffdfc5dd611774de/.github/workflows/tests.yaml#L40):
```
RAD_DIR: ${{ github.workspace }}/SOCRATES
```
which will then lead to the `FileNotFoundError`:
```
        verpath = os.path.join(os.environ.get("RAD_DIR"),"version")
        with open(verpath, 'r') as hdl:
```
I am confused why this did not lead to the same `FileNotFoundError` in our CI before.

PR #465, opened 2025-07-24
fixing_instellation_bug

rdc49:
There is a bug in the 'orbit.wrapper' file. When calculating the bolometric luminosity for the dummy stellar module, it passes the stellar radii to the function in solar units, rather than SI units.


nichollsh (approved):
Looks good. Thanks for catching this! Happy to merge when you're ready.

PR #480, opened 2025-08-05
Tests for GridPROTEUS and BayesOpt inference schemes

### LINKED ISSUE #431: Unit tests for GridPROTEUS and BO inference wrappers
GridPROTEUS and the Inference wrappers are key tools when using PROTEUS. We should make sure that these components of the code are robust, particularly since they will be accessed by new users.

We must implement unit tests for these parts of the code. This should be relatively easy to do.

nichollsh:
1. Dummy atmos_clim sets atmosphere height as multiple of scale height.
2. Allow setting p_obs via config when using AGNI. Previously was hard coded.
3. Add tests for GridPROTEUS
4. Add tests for BayesOpt 

Closes #431 and #162.

The purpose of (1) is to allow the dummy atmosphere module to return `R_obs` values which are dependent on temperature and composition. It makes the tests more robust, and the model easier to debug, because the physics is now significantly more coupled and meaningful.

Also sets the number of OpenMP threads to 2 when performing tests. This improves test performance with VULCAN, since it was previously oversubscribing the workers.

Test coverage is now 69%, so it remains unchanged.


timlichtenberg:
While waiting for the tests to pass, is there a way how we can make the description of the tests more meaningful in this view here btw? I can only see "Run Coverage and Tests", but it would be good that the title specifies which tests these are.

nichollsh:
That's a good question. Indeed it would be useful to have this.

The name of the check "Run Coverage and Tests" comes from the name of the workflow in `.github/workflow/tests.yaml`.

It is already broken down more specifically in the logfile when the tests are run. For example, here: https://github.com/FormingWorlds/PROTEUS/actions/runs/16748871635/job/47413845065 we can see the output 
```
============================= test session starts ==============================
platform linux -- Python 3.12.11, pytest-8.4.1, pluggy-1.6.0
rootdir: /home/runner/work/PROTEUS/PROTEUS
configfile: pyproject.toml
testpaths: tests
plugins: anyio-4.10.0, dependency-0.6.0, jaxtyping-0.3.2
collected 33 items

tests/integration/test_integration_dummy.py .                            [  3%]
tests/integration/test_integration_physical.py ....XXXXXXXXXXXX          [ 51%]
tests/integration/test_integration_physical_agni.py ....                 [ 63%]
tests/plot/test_cpl_colours.py ..                                        [ 69%]
tests/plot/test_cpl_helpers.py .                                         [ 72%]
tests/test_cli.py ...                                                    [ 81%]
tests/test_config.py .....                                               [ 96%]
tests/test_init.py .                                                     [100%]
```

tests/grid/base.toml
```diff
@@ -0,0 +1,205 @@
+# PROTEUS configuration file (version 2.0)
+
+# ----------------------------------------------------
+# Metadata
+version = "2.0"
+author  = "Harrison Nicholls, Tim Lichtenberg"
```
timlichtenberg, line 6:
```suggestion
```

tests/inference/base.toml
```diff
@@ -0,0 +1,188 @@
+# PROTEUS configuration file (version 2.0)
+
+# ----------------------------------------------------
+# Metadata
+version = "2.0"
+author  = "Harrison Nicholls, Tim Lichtenberg"
```
timlichtenberg, line 6:
```suggestion
```

tests/inference/dummy.infer.toml
```diff
@@ -0,0 +1,30 @@
...
+# Parameters for Bayesian optimisation
+n_workers  = 2       # Number of parallel workers
+kernel     = "RBF"   # Kernel type for GP
+max_len    = 2      # Maximum number of evaluations
+n_restarts = 10      # GP optimization restarts
+n_samples  = 1000    # Raw samples for acquisition optimization
+
+# Parameters to optimize (with bounds)
+[parameters]
+"struct.mass_tot" = [0.7, 3.0]
```
timlichtenberg, line 22: I can these "" leading to chaos already. Is there no way to write this such that the " are added later on? Doesn't need to be in this PR.

nichollsh:
Maybe something like this? We can then generate a markdown file showing the summary of the tests.
https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#adding-a-job-summary

timlichtenberg (changes requested):

timlichtenberg:
> Maybe something like this? We can then generate a markdown file showing the summary of the tests. https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#adding-a-job-summary
> 
> This has some infrastructure for making summaries automatically: https://github.com/test-summary/action

I created a new issue for this.

PR #489, opened 2025-08-13
Track ocean quantities in PROTEUS, as calculated by AGNI

### LINKED ISSUE #488: Track ocean parameters in PROTEUS helpfile
AGNI now calculates ocean depth and area, given some rainout amount. Would be good to track this in PROTEUS since the variables can now be easily accessed from the AGNI atmosphere struct.

The configuration file for Earth would be an idea test case.

nichollsh:
Ocean depth, area, and composition are calculated by AGNI (version 1.7.2) based on the amount of rain reaching the surface. This PR adds functionality to store ocean area fraction and maximum depth in the PROTEUS output helpfile. Closes #488 

These ocean quantities set to zero if using the JANUS and dummy atmosphere modules.

Adapted the `earth.toml` configuration for demonstrating ocean formation. Also updated config parser to expose more parameters to the user, rather than keeping them hardcoded.

### DIFF
--- a/input/planets/earth.toml
+++ b/input/planets/earth.toml
@@ -10,13 +10,14 @@ version = "2.0"
         write_mod   = 2
 
     [params.dt]
-        starspec     = 3e6
+        starspec     = 1e7
         starinst     = 1e2
-        minimum      = 1e4
+        minimum      = 3e4
+        initial      = 3e4
 
         [params.dt.adaptive]
-            atol         = 0.02
-            rtol         = 0.10
+            atol         = 0.03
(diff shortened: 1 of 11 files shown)


lsoucasse (approved):
Looks good.

Do you know why the "Run Coverage and Tests" check is still pending?

nichollsh:
Thanks Laurent! 

I have no idea why it's pending. This has also happened for the last few pull requests too, but it apparently a known issue with GitHub sometimes.

I think if the test jobs themselves complete fine then we are safe.

PR #492, opened 2025-08-14
Termination criterion for planet entering star's roche limit

### LINKED ISSUE #487: Termination criterion for planet entering star's roche limit
The Roche limit around a star defines the distance at which an orbiting body (e.g. a planet) would disintegrate. This occurs because the tidal potential across the planet is greater than its own gravitational binding energy.

A planet's closest approach to its star (periastron) passing within the Roche limit should then represent a case at which PROTEUS should stop, because the planet would be destroyed.

MarijnJ0:
- Calculates the periapsis distance between star-planet (perihelion) and planet-satellite (perigee)
- Warns user if planet is partially orbiting within star's Roche limit (RL) (i.e. perihelion < RL)
- Adds new termination condition to stop simulation when time-averaged orbital separation < RL (Closes #487)
- Updates demo orbit config files to include the new configuration parameters

### DIFF
--- a/input/demos/orbit_dummy.toml
+++ b/input/demos/orbit_dummy.toml
@@ -4,7 +4,7 @@ version = "2.0"
 [params]
     [params.out]
         path        = "orbit_dummy"
-        logging     = "ERROR"
+        logging     = "INFO"
         plot_mod    = 0      # Plotting frequency, 0: wait until completion | n: every n iterations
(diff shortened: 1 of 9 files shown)


nichollsh (approved):
Works on my laptop as described. Thanks Marijn! Happy for this to be merged when all the tests pass and updates from the main branch are pulled into this one.

PR #498, opened 2025-08-30
Aragog return total heat flux for Fint

### LINKED ISSUE #497: Interior flux extracted from Aragog should include all flux terms
After running the interior model, PROTEUS extracts the net heat flux at the surface (`F_int`). In the case of Aragog this is currently just the convective heat flux, but in SPIDER it includes all flux terms.

This is problematic because it then 'looks' like Aragog is yielding a zero heat flux. But in reality, it's just that the convective heat flux is ~0 at the surface and the energy is instead carried by other processes.

PROTEUS should extract `F_int` from Aragog in the same manner as SPIDER by including all terms. Should be straightforward to do.

nichollsh:
Closes #497

### DIFF
--- a/src/proteus/interior/aragog.py
+++ b/src/proteus/interior/aragog.py
@@ -339,7 +339,7 @@ class AragogRunner():
                   "T_magma": aragog_output.solution_top_temperature,
                   "Phi_global": aragog_output.melt_fraction_global,
                   "RF_depth": aragog_output.rheological_front,
-                  "F_int": aragog_output.convective_heat_flux_basic[-1, -1]}
+                  "F_int": aragog_output.total_heat_flux_basic[-1, -1]}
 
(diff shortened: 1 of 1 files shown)


nichollsh:
This seems to have worked in that F_int is now non-zero at the top of mantle. However, it's very small. 

I noticed that the mixing and separation flux terms are disabled by default. Is there a reason for this? @planetmariana @timlichtenberg?

timlichtenberg (approved):
Looks good to me, but please let Mariana also review it before merging. The reason it was the convective flux was that the others were not yet present before, but it should have been updated indeed by now.

planetmariana (approved):
Thanks a lot  @nichollsh , indeed we needed to update this new feature from Aragog. Although, the instabilities in the upper and lower node from the convective heat flux are still there, but in previous test were not affecting the total heat flux, still we need to be careful.

PR #508, opened 2025-09-11
Provide "proteus update-all"

### LINKED ISSUE #468: PROTEUS update command
Following the successful integration of the simplified installation in #440 (thanks @HannoSpreeuw), it would be good to add a `proteus update` or similar command, that automates updates across the whole ecosystem, and ensures that all sub-modules and PROTEUS itself are up-to-date. This can work similar as `proteus doctor`, but should then automatically perform the required steps to update everything.

### LINKED ISSUE #503: Note to change python and julia paths to dataserver/users/formingworlds/... on norma1/2
Kapteyn student account home directories are capped at 9GB, which is less than required for the PROTEUS dependencies. We need to add a note to
https://fwl-proteus.readthedocs.io/en/latest/kapteyn_cluster_guide.html
and
https://fwl-proteus.readthedocs.io/en/latest/installation.html
to install miniconda/-forge/pyenv and Julia at `dataserver/users/formingworlds/{user}/` instead together with appropriate terminal commands.

@lorlo9999 and @MarijnJ0: can you suggest appropriate changes to the documentation?

HannoSpreeuw:
Fixes #468 

`proteus update-all` works on a PROTEUS user install.
Has also been tested for a developer install, but only on a single machine.

The update of the developer install will be done in the same way as for the user install, i.e. through `pip install` of the required Python packages, listed in [pyproject.toml](https://github.com/FormingWorlds/PROTEUS/blob/06e43c2c0291bd2a665021c23fc4011a0102dd74/pyproject.toml#L48)
(Btw, Janus is still listed there, should it be removed?)

More tests of `proteus update-all` are welcome!

The documentation has been updated to advertise `proteus update-all`.

Fixes #503 
The Kapteyn Cluster Guide has been updated to accommodate home folders with limited disk quota.


nichollsh:
@HannoSpreeuw please let me know when this PR should be reviewed.

nichollsh:
>  (Btw, Janus is still listed there, should it be removed?)

In principle JANUS still provides an important baseline, since it's comparable with atmosphere models used in other similar/older codes to PROTEUS. It's also part of the test suite. However, we should frame AGNI as the default atmosphere model, rather than JANUS.

nichollsh (changes requested):
See comments in https://github.com/FormingWorlds/PROTEUS/pull/508#issuecomment-3311107924

HannoSpreeuw:
> Additionally, it should update the data files in `FWL_DATA` as required, which could be done using the routines in `src/proteus/data.py`.

Ah, I guess your mean `src/proteus/doctor.py`

HannoSpreeuw:
Weird, I just ran `proteus update-all` on some cluster, next `proteus doctor`.
Oddly enough, this gave
`AGNI: Update available 1.7.5 -> Double-grey radiative transfer, diagnostic variables, and minor things`.

I did not expect this.
AGNI is updated through these lines as part of `update-all`:
```
        agni_dir = root / "AGNI"
        if agni_dir.exists():
            click.secho("🧪 Updating AGNI...", fg="blue")
            try:
                subprocess.run(["git", "pull"], cwd=agni_dir, check=True)
                combined_path = os.path.join(agni_dir, "src")
                print(f"{os.path.isdir(combined_path) = }")
                subprocess.run(
                    ["bash", "src/get_agni.sh"],
                    cwd=agni_dir,
                    env=os.environ,
                    check=True,
                )
                click.secho("✅ AGNI updated", fg="green")
```
These lines seems okay - except that there is also a `git pull` in `get_agni.sh`, so the first `git pull` could be removed, unless it also updates `get_agni.sh`, which it could, theoretically.
 Anyway, I will have to investigate what is done differently by `doctor`.

Issue #502, opened 2025-09-06
BayesOpt: Key error when atmosphere escapes


nichollsh:
The inference scheme includes a check for when the atmosphere has escaped. This sets the MMW and VMRs to zero, so that any comparison with a non-zero observational value is strongly excluded when the atmosphere is gone.

However, this does not seem to work properly because of a key error, which must be fixed.

Issue #494, opened 2025-08-14
BayesOpt inefficiently explores edges of parameter space


nichollsh:
I have noticed that the inference scheme spends a lot of time exploring the bounds of the parameter space, rather than focussing on regions which might be a better fit.

From Ben:
> This boundary exploring is a characteristic of the Upper Confidence Bound acquisition function. I have been meaning to change this to the Expected Improvement, since this also performed better for PROTEUS  in our paper. This is only a 5min task, since these are implemented in BoTorch.

PR #512, opened 2025-09-15
Merge Br/inference

Ben-Riegler:
Addresses #494 and also provides more control to user with the choice of acquisition function and GP-kernel. Also addresses #502.


nichollsh:
Thanks for this, Ben. Although it looks like you have modified every Python file in the PROTEUS codebase as part of this PR, hence the large number of changes. I think this is due to a automatic code style/linter tool that you have installed, since the changes all involve replacing double quotes (`"`) with single quotes (`'`), and removing whitespace.

I am not sure it's possible to revert these changes, and in principle they do tidy up the code, so we can probably leave them for now.

nichollsh (approved):
Looks good to me! Thanks again for this, @Ben-Riegler. The tests are now passing and the linting issues are fixed. Overall this seems to have resolved the problem were the optimisation over-explores the edges of the parameter space.

PR #518, opened 2025-10-01
Add config for GJ 9827 d, and new config option

nichollsh:
* Adds config for GJ 9827 d
* Introduce configuration option for disabling partitioning of volatiles into interior

This PROTEUS PR requires that the user update AGNI to version 1.7.7, which also allows running chemistry+condensation together self-consistently (https://github.com/nichollsh/AGNI/pull/140)


nichollsh:
It should soon be ready, yeah. I just need to check that these changes on the AGNI side are working fine.

nichollsh:
@timlichtenberg this is now ready for review :)

timlichtenberg (dismissed):
One comment on documentation

input/all_options.toml
```diff
@@ -134,6 +135,11 @@ version = "2.0"
     evolve  = false             # whether to evolve the SMaxis and eccentricity
     module  = "none"            # module used to calculate tidal heating
 
+    axial_period = "none"       # planet's initial day length; will use orbital period if 'none'
```
timlichtenberg, line 138: Units?

input/all_options.toml
```diff
@@ -134,6 +135,11 @@ version = "2.0"
...
+    satellite = false           # include satellite (moon)?
+    mass_sat = 7.347e+22        # mass of satellite [kg]
+    semimajoraxis_sat = 3e8     # initial SMA of satellite's orbit
```
timlichtenberg, line 141: units?

input/all_options.toml
```diff
@@ -197,7 +205,13 @@ version = "2.0"
         solution_rtol   = 2e-2          # solver relative tolerance
         overlap_method  = "ee"          # gas overlap method
         condensation    = true          # volatile condensation
+        latent_heat     = false         # include latent heat release by phase change?
```
timlichtenberg, line 208: it seems weird to have this off by default?

input/all_options.toml
```diff
@@ -197,7 +205,13 @@ version = "2.0"
...
         real_gas        = true          # use real-gas equations of state
+        psurf_thresh    = 0.1           # bar, surface pressure where we switch to 'transparent' mode
+        dx_max          = 35.0          # maximum temperature step allowed by solver
```
timlichtenberg, line 211: units?

input/all_options.toml
```diff
@@ -234,6 +249,8 @@ version = "2.0"
     tidal_heat      = false         # enable tidal heat production
     rheo_phi_loc    = 0.6           # Centre of rheological transition
     rheo_phi_wid    = 0.2           # Width of rheological transition
+    melting_dir     = "Monteux-600" # Name of folder constaining melting data
```
timlichtenberg, line 252:
```suggestion
    melting_dir     = "Monteux-600" # Name of folder constaining melting curves
```

input/all_options.toml
```diff
@@ -249,22 +266,33 @@ version = "2.0"
...
-        gravitational_separation        = false    # enable gravitational separation
-        mixing                          = false    # enable mixing
-        dilatation                      = false    # enable dilatation source term
-        mass_coordinates                = false    # enable mass coordinates
-        tsurf_poststep_change           = 30       # threshold of maximum change on surface temperature
-        event_triggering                = true     # enable events triggering to avoid abrupt jumps in surface temperature
+        logging                  = "ERROR"  # Aragog log verbosity
+        num_levels               = 50       # Number of Aragog grid levels
+        tolerance                = 1.0e-7   # solver tolerance
+        initial_condition        = 3        # Initial T(p); 1: linear, 2: user defined, 3: adiabat
```
timlichtenberg, line 272:
```suggestion
        initial_condition        = 1        # Initial T(p); 1: linear, 2: user defined, 3: adiabat
```

timlichtenberg: "3" often leaves the mantle half melted when larger planets are chosen. This has led to confusion in some projects already. "1" is the safer option.

input/all_options.toml
timlichtenberg, line 1: Please update the minimum toml file with the same new settings

timlichtenberg (changes requested):
A few suggestions to make the .toml file better understandable

input/all_options.toml
nichollsh: The solver is less reliable with latent heat enabled, so I figured it would be better to have a configuration which 'just works' as the default, since people will be likely to adapt it for their own purposes.

input/all_options.toml
nichollsh: Should we change the default value then?

input/all_options.toml
nichollsh: Kelvin. Added.

nichollsh:
Tests are failing because I need to update the reference data, given the switch from an adiabatic to linear mantle profile.

PR #534, opened 2025-10-20
Access grid tooling with CLI, rather than as individual scripts

### LINKED ISSUE #514: Move `grid_summarise.py` tool into main `src/` codebase
The main method for interacting with PROTEUS is through the command line interface (CLI). It's now possible to dispatch single-runs through this, but now also large grids: `proteus grid -c [cfg]`.

The last remnant of the grid script being a standalone executable is the file `tools/grid_summarise.sh`. 

Ideally, this utility should be incorporated into the main command line interface, so that it can be used when PROTEUS is installed just as a package. We would then have something like `proteus gridstatus -c [cfg]` as equivalent to `./tools/grid_summarise.py [gridOutput]`.

This is separate from, but related to, the issue #396

nichollsh:
* Move `grid_summarise` script into CLI - now run by `proteus grid-summarise` command. This summarises the status of a grid, to help diagnose any problems and track its completeness.
* Move `pack` script into CLI - now run by `proteus grid-pack` command. This packages the most important data from the grid into a single zip file.
* Add information into docs about how to use these utilities.
* Add unit test which runs a simple grid and executes these utilities.
* Fix issue with `proteus doctor` command when checking for AGNI updates.
* Tidied up `tools/` folder.

Closes #514

### DIFF
--- a/src/proteus/grid/summarise.py
+++ b/src/proteus/grid/summarise.py
@@ -93,6 +93,7 @@ def summarise(pgrid_dir:str, tgt_status:str=None):
                 print("  (None)")
 
     # code cases
+    tgt_status = tgt_status.replace("status=", "code=")
     if "code" in tgt_status:
         matched = True
         code = int(tgt_status.replace(" ","").split("=")[-1])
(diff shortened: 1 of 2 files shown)


planetmariana (approved):
Awesome @nichollsh , looks good to me :)

PR #549, opened 2025-10-27
Final step in fixing 426

### LINKED ISSUE #426: Switch to Aragog fork
I was discussing with Dan recently whether we should switch our Aragog developments to a fork hosted on github.com/orgs/FormingWorlds/. Since we are at the moment the only group developing these, this would give us more control over the development. We should discuss this in the weekly meeting.

HannoSpreeuw:
## Description
A [PyPi entry called `fwl-aragog`](https://pypi.org/project/fwl-aragog/) has been created for our [Aragog fork](https://github.com/FormingWorlds/aragog).

This seems to work in the sense that `python -c "import aragog"` works after a `pip install fwl-aragog`.
We need to see if CI tests pass.

PROTEUS' `pyroject.toml` now makes use of this PyPI entry.
This means that `proteus install-all` and `proteus update-all` will `pip install fwl-aragog`.

The documentation wrt the developer install has been updated to use the fork.
A `grep` on the `docs` folder did not yield any more hits wrt moving from `Explanetology` to `FormingWorlds`, but I may have missed something.

Closes #426 

Publishing to PyPI of the [latest Aragog fork release](https://github.com/FormingWorlds/aragog/releases/tag/25.10.27) does [not seem to work yet](https://github.com/FormingWorlds/PROTEUS/issues/426#issuecomment-3450771092).

Probably related: what should happen with the [`pyproject.toml` that I used](https://github.com/FormingWorlds/PROTEUS/issues/426#issuecomment-3442332070) to create our [PyPi entry](https://pypi.org/project/fwl-aragog/) since it differs from the [current `pyproject.toml`](https://github.com/FormingWorlds/aragog/blob/56687e7d8ee0f504a5489a590b741dc0600eb4c6/pyproject.toml) and it should, as far as I can tell, since it serves a difference purpose, namely building the tar and wheel files.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,7 +49,7 @@ dependencies = [
     "fwl-mors>=25.07.15",
     "fwl-calliope>=25.05.01",
     "fwl-zephyrus>=25.03.11",
-    "fwl-aragog>=v0.2.8-alpha",
+    "fwl-aragog>=25.10.27",
     "fwl-zalmoxis>=25.09.07",
     "cmcrameri",
     "juliacall",


HannoSpreeuw:
@nichollsh Right, so in the [new pyproject.toml](https://github.com/FormingWorlds/PROTEUS/blob/b339b46de7a1dcbc46b0ad206f4f4e24f31f89a2/pyproject.toml) all poetry sections have been removed, afaict. 

Quite different from the [pyproject.toml in current main](https://github.com/FormingWorlds/aragog/blob/e997f003395126b4fb882f6966d956aa3a3cc5f4/pyproject.toml#L1).

nichollsh:
Yes, this comparison by @HannoSpreeuw is between the two repos - which have always had substantially different methods for building and packaging the projects. I did not make large changes to either of the `pyproject.toml` files.

Shall we merge this?

HannoSpreeuw:
> Isn't that PROTEUS vs. aragog, or am I missing something?

O man, sorry for the confusion. 

I guess I thought you'd be needing something like [this pyproject.toml for Aragog](https://github.com/FormingWorlds/PROTEUS/issues/426#issuecomment-3442332070) to generate the wheel and tar files, but apparently you don't.

PR #552, opened 2025-10-30
SPIDER config options: flux terms, smoothing width

### LINKED ISSUE #542: Option for dynamically turning-off the energy transport mechanisms via SPIDER wrapper
The wrapper around SPIDER inside PROTEUS (`spider.py`) contains hard-coded flags which always enable the full set of energy transport flux terms. However, when using Aragog is possible to turn these on/off via the PROTEUS configuration file.

It would be very useful to be able to turn off the energy flux terms in SPIDER, as required by the user. For benchmarking against Aragog, for testing the physics, and for comparison with fully-convective boundary layer models of interior dynamics.

Doing this will be straightforward, and only require modifications in PROTEUS: to `spider.py` and the config parser functionality to add these new keys.

@rdc49

nichollsh:
## Description
This PR adds TOML keys for configuring SPIDER, removing hard-coded behaviour of some parameters. It's now possible to enable/disable the different flux terms in SPIDER via the PROTEUS TOML files. It's now also possible to configure the window-width for smoothing material properties.

Unfortunately cannot add tests for this, because SPIDER is not handled by the GitHub workflow.

Closes #542. 

## Validation of changes
I ran the L9859d and GJ9827d cases for several iterations, and verified that they behaves the same as they did previously.

Using Habrok, Linux 4.18, Python 3.12, Julia 1.11

## Checklist

- [x] I have followed the [contributing guidelines](https://fwl-proteus.readthedocs.io/en/latest/CONTRIBUTING.html#how-do-i-contribute)

## Relevant people
Related to @rdc49's ongoing project.

### DIFF
--- a/src/proteus/config/_interior.py
+++ b/src/proteus/config/_interior.py
@@ -14,7 +14,7 @@ def valid_spider(instance, attribute, value):
 
     # at least one energy term enabled
     spider = instance.spider
-    if not (spider.conduction or spider.convection or spider.mixing or spider.gravitation_separation):
+    if not (spider.conduction or spider.convection or spider.mixing or spider.gravitational_separation):
         raise ValueError("Must enable at least one energy transport term in SPIDER")
 
 def valid_path(instance, attribute, value):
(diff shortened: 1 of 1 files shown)


planetmariana (approved):
Thanks a lot @nichollsh ! now we can have both aragog and spider with the same input parameters !

nichollsh:
Thank you both!

PR #562, opened 2025-11-17
Full physical integration of PROTEUS on CI

### LINKED ISSUE #437: Full physical integration of PROTEUS on CI
We have been talking about this for a while. We should implement one full physical integration of standard PROTEUS settings on the CI. This should not be coupled to PRs, as it will run too long, but it will be critically important for monitoring the long-term health of the codebase. In addition, it should be coupled to a routine that produce a flame graph as an output, so we can routinely analyse the performance bottlenecks of the code.

HannoSpreeuw:
Fixes #437 

1) Launches proteus with "all_options.toml" configuration 
2) Uploads its plots to "output/full_physical/plot"


HannoSpreeuw:
> Does this happen only when running `./tools/get_lovepy.sh` after doing `proteus install-all`? On the GitHub workflow, lovepy is installed _before_ PROTEUS. It could be that installing PROTEUS does something to the Julia environment that causes this problem.

Interesting question. So I changed the order:

1. pip install -e .
2. ./tools/get_lovepy.sh
3. proteus install-all

but the error persists.
But perhaps I should start off with ` ./tools/get_lovepy.sh`.

PR #596, opened 2026-01-26
VULCAN online chemical kinetics calculations

ipanagiotou02:
## Description
(This is a draft PR; copied comment from Issue #6)
In the branch `ip_vulcan`, the online calculation has been implemented such that it calls VULCAN at every loop (as of now). It creates a `vulcan.csv` file at every snapshot corresponding to the `.nc` file of the same time. For example, if the file `66097_atm.nc` is created in the data directory, `vulcan_66097.csv` is created in the offchem directory. This is still a post-processing approach, but it occurs at every snapshot instead of just at the end.

From my project, I knew that a simulation with `orbit.semimajoraxis = 0.0316` and `outgas.fO2_shift_IW = -4.0` converges relatively quickly (required 29 minutes), so I used it as an initial test. The results are shown below:

fc_vs_vulcan_vmr_fo2-4_1000inst

A few issues/remarks to make are the following:

- Time: VULCAN's runtime can range from 10 minutes to more than 1 hour, depending on the number of steps it needs to converge. In most cases, convergence is difficult to achieve, and it takes all 30,000 steps to run. This adds a lot of time for a simulation run (the above case went from 29 minutes to 5.29 hours).

- Memory: due to the above, VULCAN greatly increases the size of the proteus_00.log file (from 650 KB to 22 MB). Additionally, each additional .csv file adds 650 KB, and each additional .pkl file adds 5 MB. The example simulation produced 29 of these files times 2 for each, increasing the total simulation memory by 160 MB. For the all_options case, a run with 6 GB allowed the simulation to run for 63 loops and produce 57 vulcan.csv files/snapshots, but without converging. Here is the result up to the point it ran:

fc_vs_vulcan_vmr_all_options

Any feedback that you have is appreciated!

## Checklist

## Relevant people
@timlichtenberg @nichollsh @shami-EEG


nichollsh (changes requested):
The tests are failing because the `run_chemistry` function is returning `None` rather than the Pandas DataFrame of mixing ratios. See comment below about avoiding test regression for offline chemistry, and adding new test for online chemistry.

Also the code style check is failing, just because of some whitespace discrepancies. Also should be easy to fix.

Are the offline/online/manual chemistry 'modes' described in the documentation anywhere? The docs should also briefly explain these differences and their reasons for other users.

In principle these changes look good. Should be easy to fix, and would then be ready to be merged into main :)

timlichtenberg:
@nichollsh I think this should be merged. Can you please go through this qualitatively once and check for any major issues? If there are none we should merge it. As far as I understand this now covers fastchem3 inside AGNI and will still have the inconsistency with the fastchem2 in VULCAN, but technically it enables coupled disequilibrium chemistry runs, which is a good step.

PR #647, opened 2026-02-23
fix(doctor): use semantic versioning for update checks

### LINKED ISSUE #646: Proteus doctor command suggests downgrade
## Problem description
The `proteus doctor` command is meant to help a user diagnose their proteus installation, check which modules need updating, and confirm environment variables are set.

However, it seems to suggest an "update is available" by prompting a version downgrade. Also, it does not check for  the `FC_DIR`, `ZALMOXIS_ROOT`, or `LA_DIR` environment variables.

## Expected behavior
* Respect ascending order of version numbering by only suggesting upgrades to later versions. I think this is because the code files express more recent version numbers than the GitHub releases.
* Check missing environment variables
* Would also be good if it printed Python, Julia (etc.) versions, since we have specific requirements on these
* Could also check the SOCRATES, SPIDER, and FastChem versions.

## Evidence
```
$ proteus doctor
Packages
aragog: No package metadata was found for aragog
fwl-calliope: ok
fwl-janus: ok
fwl-proteus: Update available 25.11.19 -> 25.10.15
fwl-mors: ok
fwl-zephyrus: Update available 25.10.28 -> 25.3.11
fwl-zalmoxis: ok
AGNI: Update available 1.8.6 -> 1.8.1

Environment variables
FWL_DATA: ok
RAD_DIR: ok
```

## My computer
Fedora 43, Python 3.12, Julia 1.11

## Relevant people
@timlichtenberg @stuitje @egpbos

DebuggingMax:
## Summary
Fixes #646 - The doctor command was incorrectly suggesting 'upgrades' that were actually downgrades (e.g., 25.11.19 -> 25.10.15).

## Changes
- **Fixed version comparison**: Use `packaging.version.Version` for proper semantic version comparison instead of simple string inequality
- **Added environment variable checks**: FC_DIR, ZALMOXIS_ROOT, LA_DIR now included in the doctor output
- **Added runtime version display**: Python and Julia versions are now shown in a new 'Runtime versions' section
- **Added unit tests**: New tests for version comparison logic to prevent regression

## Technical Details
The bug occurred because the original code used `current_version != latest_version` which doesn't understand semantic versioning. For example, `25.11.19` would be considered 'different' from `25.10.15`, triggering an update suggestion even though `25.11.19` is actually newer.

The fix uses `packaging.version.Version` to properly compare versions, only suggesting updates when `current_ver < latest_ver`.

## Testing
- Added unit tests in `tests/test_doctor.py` specifically for version comparison
- Updated existing CLI tests in `tests/test_cli.py` to verify new output sections

### DIFF
--- a/src/proteus/doctor.py
+++ b/src/proteus/doctor.py
@@ -2,6 +2,9 @@ from __future__ import annotations
 
 import importlib.metadata
 import os
+import platform
+import shutil
+import subprocess
 from functools import partial
 from importlib.metadata import PackageNotFoundError
 from typing import Callable
@@ -9,6 +12,7 @@ from typing import Callable
 import click
 import requests
 from attr import dataclass
+from packaging.version import Version, InvalidVersion
 
 from proteus.utils.coupler import _get_agni_version, get_proteus_directories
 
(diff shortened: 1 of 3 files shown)


timlichtenberg:
No bot PRs please, thanks.

PR #665, opened 2026-03-28
Initial support for aerosols in PROTEUS, improve chemistry plot, and fix broken links in readme

### LINKED ISSUE #663: Configure AGNI aerosol/haze parametrisation via PROTEUS
## Feature description
AGNI now supports aerosols and hazes, in addition to water clouds. This interface still needs to be developed to coupled it with chemical modelling, and is disabled by default. PROTEUS needs to set the relevant function arguments to enable these effects, ideally configurable via the existing TOML config parser.

## Preferred solution
Optionally enable and configure aersols/clouds/hazes in AGNI via the PROTEUS config parser.

## Additional information
Related to https://github.com/nichollsh/AGNI/issues/119 and https://github.com/nichollsh/AGNI/issues/119

## Relevant people
@timlichtenberg

### LINKED ISSUE #664: Links in PROTEUS readme are broken
## Problem description
Several of the links in the PROTEUS readme file are broken. For example: https://proteus-framework.org/PROTEUS/usage.html

## Expected behavior
Links go to the correct page.

## My computer
Firefox on Fedora Linux.

## Relevant people
@stuitje @timlichtenberg

nichollsh:
## Description
* Adds option for enabling aerosols in AGNI radiative transfer. Closes #663.
* Downloads aerosol scattering data from Zenodo.
* Reads aerosol MMR profiles from AGNI output NetCDF files.
* Updates chemistry plot to include aerosol and cloud profiles, with T(p) for context
* Fixes broken links in readme. Closes #664.

This PR aims to _introduce_ aerosol support via PROTEUS. We will need to re-visit it via future PRs, when some kind of microphysical/chemical scheme is implemented in AGNI to derive aerosol mixing ratio profiles.

## Validation of changes
PROTEUS runs fine and makes the new plot (see below). Note that the aerosol profiles are at MMR=0 because AGNI does not yet have a scheme for calculating these profiles. Also added tests for these new features (manually and using Copilot), which pass on my laptop and on the github runners.

image

## Checklist

## Relevant people
@timlichtenberg

### DIFF
--- a/tests/plot/test_cpl_chem_atmosphere.py
+++ b/tests/plot/test_cpl_chem_atmosphere.py
@@ -40,9 +40,7 @@ def test_plot_chem_atmosphere_no_files(mock_glob):
 @patch('proteus.plot.cpl_chem_atmosphere.read_ncdf_profile')
 @patch('proteus.plot.cpl_chem_atmosphere.glob.glob')
 @patch('proteus.plot.cpl_chem_atmosphere.os.path.join')
-def test_plot_chem_atmosphere_basic(
-    mock_join, mock_glob, mock_read, mock_subplots
-):
+def test_plot_chem_atmosphere_basic(mock_join, mock_glob, mock_read, mock_subplots):
     """
     Test basic plot generation with gas species only.
 
@@ -111,9 +109,7 @@ def test_plot_chem_atmosphere_basic(
 @patch('proteus.plot.cpl_chem_atmosphere.read_ncdf_profile')
 @patch('proteus.plot.cpl_chem_atmosphere.glob.glob')
 @patch('proteus.plot.cpl_chem_atmosphere.os.path.join')
-def test_plot_chem_atmosphere_with_clouds(
-    mock_join, mock_glob, mock_read, mock_subplots
-):
+def test_plot_chem_atmosphere_with_clouds(mock_join, mock_glob, mock_read, mock_subplots):
     """
     Test plot generation including cloud profiles.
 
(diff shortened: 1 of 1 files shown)


timlichtenberg (changes requested):
Awesome work! When fully integrated this will open a whole new category of applications. A few things to address before merging, most important is a potential crash when plotting output from runs without aerosols.

1. **`KeyError` on `atm_profile['aerosols']`** in `cpl_chem_atmosphere.py`: the aerosol loop accesses this key unconditionally, but `read_ncdf_profile` only populates it when the NetCDF has aerosol data. The cloud block right above correctly guards with `if 'cloud_mmr' in atm_profile.keys()`, but the aerosol block does not. This will crash on any pre-aerosol run output. Should be `atm_profile.get('aerosols', [])`.

2. **`num_aerosols` starts at 1, not 0.** This inflates the legend column count. If there are no aerosols and no cloud, it still counts 1. Should start at 0 and increment separately when the cloud is plotted.

3. **`fig.text` with `transform=ax1.transAxes`**: this works by accident because (0.02, 0.98) in `ax1` axes coords happens to land in the right spot. Should be `ax1.text(...)` instead. If the layout changes this will silently break.

4. **`aer_mmr` reading in `common.py`**: the guard checks `'aer_mmr' in ds.variables` but then unconditionally reads `ds.variables['aerosols']` for species names. If the file has one but not the other, this crashes. Add `and 'aerosols' in ds.variables.keys()` to the guard.

5. **No OSF fallback for scattering data.** All other entries in `DATA_SOURCE_MAP` have both Zenodo and OSF IDs. I know we move to other sources, but should we add it until then?

PR #668, opened 2026-04-14
Boundary interior module

### LINKED ISSUE #525: Implement middle-complexity boundary layer interior evolution model
It would be useful, at least for the purposes of comparison, to implement a boundary-layer model for mantle evolution in PROTEUS.

This would calculate a potential temperature and heat flux which evolves over time, assuming that the mantle is fully convective and adiabatic. The top CBL thickness would evolve over time too.

Should first be done directly in PROTEUS without creating a new module for it.

Related: https://ui.adsabs.harvard.edu/abs/2019GeoJI.217...75W/abstract

@timlichtenberg @planetmariana

rdc49:
This pull request is for the boundary interior module. This is a 0D, parameterised mantle evolution model, based on the model in Schaefer et. al. (2016). It is similar to the dummy module, but with some key differences:

-The ODE solver solves for the surface temperature and the mantle potential temperature for a given time step
-The surface temperature decreases via the cooling set by the atmosphere, and increases via heat exchange with the mantle. The differential equation uses the heat capacity of the atmosphere and a boundary layer of variable thickness.
-The mantle potential temperature decreases via heat exchange with the surface and increases via latent heat release, radiogenic heating and tidal heating. It uses the heat capacity of the molten mantle.
-Heat exchange between the mantle operates via parameterised convection (equations 14 and 15 in the Schaefer paper)
-Latent heat release is modelled assuming that the radius of solidification is a simple function of the melt fraction.

The model works with AGNI, ZALMOXIS and lovepy.


nichollsh:
> The main issue it has when trying to operate alongside AGNI is time-stepping: large atmospheric fluxes result in too steep a decrease in surface temperature for a given time step. In theory, this could be overcome by playing with the adaptive time-stepping parameters, but this is fiddly, and results in the model taking a very long time to run.

As mentioned earlier today, @rdc49, one potential option to avoid taking too-large timesteps is to incorporate 'callback' events. We already do this in SPIDER and Aragog, where the time-integration of the interior is stopped prematurely if the $T_s$ or $\Phi$ change more than a set threshold during a single iteration.

You can do this with `solve_ivp` by setting `events=[event_tsurf, event_phi]` where `event_tsurf` and `event_phi` are functions that return a float. The return should switch from >0 to <0 when the 'event' occurs, so e.g. `event_tsurf` would return `tsurf_old - tsurf_maxdiff - tsurf_new`. 

https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.solve_ivp.html

input/ensembles/example.infer.toml
nichollsh, line 1: Please undo changes to this file.

input/ensembles/boundary_interior_test.toml
nichollsh, line 1: Is it necessary to keep all of these different config files in the `ensembles/` folder? I worry it will become cluttered by files used for one-time purposes. Perhaps you can keep all these config files stored somewhere else safely?

input/demos/escape.toml
nichollsh, line 1: The changes made to this config will break, with the new syntax introduced by Karen.

input/chili/protocol/tr1b.toml
```diff
@@ -124,7 +137,7 @@ solve_energy = true
 solution_atol = 0.1
 solution_rtol = 0.1
 overlap_method = "ee"
-rainout = false
+condensation = false
```
nichollsh:
```suggestion
rainout = false
```

input/demos/sat_physical.toml
```diff
@@ -125,7 +125,7 @@ version = "2.0"
         solution_atol   = 1e-3          # solver absolute tolerance
         solution_rtol   = 2e-2          # solver relative tolerance
         overlap_method  = "ee"          # gas overlap method
-        rainout   = false         # volatile condensation
+        condensation    = false         # volatile condensation
```
nichollsh:
```suggestion
        rainout    = false         # volatile condensation
```

input/chili/intercomp/_base.grid.toml
```diff
@@ -3,13 +3,13 @@ symlink = ""
 ref_config = "input/chili/intercomp/_base.toml"
 use_slurm = true
 
-max_jobs = 9       # maximum number of concurrent tasks (e.g. 500 on Habrok)
+max_jobs = 16       # maximum number of concurrent tasks (e.g. 500 on Habrok)
```
nichollsh:
```suggestion
max_jobs = 9       # maximum number of concurrent tasks (e.g. 500 on Habrok)
```

input/chili/intercomp/_base.grid.toml
```diff
@@ -3,13 +3,13 @@ symlink = ""
...
 max_days = 1        # maximum number of days to run (e.g. 1)
 max_mem  = 3        # maximum memory per CPU in GB (e.g. 3)
 
 ["delivery.elements.H_kg"]
     method = "direct"
-    values = [1.60e20, 7.80e20, 16e20]
+    values = [1.60e20, 4.70e20, 7.80e20, 16e20]
```
nichollsh:
```suggestion
    values = [1.60e20, 7.80e20, 16e20]
```

input/chili/intercomp/_base.toml
nichollsh, line 1: Please revert changes made to this file

input/chili/protocol/earth.toml
nichollsh, line 1: Please revert changes made to this file

input/chili/protocol/tr1b.toml
nichollsh, line 1: Please revert changes made to this file

input/chili/readme.md
```diff
@@ -1,6 +1,5 @@
 Configuration files for the CHILI model intercomparison project.
 Website: https://nexss.info/cuisines/chili-mip/
```
nichollsh, line 2:
```suggestion
Website: https://nexss.info/cuisines/chili-mip/
GitHub for CHILI: https://github.com/projectcuisines/chili
```

input/planets/l9859d.toml
nichollsh, line 1: Please revert changes made to this file

input/planets/trappist1c.toml
nichollsh, line 1: Can probably just delete this file...

src/proteus/atmos_clim/wrapper.py
```diff
@@ -147,7 +151,9 @@ def run_atmosphere(
             if no_atm:
                 activate_julia(dirs, config.atmos_clim.agni.verbosity)
                 # surface temperature guess
-                hf_row['T_surf'] = hf_row['T_magma']
+
+                if config.interior.module != 'boundary':
+                    hf_row['T_surf'] = hf_row['T_magma']
```
nichollsh: Why does the boundary layer model require special treatment here?

src/proteus/config/_interior.py
```diff
@@ -186,6 +196,108 @@ def valid_interiordummy(instance, attribute, value):
...
+
+    rtol: float = field(default=1e-6, validator=gt(0))
+    atol: float = field(default=1e-9, validator=gt(0))
+
+    T_p_0: float = field(default=3500.0, validator=ge(0))
+
+    T_solidus: float = field(default=1420.0, validator=ge(0))
+    T_liquidus: float = field(default=2020.0, validator=gt(0))
+
+    critical_melt_fraction: float = field(default=0.4, validator=(gt(0), lt(1)))
```
nichollsh: This config option is effectively duplicating the existing one called `rheo_phi_loc`. We should not have both.

src/proteus/config/_interior.py
```diff
@@ -186,6 +196,108 @@ def valid_interiordummy(instance, attribute, value):
...
+    thermal_expansivity: float = field(default=2e-5, validator=gt(0))  # 1/K
+
+    # Viscosity parameterisation
+    viscosity_model: int = field(
+        default=2, validator=in_((1, 2, 3))
+    )  # 1=constant, 2=aggregate, 3=Arrhenius
+    eta_constant: float = field(default=1e2, validator=gt(0))  # Pa s, for model 1
+
+    # Aggregate viscosity parameters
+    transition_width: float = field(default=0.2, validator=(gt(0), lt(1)))  # -
```
nichollsh, line 284: I think this is also a duplicate of the existing parameter `rheo_phi_wid`? Should not have both, if they represent the same thing.

src/proteus/interior/wrapper.py
```diff
@@ -83,6 +86,39 @@ def determine_interior_radius(dirs: dict, config: Config, hf_all: pd.DataFrame,
...
+    # then run the boundary module once to populate thermodynamic outputs.
+    if config.interior.module == 'boundary':
+        M_target = config.struct.mass_tot * M_earth
+        rho_core = config.struct.core_density
+        rho_silicate_fixed = config.interior.boundary.silicate_density
+        corefrac_r = config.struct.corefrac
+
+        # Mixed bulk density implied by a core radius fraction corefrac_r.
+        rho_bulk = rho_core * corefrac_r**3 + rho_silicate_fixed * (1.0 - corefrac_r**3)
+        hf_row['R_int'] = float((3.0 * M_target / (4.0 * np.pi * rho_bulk)) ** (1.0 / 3.0))
```
nichollsh: This is not quite right, because `M_target` includes the mantle+core **and** the volatiles. You cannot simply substitute `M_target` into this equation to get `R_int` analytically, because `M_target` is not the mass of the mantle+core. While, at the same time, the mass of volatiles is a function of the mantle mass, because we often define these in terms of ppmw.

I suggest adopting the same method as used by the dummy interior. Perhaps they can be combined?

src/proteus/interior/zalmoxis.py
nichollsh, line 1: Why update the EOS stuff in this file for Zalmoxis?

src/proteus/proteus.py
```diff
@@ -405,11 +409,27 @@ def start(self, *, resume: bool = False, offline: bool = False):
             )
 
             ############### INTERIOR
+            if self.config.interior.module == 'boundary' and self.loops['total'] == 0:
+                self.hf_row['M_mantle'] = (
+                    self.config.struct.mass_tot * M_earth - self.hf_row['M_core']
```
nichollsh: This is also not correct, because `mass_tot` includes the volatiles.

src/proteus/interior/boundary.py
```diff
@@ -0,0 +1,705 @@
...
+
+        cp_layer = getattr(getattr(atmos_o, '_atm', None), 'layer_cp', None)
+        if cp_layer is not None:
+            cp_arr = np.asarray(cp_layer, dtype=float).ravel()
+            valid = np.isfinite(cp_arr)
+            if np.any(valid):
+                # Use profile-mean atmospheric heat capacity across all valid layers.
+                self.atmosphere_heat_capacity = float(np.mean(cp_arr[valid]))  # J/kg/K
+            else:
+                self.atmosphere_heat_capacity = 1.7e4  # J/kg/K for H2 at 2000K
```
nichollsh: I suggest making this a config option, rather than a hardcoded value.

nichollsh: Also, same goes for its usage on the line just below.

src/proteus/interior/boundary.py
```diff
@@ -0,0 +1,705 @@
...
+        # Prefer Zalmoxis-derived CMB radius when available.
+        # Fall back to corefrac-based radius for non-Zalmoxis runs.
+        self.core_radius = float(
+            hf_row.get('R_core', config.struct.corefrac * self.planet_radius)
+        )
+        self.core_frac = self.core_radius / self.planet_radius
+
+        self.mantle_radius = self.planet_radius - self.core_radius
+        self.mantle_volume = (4 / 3) * np.pi * (self.planet_radius**3 - self.core_radius**3)
+        self.mantle_mass = self.planet_mass - self.core_mass
```
nichollsh: Again, this is not correct because planet mass includes volatiles.

src/proteus/interior/boundary.py
```diff
@@ -0,0 +1,705 @@
...
+        -------
+        float
+            Rayleigh number [dimensionless]
+        """
+        # Determine viscosity based on configuration
+        eta = self.viscosity(T_p, T_surf, phi)
+
+        # Calculate Rayleigh number
+        Ra = (
+            self.bulk_density
```
nichollsh: This `bulk_density` variable name is confusing, since it's not actually the bulk density. It's the mantle density.

src/proteus/interior/boundary.py
```diff
@@ -0,0 +1,705 @@
...
+            Time derivatives [dT_p/dt, dT_surf/dt, drs/dt] where:
+            - dT_p/dt is the rate of change of potential temperature [K/s]
+            - dT_surf/dt is the rate of change of surface temperature [K/s]
+        """
+        T_p, T_surf = y
+
+        if T_surf > T_p:
+            T_surf = (
+                T_p - 1.0
+            )  # Ensure surface temperature does not exceed potential temperature
```
nichollsh: I am not sure that this makes sense. In some cases, I can imagine that Ts could exceed Tp. For example, if the planet is exposed to more irradiation at a later stage of its life (e.g. late migration of super Earths is expected). Could you remove this?

src/proteus/interior/boundary.py
```diff
@@ -0,0 +1,705 @@
...
+
+        output = {
+            'T_magma': T_p_final,
+            'T_pot': T_p_final,
+            'T_surf': T_surf_final,
+            'F_int': self.f_atm,
+            'Phi_global': phi_final,
+            'Phi_global_vol': phi_final,
+            'F_radio': f_radio_final / (4 * np.pi * self.planet_radius**2),
+            'RF_depth': phi_final * (1.0 - self.core_frac),
```
nichollsh, line 722: Is this right? The `phi_final` is a melt fraction by mass, while the `RF_depth` is a volume-based variable. They might not scale linearly in a spherical geometry, like this.

src/proteus/config/_interior.py
```diff
@@ -186,6 +196,108 @@ def valid_interiordummy(instance, attribute, value):
...
+    dynamic_viscosity: float
+        Reference dynamic viscosity [Pa s] for Arrhenius solid mantle model.
+    activation_energy: float
+        Activation energy [J/mol] for Arrhenius solid mantle model.
+    viscosity_prefactor: float
+        Viscosity prefactor [Pa s] for Vogel-Fulcher-Tammann magma ocean model.
+    viscosity_activation_temp: float
+        Activation temperature [K] for Vogel-Fulcher-Tammann magma ocean model.
+    logging: bool
+        Whether to include interior logging in the output.
```
nichollsh:
```suggestion
        Whether to create diagnostic CSV data files from boundary interior module.
```

nichollsh (changes requested):
Thank you, @rdc49, it is very cool to see this boundary layer module already implemented. This will be beneficial for application to sub-Neptunes, benchmarking the other modules, comparison with other codes (e.g. GOOEY), and for usage in papers which are not specifically focussed on the interior dynamics resolved by SPIDER/Aragog.

I have a few comments below. Some of them simply ask that you restore some of the files that were changes unnecessarily, and suggest removing duplicated config options. However, there are some more substantial issues with how mantle mass is treated.

nichollsh:
I've made an initial code review for this PR above. Once these issues are resolved/closed, I will be very happy to run the new module and perform another review. Lots of valuable changes here, so would love to help get this merged during this week.

src/proteus/interior/zalmoxis.py
rdc49: Sorry, I forgot to flag this up in the PR. I couldn't get Zalmoxis to work with the tuple containing the dictionaries, so I got co-pilot to make this edit to the code. I assumed this was a bug in Zalmoxis, but it might be possible that I'm using an outdated version. However, when I run proteus doctor, it says that my version of Zalmoxis is up to date.

src/proteus/interior/boundary.py
rdc49: Are the mass melt fraction and the volume melt fraction not the same thing, assuming that the molten material and the solid material have the same density?

nichollsh: If they have the same density, then yes, that's true. I suppose if the boundary layer model assumes that the two densities are the same then this approximation is fine to keep. It's definitely better to have a simpler scaling here.

Although it should be noted that both of the melt/solid densities are strong functions of temperature and pressure. This means that the solid>liquid density in the upper mantle, and (curiously) also means that liquid>solid density in the deep mantle.

nichollsh:
@rdc49 please feel free to re-request my review of this PR when you are ready. I will check it then.

rdc49:
Hi Harrison, I've made all the changes to the branch you requested. I'm happy for you to have another look at the pull request.

src/proteus/interior/boundary.py
```diff
@@ -0,0 +1,709 @@
...
+            Rate of change of surface temperature [K/s]
+        """
+        phi = self.melt_fraction(T_p)
+        q_m_val = self.q_m(T_p, T_surf, phi)
+
+        if T_p == T_surf:
+            delta = 1e-3  # Small temperature difference to avoid division by zero
+        else:
+            delta = self.thermal_conductivity * (T_p - T_surf) / q_m_val
+
```
nichollsh, line 569: The `delta` variable representing the CBL thickness should be stored in the helpfile, since it's important and evolving over time. Please make sure that this quantity is stored.

Also, if it's being stored for the 'boundary' module, it should also be stored by the output from SPIDER/Aragog/Dummy too. These will be constant over time.

nichollsh (changes requested):
Overall looks great! Thanks again, @rdc49. 

I have one comment about storing `delta` over time, since it's useful to record this. There are also some tests which need to be updated/expanded, and the code style check is currently failing.

Should be quick to fix these.

nichollsh:
@rdc49, thanks for making these final changes. I noticed a few more minor issues that would be easier to correct directly rather than describe here. Also, the boundary.py tests were all failing because of some syntax issues that I have now fixed. I will make these changes directly in the branch and merge it manually (since GH will block my review after modifying the branch).

Really excellent to have this module in PROTEUS, so thanks again for adding it. There are immediate applications for it, and longer term utility too.

PR #669, opened 2026-04-20
Fix doctor command and check for SOCRATES updates

### LINKED ISSUE #646: Proteus doctor command suggests downgrade

nichollsh:
## Description
* Uses the `packaging`  package to reliably check if modules are up to date with `proteus doctor`
* Checks SOCRATES against latest release on FormingWorlds/SOCRATES
* Checks correct `fwl-aragog` package, rather than deprecated `aragog` one
* Add tests for `proteus doctor` command
* Did not add new documentation since the command is already described

Closes #646 

Future PRs should also check the Obliqua repo for updates, once a release has been made for it.

## Validation of changes
Works on Fedora 43 with Python 3.12:
```
(proteus) [16:47:41] harrison@delphinium:~/Projects/PROTEUS
$ proteus doctor
Packages
fwl-aragog: ok
fwl-calliope: ok
fwl-janus: ok
fwl-proteus: Local version 25.11.19 is newer than latest release 25.10.15
fwl-mors: ok
fwl-zephyrus: Local version 25.10.28 is newer than latest release 25.3.11
fwl-zalmoxis: ok
AGNI: ok
SOCRATES: ok

Environment variables
FWL_DATA: ok
RAD_DIR: ok
ZALMOXIS_ROOT: ok
FC_DIR: ok
LA_DIR: Variable not set.
```

## Checklist

## Relevant people

@timlichtenberg @egpbos


nichollsh:
@egpbos @timlichtenberg the patch coverage is now 100% and the tests are passing, so the PR is ready for review.

egpbos:
Maybe good to make an issue to remind ourselves that we need to add Obliqua checking after its release.

nichollsh:
@egpbos, I have now done this. Please could you review this PR when you are available?

src/proteus/utils/coupler.py
egpbos, line 83: Outside the scope of this PR, but just a minor comment about naming convention: leading-underscore functions are supposed to be "private" / module-internal (e.g. `from proteus.utils.coupler import *` does not import `_get_socrates_version`). The fact that this one is imported elsewhere is suspicious ;) I think in this case this function could simply be a "public" function without an underscore, no?

src/proteus/doctor.py
```diff
@@ -30,15 +35,25 @@ def latest_version(self) -> str: ...
 
     def get_status_message(self) -> str:
         try:
-            current_version = self.current_version()
-            latest_version = self.latest_version()
+            current_version = Version(self.current_version())
```
egpbos: Could we also return a Version object from `current_version` and `latest_version` directly? This would help debuggability, since we try-capture two calls here that could both fail, better to have them fail separately at the point of origin.

Btw, what exceptions do we expect here that we need a try-except at all? If it's meant to capture unimplemented child class methods then another way to do that that avoids a try-except block is to use `@abc.abstractmethod`.

egpbos:
@nichollsh I already started 😄

src/proteus/doctor.py
```diff
@@ -30,15 +35,25 @@ def latest_version(self) -> str: ...
...
-            message = click.style(str(exc), **ERROR_STYLE)
+            message = click.style(f'{exc.__class__.__name__} - {exc}', **ERROR_STYLE)
+
         else:
-            if current_version != latest_version:
+            if latest_version > current_version:
                 message = click.style(
                     f'Update available {current_version} -> {latest_version}', fg='yellow'
                 )
+            elif latest_version < current_version:
```
egpbos, line 49: Nice that this is captured. Maybe add a "Eagerly awaiting your PR!" message 😉

src/proteus/doctor.py
```diff
@@ -73,18 +88,30 @@ def latest_version(self) -> str:
         response = requests.get(
             f'https://api.github.com/repos/{self.owner}/{self.name}/releases/latest'
         )
-        return response.json()['tag_name']
+        if not response.ok:
+            response.raise_for_status()
+
+        # handle
```
egpbos:
```suggestion
```

Pointless comment.

src/proteus/doctor.py
egpbos:
```suggestion
        response.raise_for_status()
```

src/proteus/doctor.py
```diff
@@ -73,18 +88,30 @@ def latest_version(self) -> str:
...
+        try:
+            return response.json()['tag_name']
+        except KeyError as exc:
+            raise InvalidVersion(
+                f'Could not retrieve latest version for {self.name} from GitHub.'
+            ) from exc
```
egpbos: The `response.raise_for_status()` will already throw on a 404 or other HTTP error, so the try-except is never reached. The only situation the try-except would then catch is e.g. an API change on the GitHub side (right now, the only "ok" code GitHub returns is 200, in which case the `tag_name` should be there). If we want that, we should probably change the error message to say that. But unless there's something I'm overlooking, I would opt to simplify as follows:

```suggestion
        response.raise_for_status()
        return response.json()['tag_name']
```

Note btw that `response.ok()` itself calls `response.raise_for_status()`, so the `if` check is not necessary :) I also suggested removing that above in the `PythonPackage` class.

tests/test_doctor.py
```diff
@@ -0,0 +1,149 @@
...
+    package = DummyPackage('fwl-proteus', current='main', latest='v25.11.19')
+
+    message = package.get_status_message()
+
+    assert 'Update available' not in message
+    assert "InvalidVersion - Invalid version: 'main'" in message
+
+
+@pytest.mark.unit
+def test_doctor_checks_expected_environment_variables():
```
egpbos: What does this test, except the existence of a hardcoded set?

tests/test_doctor.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+@pytest.mark.unit
+def test_git_package_latest_version_reads_tag_name_from_json_response():
+    package = GitPackage(name='SOCRATES', owner='FormingWorlds', version_getter=Mock())
+    response = Mock(ok=True)
+    response.json.return_value = {'tag_name': 'v2026.01'}
+
+    with patch('proteus.doctor.requests.get', return_value=response):
+        assert package.latest_version() == 'v2026.01'
```
egpbos: If we change the return class to a `Version` object, this should be modified accordingly.

tests/test_doctor.py
```diff
@@ -0,0 +1,149 @@
...
+
+@pytest.mark.unit
+def test_git_package_latest_version_raises_invalid_version_without_tag_name():
+    package = GitPackage(name='SOCRATES', owner='FormingWorlds', version_getter=Mock())
+    response = Mock(ok=True)
+    response.json.return_value = {'name': 'latest'}
+
+    with patch('proteus.doctor.requests.get', return_value=response):
+        with pytest.raises(
+            InvalidVersion, match='Could not retrieve latest version for SOCRATES'
```
egpbos: With my suggestion above for `GitPackage.latest_version`, this will raise just a KeyError. Essentially, this would then really just be a test of the GitHub API, so I would remove this test (if my suggestion is accepted).

src/proteus/doctor.py
egpbos: The GitHub suggestion doesn't really work here, so just to clarify: the `if not response.ok:` can simply be removed, it is superfluous (see also suggestion for `GitPackage.latest_version()`).

egpbos (commented):
Looks good! Some suggestions below.

src/proteus/utils/coupler.py
nichollsh: I agree. Same probably goes for the other `_get_*_version` functions too. Can do this in another PR.

src/proteus/doctor.py
nichollsh: A few different exceptions. For example, when the request fails for Git-based packages it will throw an exception that will propagate to this try/catch statement.

tests/test_doctor.py
nichollsh: It checks that this set of variables is consistent. However, I agree that this is probably more burdensome than it is useful, since updating `VARIABLES` then requires updating this test. I'll remove it.

src/proteus/doctor.py
nichollsh: Done

nichollsh:
@egpbos I implemented your suggestions about the return type of the version-getter functions, and updated the tests. They pass on my computer.

egpbos:
Great! One final thing: you added example.grid.toml in dea586e, was that intended?

nichollsh:
Thanks for catching this, @egpbos - it was added by mistake.

egpbos (approved):
LGTM! :ship:

PR #634, opened 2026-02-16
Initial support for LavAtmos and ThermoEngine (rock vapour calculations)

leojola:
## Description

This pull request is co-authored by @leojola and @nichollsh together.

Touches on several parts of the code with the main intent of including rock vapours into the outgassing scheme. Firstly, PROTEUS runs the normal volatile partitioning calculation. Then, it runs LavAtmos+ThermoEngine to determine the amount of rock vapours produced. These calculations are then summed together. This does not conserve the total mass, since the atmosphere is grown by vapourisation but the mantle is not shrunk by an equivalent amount of mass. This is documented in the documentation. 

There is an example configuration in the `input/tutorials/` folder.

There are also QoL changes to the code. Includes plotting improvements.

### Rock vapour support
- Added the option of including rock vapour calculations in PROTEUS, by coupling with LavAtmos. If `vapourise=true` is set in the input file, LavAtmos and CALLIOPE are run separately and then their results are summed together.
- Stores `P_vap` and `P_vol` as helpfile keys - these sum to `P_surf`. Updates the total MMW consistently with this.
- Stores the fO2 calculated by LavAtmos
- Adds a new configuration section `outgas.lavatmos` with relevant keys. These include numerical and physical options.

### Remove albedo lookup
- Removed unused albedo lookup functionality, originally developed for this project, and no longer needed

### Dependencies
- LavAtmos needs to be installed using the script added to `tools/`
- ThermoEngineLite needs to be installed using the script added to `tools/`

### Tests
- Added tests for this new vapourisation functionality, including on the lavatmos coupling and the PROTEUS wrappers
- Removed tests relating to the albedo-lookup stuff, which is no longer present

### Quality of Life (QoL) updates
- Removed duplicate "resume" banner in logs
- Adds rock vapours to the PROTEUS plotting functions
- Updated `cpl_global` and `cpl_escape` for readability with large numbers of species. Updated `cpl_global` x-axis label for clarity
- Updated log formatter to write "WARNING" as "WARN" so that log line-prefixes have consistent lengths
- Plots are made every `plot_mod` iterations without being gated by the `dt_write_rel`. This was previously being implicitly limited by the data-writer gate, although this was never intended, since the `plot_mod` check is separate.

### Development
- Closes #215 
- Closes https://github.com/FormingWorlds/ThermoEngineLite/issues/2
- Closes #726 

## Checklist

[...]

PR #688, opened 2026-06-08
Improved BayesOpt performance using new stopping criterion. Documentation info for IoA cluster. Catch BL errors.

### LINKED ISSUE #680: Boundary layer interior model silently erroring
## Problem description

The new boundary layer mantle dynamics model sometimes fails to integrate during a PROTEUS time-step. The error is "Required step size is less than spacing between numbers." 

This occurs silently during a PROTEUS run, at the moment, and leads to `dt=0` and the simulation not making any progress. Eventually, it will exit when the max iterations are reached.

## Expected behavior

This error should be captured from `solve_ivp` and at least logged. Perhaps adjustments can be made to the tolerance at runtime, to avoid these scenarios?

E.g. adding the following code after the `solve_ivp` call in `boundary.py`...

```python
        # Check reason for solver termination
        if sol.status == 1:
            log.info("Surface temperature change triggered time-step trucation")
        elif sol.status == -1:
            log.warning("Boundary layer interior failed to integrate - solver error")
            log.warning(f"    solver msg: {sol.message}")
```

## My computer
Running PROTEUS on Habrok. Git hash `7c5ff466b977`. Julia 1.12 and Python 3.12.

## Additional notes
Potentially resolved by ongoing PR: https://github.com/FormingWorlds/PROTEUS/pull/678

## Relevant people
@timlichtenberg @egpbos @rdc49

### LINKED ISSUE #686: Install instructions for PETSc on IoA cluster
## Feature description
The install instructions describe specific steps for getting things working on different clusters. There's nothing for the IoA cluster so far, and I had to make some custom changes. These should be documented.

## Preferred solution

* Install python with miniforge
* Set Julia `JULIA_DEPOT_PATH` to data server, rather than home directory
* Install Julia into data server, and **DO NOT** follow the instructions about setting `PYTHON_JULIAPKG_EXE`.
* `module load netcdf openmpi mpich cuda`
* Update the `if` statement in `get_petsc.sh` to retain the `--download-f2cblaslapack=1` flag for IoA cluster. At the moment, this isn't passed because the `/etc/redhat-release` check is satisfied.

I am not sure that we even need to set `PYTHON_JULIAPKG_EXE` anymore, since we use Julia 1.12?

## Relevant people
@rdc49

### LINKED ISSUE #687: Stopping criterion on wall-clock runtime
## Feature description
Sometimes simulations get 'stuck', leading to very long wall-clock runtimes. We already track runtime using the `runtime` variable in the helpfile (in seconds since simulation started). Stopping the simulation when this exceeds a certain value would be useful. This would see immediate application in the BO scheme.

## Preferred solution
* Add new stopping criterion called `clock`, with a `clock.maximum` attribute that indicates the maximum time for which the simulation can run
* If exceeds this time, the model exits with status code=11 (this code is currently unused).

### LINKED ISSUE #700: PROTEUS sometimes leaves `hf_row` keys empty, leading to incorrect helpfile data
## Problem description

I noticed some weird 'jumps' in the `R_obs` value recorded by the helpfile. This was because pandas is interpreting my simulated `T_obs` value as the radius. This is because the number of columns in the helpfile row is now sometimes inconsistent between iterations.

## Expected behavior
We should ensure that `hf_row` values are always filled. Even with zeros or NaNs, no matter the physics being simulated.

## Evidence

Image

## My computer

IoA Cluster running RedHat Linux, Python 3.12, Julia 1.12

## Additional notes
Happening in my branch linked to PR #688, after merging recent main branch. I suspect this is also happening on the main branch, too.

## Relevant people
@timlichtenberg @egpbos

nichollsh:
## Description

These changes improve the robustness, usability, and performance of the existing Bayesian Optimisation scheme. They also include other modifications to PROTEUS  that arose during this part of my BayesOpt research.

### BayesOpt updates

* BO scheme now log10-transforms parameters appropriately, rather than just the observables. This means that they are sampled on appropriate scales.
* Adds new stopping criterion for when runtime exceeds a certain value (closes #687)
* Can specific failure states for BO scheme. This means that models can terminate under these conditions (e.g. solidification) and represent disfavourable outcomes with very negative objective values
* Updates configs for ongoing BO paper

### Doc updates

I have run PROTEUS on the IoA cluster and noticed a few specifics.

* Modifications to `get_petsc.sh` to enable install on IoA cluster (closes #686)

### Boundary layer model fixes

While running sims for the BO paper, I have been using the new `boundary` interior module. This includes some fixes.

* Boundary layer model prints warning when failing (closes #680)
* Correctly calculate an 'average' atmosphere $C_p$ by weighting layer values with their mass

### Atmosphere model fixes

* Option to enable/disable temperature dependent thermodynamic atmospheric properties when using AGNI (`thermo_functions`).
* Only write atmosphere NetCDF files when `is_snapshot=True`, which brings the atmosphere I/O in line with the interior energetic I/O behaviour.
* Ensure that NaN values are not returned from atmosphere diagnostics. These were propagating into missing helpfile columns (see #700). 
* The reference optical depth is now derived from a single index, which can later be obtained from a given wavelength, rather than averaging over all bands arbitrarily.

## Validation of changes

* Tests pass on IoA cluster (RedHat with Python 3.12, Julia 1.12)
* Tests pass on GitHub Ubuntu latest runner (Python 3.12)

## Checklist

## Relevant people
@stuitje @rdc49 @timlichtenberg


nichollsh:
Tests now passing. @rdc49 @stuitje @timlichtenberg, this is now ready for a first review. I don't really want to expand this PR any more since it already covers multiple issues.

stuitje:
> Tests now passing. @rdc49 @stuitje @timlichtenberg, this is now ready for a first review. I don't really want to expand this PR any more since it already covers multiple issues.

Thanks for including me as reviewer :) I will review this tomorrow, hope that works!

nichollsh:
No problem, @stuitje, thanks for taking the time to look at it!

Just another comment: I have made some additional changes in order to mitigate the issue highlighted in https://github.com/FormingWorlds/PROTEUS/issues/700. This issue should be addressed more robustly elsewhere, but an immediate patch is also required.

https://github.com/FormingWorlds/PROTEUS/issues/700#issuecomment-4671786405

nichollsh:
@timlichtenberg @stuitje @rdc49 The tests are now passing again. Sorry for making additional changes after requesting the first review - I'll try to avoid making any more in this branch.

src/proteus/inference/objective.py
```diff
@@ -348,18 +365,31 @@ def f(x_norm: torch.Tensor) -> torch.Tensor:
...
             ref_config=ref_config,
             output=output,
+            failure_codes=failure_codes,
         )
+
+        J_eval = J_context(x_raw)
+
+        # Check J is finite
+        if not torch.isfinite(J_eval).all():
+            x_param = {parameters[i]: x_raw[0, i].item() for i in range(d)}
```
stuitje: shouldn't you use param_keys here instead of parameters (which is a dict, correct?) ?

docs/How-to/inference.md
stuitje, line 16: Should now also contain `transforms.py`

src/proteus/atmos_clim/agni.py
```diff
@@ -123,65 +122,81 @@ def _summarise_tau_band(atmos) -> tuple[float, float]:
...
+    t_conv_arr = np.nan_to_num(t_conv_arr, nan=0.0)
+    t_rad_arr = np.nan_to_num(t_rad_arr, nan=0.0)
+
+    # Get maximum Rayleigh number
+    Ra_max = float(np.amax(ra_arr))
+
+    # Get maximum t_conv_t_rad
+    mask_c = ra_arr > 1e-9
+    if np.any(mask_c):
+        ratio = np.amax(t_conv_arr[mask_c] / t_rad_arr[mask_c])
```
timlichtenberg: This drops the divide-by-zero floor the old code had (`max(t_rad, 1e-300)`). `t_rad_arr` is `nan_to_num`'d to 0.0 just above, but `mask_c` selects on `Ra > 1e-9`, so a convective level whose radiative timescale was NaN now divides by zero and `np.amax` returns `inf`. That flows into `hf_row['atm_t_conv_over_t_rad']`, and the `ExtendHelpfile` guard uses `pd.isna`, which doesn't catch `inf`, so it reaches the CSV. Only a diagnostic column, but worth restoring the floor:

```suggestion
        ratio = np.amax(t_conv_arr[mask_c] / np.maximum(t_rad_arr[mask_c], 1e-300))
```

src/proteus/atmos_clim/agni.py
```diff
@@ -123,65 +122,81 @@ def _summarise_tau_band(atmos) -> tuple[float, float]:
...
+    wl_idx = max(0, min(wl_idx, nbands - 1))
+
     # Accept a cell-centre-sized level axis (nlev_l - 1) too: the TOA and
     # surface values sit at indices 0 and -1 on either grid, so the
     # reduction is identical and the helper tolerates both conventions.
     level_sizes = {nlev_l, nlev_l - 1}
     if tau_arr.ndim == 2 and tau_arr.shape[0] in level_sizes and tau_arr.shape[1] == nbands:
-        toa = float(tau_arr[0, :].mean())
-        surf = float(tau_arr[-1, :].mean())
+        toa = float(tau_arr[0, wl_idx])
```
timlichtenberg, line 155: `tau_arr[0, wl_idx]` has no finite guard, so an `inf` from an RT divergence in the chosen band propagates into the helpfile (a NaN gets caught downstream, an inf doesn't). Also: the `return 0.0, 0.0` error paths in this function collide with a physically real tau=0 (transparent atmosphere), so a read failure is indistinguishable from a genuine transparent column. A cheap guard on the indexed reads (here and lines 155-156 for the transposed case) fixes the inf case: `float(...) if np.isfinite(...) else 0.0`.

src/proteus/interior_energetics/boundary.py
```diff
@@ -48,24 +48,33 @@ def __init__(
...
+            [config.interior_energetics.boundary.atm_heat_capacity],
+        )
+        ma_layer = getattr(getattr(atmos_o, '_atm', None), 'layer_σ', [1.0])
+
+        # Get extracted values
+        cp_arr = np.array(cp_layer, dtype=float)
+        ma_arr = np.array(ma_layer, dtype=float)
+
+        # Skip NaN values
+        valid = np.isfinite(cp_layer) & np.isfinite(ma_layer)
```
timlichtenberg: `valid` is computed from the raw `getattr` results (which can be juliacall proxies) rather than the already-converted arrays. Simpler and avoids relying on `np.isfinite` on the proxy:

```suggestion
        valid = np.isfinite(cp_arr) & np.isfinite(ma_arr)
```

Separately, on line 76 `np.average(cp_arr, weights=ma_arr)` raises `ZeroDivisionError` if the surviving layer masses sum to zero; the `len == 0` fallback doesn't cover an all-zero-mass column, so that should fall back to the config default too.

src/proteus/utils/constants.py
```diff
@@ -44,7 +44,7 @@
 
 # Supported gases
 vol_list = ['H2O', 'CO2', 'O2', 'H2', 'CH4', 'CO', 'N2', 'NH3', 'S2', 'SO2', 'H2S']
-vap_list = ['SiO', 'SiO2', 'MgO', 'FeO2']
+vap_list = ['SiO']  # just one for now
```
timlichtenberg, line 47: Cutting `vap_list` from 4 species to 1 removes ~33 helpfile columns (11 per gas via `GetHelpfileKeys`). Two follow-ons: (a) resuming a run whose CSV carries the old SiO2/MgO/FeO2 columns may mismatch the current schema, worth checking against the resume path; (b) the fixture at `tests/outgas/test_wrapper.py:700-702` still sets `SiO2_kg_atm`/`MgO_kg_atm`/`FeO2_kg_atm`, now dead keys. If the reduction is deliberate, fine, just flagging the resume angle. (The other constant removals in this file have no remaining references, so those are clean.)

tools/migrate_data/v2_defaults.json
```diff
@@ -235,6 +236,8 @@
  "params.out.remove_sf": false,
  "params.out.write_mod": 1,
  "params.resume": false,
+ "params.stop.clock.enabled": true,
```
timlichtenberg, line 239: This turns the clock stop on for every migrated v2 config (7-day max), but the attrs `StopClock.enabled` default is `False`. So a fresh v3 config has the clock off while a migrated one silently gets a one-week wall-clock termination nobody asked for. Either default the attrs class to match, or drop `enabled: true` here so it stays opt-in.

tests/utils/test_terminate.py
```diff
@@ -106,6 +108,16 @@ def test_check_solid_skips_when_above_crit(patch_statusfile):
     assert patch_statusfile == []
 
 
+@pytest.mark.unit
+def test_check_clock_skips_when_below_max(patch_statusfile):
+    """Clock: below maximum keeps simulation running."""
+    cfg = _cfg()
+    h = _handler(cfg)
+    h.hf_row['runtime'] = 500.0
+    assert terminate._check_clock(h) is False
```
timlichtenberg, line 117: This covers only the below-max branch. `_check_clock`'s actual job, terminating with status 11 when `runtime >= maximum`, has no test, unlike every other `_check_*` here which has a trigger+skip pair. Worth adding a `test_check_clock_triggers_at_max` that sets `runtime = maximum` and asserts the return is `True` and status 11 was written.

tests/inference/test_objective.py
```diff
@@ -113,6 +113,7 @@ def test_run_proteus_success_handles_escaped_atmosphere(monkeypatch, tmp_path):
     assert obs['P_surf'] == pytest.approx(0.0)
     assert obs['atm_kg_per_mol'] == pytest.approx(0.0)
     assert len(updates) == 2
+    assert status >= 0
```
timlichtenberg: The mock doesn't write a status file, so `status` falls back to the default 20; `status >= 0` is then trivially true and pins nothing (a regression changing the fallback to any non-negative code still passes). Either assert `status == 20` with a note that the file is absent, or write a real status file and assert the intended code. Same at line 243.

tests/atmos_clim/test_agni.py
```diff
@@ -647,56 +649,39 @@ def __array__(self, *_a, **_k):
...
-    level (smallest index where mask_c is True).
-
-    Discriminating: a regression that took the LAST convective level
-    (np.where(mask_c)[0][-1]) would land at index 3, where
-    t_conv / t_rad = 1e2 / 1e4 = 1e-2. The correct topmost index is
-    1, where the ratio is 1.0e3 / 1.0e6 = 1e-3. Order-of-magnitude
-    separation.
-    """
+def test_summarise_diagnostics_picks_convective_level():
+    """Get diagnostics at correct level"""
```
timlichtenberg: Two things: the docstring no longer states the physical scenario (the lint rule wants that), and the test now pins `max(t_conv/t_rad)` over all convective levels while the name `_picks_convective_level` and the `physics_invariant` marker imply the old "topmost level" contract. A value that distinguishes "topmost" from "any convective" would keep the discrimination the old test had.

tests/inference/test_plot.py
```diff
@@ -33,9 +33,11 @@
 import toml
 
 import proteus.inference.plot as plot_mod
+import proteus.inference.transforms as transforms_mod
 
 pytestmark = [pytest.mark.unit, pytest.mark.timeout(30)]
 
+pytestmark = [pytest.mark.unit, pytest.mark.timeout(30)]
```
timlichtenberg: Duplicate `pytestmark` (also on line 38); the second silently overwrites the first. Harmless but dead, drop one.

timlichtenberg (changes requested):
Thanks Harrison, strong PR overall and the BayesOpt transforms and tests are genuinely good. One blocker before merge: the `_summarise_diagnostics` divide-by-zero can write `inf` into the helpfile (inline note on agni.py). Two behavioural surprises worth confirming as intentional rather than fixing: the migrated-config 7-day clock default, and the `vap_list` schema reduction with its resume implication. The rest are test-tightening and one dead duplicate line. Details inline.

input/inference/example.infer.toml
stuitje, line 1: there is still an `input/example.infer.toml` present as well, I would remove it to avoid confusion

input/inference/example.infer.toml
```diff
@@ -0,0 +1,71 @@
...
+seed = 42
+
+# Logging level ("DEBUG", "INFO", "WARNING", "ERROR")
+logging = "INFO"
+
+# Path to output folder where inference will be saved (relative to PROTEUS output folder)
+output = "bayesopt_infer_se/"
+
+# Path to base (reference) config file relative to PROTEUS root folder
+ref_config = "output/bayesopt_se/init_coupler.toml"
```
stuitje: Perhaps a good idea to add a line about this in the documentation, for people running the example as-is as a quick test, to change this to an appropriate PROTEUS config (otherwise there will be FileNotFound error)

src/proteus/inference/objective.py
nichollsh: Yep. Good catch. Will fix.

docs/How-to/inference.md
nichollsh: Added

src/proteus/atmos_clim/agni.py
nichollsh: Ideally we should not get NaN values for t he convective and radiative timescales, but I agree it should be handled robustly. I'll implement this suggestion.

src/proteus/atmos_clim/agni.py
nichollsh: > collide with a physically real tau=0 (transparent atmosphere), so a read failure is indistinguishable from a genuine transparent column

I'll add code to raise a warning if the optical depth is zero, so that a user knows if/when this occurs. Will add a guard here too to check for NaN or Inf.

src/proteus/interior_energetics/boundary.py
nichollsh: 1. Good catch - thanks. Implemented your first suggestion.
2. Now checks for sum=0, and falls back to default in that case.

src/proteus/utils/constants.py
nichollsh: This was an intentional decision. The rock vapour physics is being handled properly by Leoni via Lavatmos in a separate branch. In the meantime, arbitrarily adding a bunch of species just adds simulation overhead by requiring AGNI to load their thermodynamic data into memory, from disk.

I've updated the test to remove the dead keys.

tests/utils/test_terminate.py
nichollsh: Added another test which checks for the termination criterion being triggered.

tests/inference/test_objective.py
nichollsh: Now checks that `status==20` rather than being triggered to another non-20 value by the `failure_keys`.

tests/inference/test_plot.py
nichollsh: Removed - thanks

tests/atmos_clim/test_agni.py
nichollsh: Agreed. Previously the behaviour was to pick the topmost convective level, but it's more intuitive (with the new behaviour) to find the layer with the strongest convection, which is why I made these changes. However, I missed updating the function names appropriately here.

Updated test name + docstring.

input/inference/example.infer.toml
nichollsh: Removed. Thanks. I think having these in their own directory makes more sense, and we don't need both.

input/inference/example.infer.toml
nichollsh: I've now updated the `ref_config` to point to a real file.

tools/migrate_data/v2_defaults.json
nichollsh: Good catch. I've updated the `attrs` class `StopClock` to match by enabling the termination criterion by default. Seems like any simulation running for >1 week would be problematic, so enabled by default makes sense.

src/proteus/inference/objective.py
nichollsh: Yep. Fixed.

nichollsh:
Thanks, @timlichtenberg and @stuitje, for the constructive reviews! These are really useful comments and suggestions.

I've responded to each of your comments above, and made the required changes in the latest commits. Please take another look at this when you have the time.

rdc49 (approved):
These changes to the boundary layer look good, thanks for catching that error with the atmospheric heat capacity. What was the setup you used that caused the boundary module to get stuck on a timestep of 0? The stopclock feature is also a nice addition.

Out of curiosity, why was the 'conduction' parameter for AGNI changed to false rather than true in the default setup? Also, why were SiO2, MgO and FeO2 removed from the list of rock vapor species?

input/inference/example.toml
```diff
@@ -0,0 +1,603 @@
...
+    surf_state                = "fixed"       # energy balance: mixed_layer | fixed | skin
+    surface_d                 = 0.01         # conductive skin thickness [m]
+    surface_k                 = 2.0          # conductive skin conductivity [W/m/K]
+
+    # Solver limits (agni, janus only)
+    tmp_minimum               = 0.5          # temperature floor [K]
+
+    [atmos_clim.agni]
+
+        spectral_file   = "output/runtime.sf"
```
stuitje, line 532: This file doesn't exist for me, perhaps it point to a spectral file in FWL_DATA?

timlichtenberg (approved):
Thanks Harrison, all my points from the previous round are addressed and verified against the latest commits. The divide-by-zero floor on the timescale ratio, the finite guard plus transparent-mode handling on the optical depth reduction, the boundary-layer cp fix using the converted arrays with a zero-mass fallback, and the param_keys fix in the objective warning path all look correct. The test tightening (status pin at 20, the new clock trigger test, the renamed diagnostics test) is good too. I ran the affected unit suites locally and they pass.

Two small non-blocking notes for later, no need to hold this PR:

- The new clock trigger test uses runtime = 1e99, which covers the trigger branch but not the exact runtime == maximum boundary. Optional to tighten.
- Flipping StopClock.enabled to True is a sensible global runaway guard; worth remembering that runtime resets per invocation, so it caps a single continuous run rather than a resumed campaign.

Approving.

input/inference/example.toml
nichollsh: The intended use-case of this key is to provide a path to a spectral file that already contains a stellar spectrum + rayleigh coeffs. This is in contrast to one of the FWL_DATA files.

nichollsh:
Thanks everyone! To respond to these final comments...

> What was the setup you used that caused the boundary module to get stuck on a timestep of 0? 

I haven't found any consistent patterns in when this happens, unfortunately.

> Out of curiosity, why was the 'conduction' parameter for AGNI changed to false rather than true in the default setup? 

It doesn't make any physical difference in most scenarios. It's intended for future work where we consider nighside columns and the conduction becomes more important in the upper atmosphere. Makes sense to disable for now.

> Also, why were SiO2, MgO and FeO2 removed from the list of rock vapor species?

They were initially added arbitrarily and only increased the 'setup' time of AGNI atmosphere objects. We can bring these back as required, when addressing #634

nichollsh:
> Flipping StopClock.enabled to True is a sensible global runaway guard; worth remembering that runtime resets per invocation, so it caps a single continuous run rather than a resumed campaign.

Good idea. We should implement this into the 'grid' and 'infer' tooling.

PR #678, opened 2026-05-13
Interior refactor: Aragog, Zalmoxis, atmodeller, dummy modules, testing overhaul, version pinning, oxygen accounting

timlichtenberg:
**This PR is ready for review.** It is the larger, integrating half of a coupled change with FormingWorlds/CALLIOPE#20 (chemistry). CALLIOPE#20 has merged and shipped as `fwl-calliope` 26.06.01, which this PR pins, so the dependency is satisfied; see the review guidance and timeline below.

**Review timeline.** I would like the review to take about a week. Please send feedback by **Friday 5 June**; I will incorporate comments and aim to **merge by Monday 8 June**.

Reviewers: @maraattia, @nichollsh, @rdc49, @planetmariana, @EmmaPostolec, @stuitje, @egpbos, @MarijnJ0, @Emeline0110 @IKisvardai. General comments are welcome, not just on the points I flag below.

**How to review.** This PR is too large to read line by line, and most of it is documentation and tests. Please do not try to read every diff. Instead, serve the docs locally (`zensical serve`, see `docs/How-to/documentation.md`) and follow a few of the tutorials end to end to confirm they work for you. For the implementation itself, please focus only on the specific places where I tag you in the text below; that is where I want eyes on the actual code.

**Paired PR and merge order.** This PR depends on FormingWorlds/CALLIOPE#20, the chemistry half of the oxygen accounting. CALLIOPE#20 has merged and is released as `fwl-calliope` 26.06.01; this PR pins that release (`fwl-calliope>=26.06.01`), so the dependency is satisfied and this PR is ready to merge on its own.

## Description

`tl/interior-refactor` has been in development for a few months. The change the branch is named for is the incorporation of two new Python interior backends, Aragog (thermal evolution) and Zalmoxis (interior structure), and the reorganisation of the interior into two clean axes around them. Built on top of that are a one-command automatic installer, central module version pinning, a new config format version, a documentation restructure, a test-framework overhaul, a full set of dummy module backends, atmodeller as a fully supported outgassing backend, and the #677 whole-planet oxygen accounting fix. The branch also merges in the ten PRs that landed on `main` in the meantime, so it stays current with `main` rather than diverging from it; that is catch-up, not new capability. Most of the change is the docs overhaul, the migrated config files, and the expanded test suite, not new physics.

### Interior backends: Aragog and Zalmoxis

[...]

Issue #436, opened 2025-06-25
Ready for HPC applications -> Snellius


timlichtenberg:
PROTUS currently runs reasonably well on ~hundreds of simultaneous CPUs on Habrok. However, performing larger grids (~thousands) or even more (for example for #424) is not yet stable enough and out of reach at present. In order to get this done, we need to test larger grids of simulations on Snellius and produce a scaling graph to test performance up to the maximum of the currently-allowed Snellius project. The documentations needs to be updated accordingly (https://fwl-proteus.readthedocs.io/en/latest/snellius_cluster_guide.html).

nichollsh:
When you say "scaling graph", what does this mean exactly? Maybe a plot of simulation-years vs number of cores?

timlichtenberg:
Yes, that's what I mean.

planetmariana:
Following our discussion today, I’ve noticed an issue when running grids on Snellius: only a small fraction of simulations initialize successfully. On Habrok, this occasionally happens for a few cases, usually when a specific node fails to initialize the simulation, in which case no folder is created for that run. On Snellius, however, this seems to happen on a much larger scale. For example, in a grid of 300 simulations, only about 12 cases initialize correctly and produce their corresponding output folders and logs. In addition, the successful cases on Snellius appear in a random order, whereas on Habrok they usually start running in numerical order. I checked my quota and still have plenty of space, so it does not seem to be related to storage limits. I’m flagging this here in case @egpbos encounters the same behavior.

```msastre@home1
  Quota                Limit                Usage               
  200.00 GiB           210.00 GiB           11.4708%        GiB        [===                           ]
  1000000 Inodes       1200000 Inodes       21.8873%        Inodes     [======                        ]


msastre@wstor_scratch1
  Quota                Limit                Usage               
  8.00 TiB             10.00 TiB            0.0000%         TiB        [                              ]
  3000000 Inodes       4000000 Inodes       0.0019%         Inodes     [                              ]`

egpbos:
[...]

Issue #29, opened 2023-05-04
Dynamic output folders to allow multiple instances of PROTEUS


nichollsh:
Currently PROTEUS outputs everything to `output/`. This prevents multiple instances of PROTEUS from being run simultaneously because each instance would overwrite and read another instance's data. It would be good to have a custom working directory to allow multiple instances of PROTEUS to be run simultaneously.

One issue with this is that VULCAN is inflexible with where it outputs files. I believe it is fixed to `VULCAN/output/`, so we will need to tell VULCAN to generate output files with unique file names which are unlikely to clash with other, and the copy them to the appropriate location.

PR #692, opened 2026-06-09
set output root path with an environment variable

egpbos:
## Description
By default, the output root path will still be the PROTEUS/output folder. This commit adds the option to change that root path by setting the PROTEUS_OUTPUT_PATH environment variable. I want to use this to run multiple PROTEUS instances at once (e.g. for #436). I know that this should in principle be done using the grid mechanism, but in my case I also need to add additional instrumentation to each run and I think this would be the quickest way to achieve this. It could also be useful in general for those that want to keep their PROTEUS directory clean of anything except code.

The only related issue I could find is #29, but the fix for that only added the named output subdirectories, it did not allow for changing the base path.

## Validation of changes
I did not run extensive tests, nor did I write the appropriate docs. This PR was meant in the first place to show and discuss the idea (and basic implementation). If it is appreciated, I can do the rest of the checklist below.

I did check whether this conflicts with #678 and it doesn't. I do see that in that PR my problem of running PROTEUS multiple times is also solved in a different way, with the auto output mode. That is also great (and I would use it now if #678 was done), but changing the root path is orthogonal and I think still useful to have on its own (e.g. as mentioned above for those that want to keep code and data separate).

## Checklist

- [ ] I have followed the [contributing guidelines](https://proteus-framework.org/PROTEUS/CONTRIBUTING.html#how-do-i-contribute)
- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my code

## Relevant people
@timlichtenberg @nichollsh

### DIFF
--- a/src/proteus/utils/coupler.py
+++ b/src/proteus/utils/coupler.py
@@ -960,6 +960,11 @@ def get_proteus_directories(outdir='_unset') -> dict[str, str]:
         Proteus directories dict
     """
     root_dir = get_proteus_dir()
+    if os.environ.get('PROTEUS_OUTPUT_PATH') is None:
+        output_root = os.path.join(root_dir, 'output')
+    else:
+        output_root = os.environ.get('PROTEUS_OUTPUT_PATH')
+        log.info(f'Output directory root from PROTEUS_OUTPUT_PATH environment variable: {output_root}')
 
     return {
         'proteus': root_dir,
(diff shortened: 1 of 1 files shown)


timlichtenberg:
Good additional functionality imo, happy for you to develop the appropriate tests and documentation and then we can merge it. :)

timlichtenberg:
I am closing this PR as we want to avoid having PRs sitting in the loop that are not actioned. Can be re-opened once the branch is actioned.

nichollsh:
Agreed with this approach. I recall that we once discussed having a "timer" on stale pull requests. Maybe worth re-visiting the idea down the line?

PR #698, opened 2026-06-10
Run install-all and update-all from any directory

### LINKED ISSUE #693: update-all doesn't work outside of the PROTEUS directory
## Problem description
When you run `proteus update-all` outside of the PROTEUS directory, an error is raised.

## Expected behavior
I would expect the updater to run from anywhere the command is available.

## Evidence
The error message:
```
$ proteus update-all
Obtaining file:///MYHOMEDIR
ERROR: file:///MYHOMEDIR does not appear to be a Python project: neither 'setup.py' nor 'pyproject.toml' found.
  File "MYHOMEDIR/miniforge3/envs/proteus/bin/proteus", line 6, in <module>
    sys.exit(cli())
             ^^^^^
  File "MYHOMEDIR/miniforge3/envs/proteus/lib/python3.12/site-packages/click/core.py", line 1485, in __call__
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "MYHOMEDIR/miniforge3/envs/proteus/lib/python3.12/site-packages/click/core.py", line 1406, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "MYHOMEDIR/miniforge3/envs/proteus/lib/python3.12/site-packages/click/core.py", line 1873, in invoke
  File "MYHOMEDIR/miniforge3/envs/proteus/lib/python3.12/site-packages/click/core.py", line 1269, in invoke
  File "MYHOMEDIR/miniforge3/envs/proteus/lib/python3.12/site-packages/click/core.py", line 824, in invoke
    return callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "SOMEOTHERDIR/PROTEUS/src/proteus/cli.py", line 820, in update_all
    subprocess.run([sys.executable, '-m', 'pip', 'install', '-U', '-e', '.'], check=True)
  File "MYHOMEDIR/miniforge3/envs/proteus/lib/python3.12/subprocess.py", line 571, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['MYHOMEDIR/miniforge3/envs/proteus/bin/python', '-m', 'pip', 'install', '-U', '-e', '.']' returned non-zero exit status 1.
```

## My computer
macOS, Python 3.12

## Additional notes
The root cause is the `root` path that is set to `cwd()` here: https://github.com/FormingWorlds/PROTEUS/blob/1edab021b163e0de6f56c0839a7846fbea26ad5d/src/proteus/cli.py#L935. We could instead use `__file__` or make a nice `get_proteus_git_root` helper function (unless this is the only place it's ever used).

## Relevant people
@timlichtenberg @nichollsh

timlichtenberg:
## Description

`proteus update-all` crashed when run from outside the PROTEUS directory: it derived its root from the current working directory and ran `pip install -U -e .` there, so pip tried to install whatever folder the user was standing in. `install-all` carried the same dependency in quieter forms: it cloned AGNI into the working directory and invoked the SOCRATES install script by a relative path.

Both commands now resolve the source tree from the installed package location through a shared helper, so they work from any directory:

- The editable pip reinstall targets the resolved root instead of `.`
- The SOCRATES install script is invoked by its absolute path under the root
- The AGNI clone gets an explicit destination under the root
- The disk-space gate measures the filesystem the install actually writes to
- The default `--config-path` for the data download is anchored to the root; an explicitly passed relative path still resolves against the working directory
- A wheel install run from inside a checkout falls back to using that checkout; with no checkout reachable at all, the commands exit with a clear message instead of a pip traceback

Closes https://github.com/FormingWorlds/PROTEUS/issues/693 (the command now works from any directory).

## Validation of changes

- Reproduced the reported failure on the unmodified code from a foreign working directory (the editable pip target resolves to `.`) and verified the fixed code targets the source tree instead
- 11 existing `install-all` / `update-all` tests updated to pin the resolved root explicitly; 8 new tests cover the run-from-anywhere contract (pip target, script path, clone destination, disk gate, config anchoring, output directory) and the failure and fallback paths of the root resolution
- Full unit suite passes locally: 2362 passed, 7 skipped
- `bash tools/validate_test_structure.sh` and `python tools/check_test_quality.py --check` pass; ruff check and format clean

Test configuration: macOS (Apple Silicon) with Python 3.12.

## Checklist

### DIFF
--- a/src/proteus/cli.py
+++ b/src/proteus/cli.py
@@ -58,6 +58,7 @@ if _should_apply_deterministic(sys.argv, os.environ):
 import shutil  # noqa: E402
 import subprocess  # noqa: E402
 import tempfile  # noqa: E402
+import tomllib  # noqa: E402
 
 import click  # noqa: E402
 
@@ -781,6 +782,36 @@ def _update_input_data(config_path: Path):
         return False
 
 
+def _is_proteus_root(path: Path) -> bool:
+    """Return True when ``path`` holds the PROTEUS source tree.
+
+    Identity is checked by the project name in ``pyproject.toml``, so an
+    arbitrary Python project, a stale editable-install pointer, or a
+    directory that merely contains some ``pyproject.toml`` is never
+    accepted as the install target. Any filesystem error while probing
+    counts as "not a PROTEUS root".
+
+    Parameters
+    ----------
+    path : Path
+        Candidate directory.
+
+    Returns
+    -------
+    bool
+        Whether the directory holds the ``fwl-proteus`` project.
+    """
+    try:
+        pyproject = path / 'pyproject.toml'
+        if not pyproject.is_file():
+            return False
+        with open(pyproject, 'rb') as fh:
+            name = tomllib.load(fh).get('project', {}).get('name', '')
+    except (OSError, tomllib.TOMLDecodeError):
+        return False
+    return name == 'fwl-proteus'
+
+
 def _resolve_proteus_root() -> Path:
     """Return the PROTEUS source tree root, independent of the working directory.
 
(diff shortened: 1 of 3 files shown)


src/proteus/cli.py
```diff
@@ -932,9 +986,8 @@ def update_all(export_env: bool, config_path: Path):
         raise SystemExit(1)
     """Update SOCRATES, AGNI, and refresh PROTEUS environment."""
```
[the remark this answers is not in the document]
timlichtenberg: > Convert it to a comment or remove it to avoid confusion and potential linting of pointless string statements.

Removed in a5a87b24, together with the matching stray string in install-all.

src/proteus/cli.py
```diff
@@ -932,9 +986,8 @@ def update_all(export_env: bool, config_path: Path):
...
 
-    root = Path.cwd()
     # --- Step 1: update all Python packages ---
-    subprocess.run([sys.executable, '-m', 'pip', 'install', '-U', '-e', '.'], check=True)
+    subprocess.run([sys.executable, '-m', 'pip', 'install', '-U', '-e', str(root)], check=True)
 
     # --- Step 2: FWL_DATA check ---
     try:
```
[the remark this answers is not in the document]
timlichtenberg: > If update-all should fail when FWL_DATA is not available, check for directory existence (or missing env var) explicitly so users get the intended actionable error instead of silently continuing with a non-existent path.

Done in a5a87b24: the dead try/except is replaced by an explicit directory-existence check that aborts with the install-all hint, and a test pins the abort path.

nichollsh (approved):
I am happy to approve this PR, as a maintainer, in order to allow it to be merged. However, I would appreciate if @egpbos checked that it also resolves his problem since he raised the original issue.

timlichtenberg:
@egpbos Can you review this still today? If not I would suggest that we'll go ahead with the merge today.

timlichtenberg:
[pressed the wrong button] 🙈

src/proteus/cli.py
```diff
@@ -811,20 +904,17 @@ def install_all(export_env: bool, config_path: Path):
...
     click.secho(f'[+] FWL_DATA directory: {fwl_data}', fg='green')
 
     # --- Step 2: Install SOCRATES ---
-    root = Path.cwd()
     socrates_dir = root / 'socrates'
     if not socrates_dir.exists():
         click.secho('[+] Installing SOCRATES...', fg='blue')
         try:
-            subprocess.run(['bash', 'tools/get_socrates.sh'], check=True)
+            subprocess.run(['bash', str(root / 'tools' / 'get_socrates.sh')], check=True)
```
egpbos, line 917: Will get_socrates.sh handle running outside of the PROTEUS directory properly? Or will it get installed outside of PROTEUS root? If so, is that an issue (I reckon no, because of RAD_DIR)?

tests/test_cli.py
```diff
@@ -1390,10 +1390,36 @@ def _stub_disk_usage_low():
...
+    """Point FWL_DATA at an existing directory, as after an installation."""
+    fwl = tmp_path / 'fwl_data'
+    fwl.mkdir(exist_ok=True)
+    monkeypatch.setenv('FWL_DATA', str(fwl))
+
+
 @pytest.mark.unit
 def test_install_all_aborts_on_low_disk(monkeypatch, tmp_path):
     """``install-all`` exits non-zero when free disk is below the 5 GB threshold."""
+    _pin_proteus_root(monkeypatch, tmp_path)
```
egpbos, line 1416: Can this not be done (here and everywhere below) with just a fixture instead of a function call?

egpbos: Never mind, I see that below it actually replaces an existing function call.

egpbos (approved):
LGTM, thanks! Just one question below. I did not check the tests in detail.

One thing I did notice that's perhaps worth thinking about for a future pass: I see a lot of tests with multiple asserts, which leads to later asserts not running if an earlier one fails. This can make debugging a bit more cumbersome. In "hierarchical" cases it does make sense (e.g. you first check the type and if that is already wrong then checking ranges or even exact values doesn't make sense anymore), but when it's really multiple aspects that should independently be true then ideally one would create a separate test. This can be structured as a class with a setup part and then one method per assert. Anyway, of course this all only matters when tests actually fail, which we hope they never do ;)

src/proteus/cli.py
timlichtenberg: No issue here. `get_socrates.sh` resolves its own root from `$0`, not from the working directory (`root=$(dirname $(realpath $0))/..`, then `socpath="$root/socrates"`). Now that `install-all` passes the absolute script path, SOCRATES always installs into `<PROTEUS root>/socrates` regardless of where you run the command from, so it never lands outside the root. RAD_DIR would cover us either way, but the install path is anchored to the script location, not the cwd.

PR #702, opened 2026-06-11
Integrating petitRADTRANS with PROTEUS

### LINKED ISSUE #319: Postprocessing steps: comparisons to observations
The user can currently postprocess PROTEUS outputs (and grids of outputs) using the postprocess.jl scripts in the `tools/` folder. These recalculate the radiative transfer, which is useful for analysis of spectral features in emission. 

However, this process is quite hands-on and requires the user to run the scripts in the terminal themselves. It would be ideal if postprocessing were performed automatically according to a configuration option.

Ultimately, we could even use [petitRADTRANS](https://petitradtrans.readthedocs.io/en/latest/) to do this. With pRT we should be able to easily generate emission and transmission spectra at high resolution, by importing it as a Python library. 

- [x] Run offline chemistry via CLI (#365)
- [ ] Run offline chemistry automatically after simulation terminates (https://github.com/nichollsh/VULCAN/issues/1)
- [x] Calculate transmission/emission spectra (#339)
- [ ] Synthetic JWST (https://github.com/pcubillos/gen_tso)
- [ ] Synthetic LIFE (https://lifesim.readthedocs.io/en/latest/)

### LINKED ISSUE #652: PLATON installation fails, update needed
## Problem description
Installing PLATON using the script fails with an error caused by removal in setuptools of the `pkg_resources` module.

## Expected behavior
The failing call is to a script that should run an example in order to trigger downloading the 10GB of data files.

## Evidence
Running the script produces the following output (at the end):
```sh
[...]
Successfully installed configparser-7.2.0 corner-2.2.3 dynesty-3.0.0 emcee-3.1.6 future-1.0.0 nose-1.3.7 platon-6.3
  File "/Users/pbos/projects/interra/PROTEUS/platon//examples/transit_depth_example.py", line 4, in <module>
    from platon.transit_depth_calculator import TransitDepthCalculator
  File "/Users/pbos/projects/interra/PROTEUS/platon/platon/transit_depth_calculator.py", line 1, in <module>
    from pkg_resources import resource_filename
ModuleNotFoundError: No module named 'pkg_resources'
Done!!
```

## My computer
MacOS Tahoe 26.3 (Intel), Python 3.12.

## Additional notes
The issue is fixed upstream: https://github.com/ideasrule/platon/commit/93e127c5685d1190cb6ac80521a4335592e3e60f. Ideally, we'd merge that into the nichollsh fork. However, I see that some modifications have been made there and the fork is actually 10 commits behind upstream, so I don't know if this would be too much work.

A temporary workaround could be to pin setuptools: https://stackoverflow.com/a/79888812

## Relevant people
@nichollsh

rdc49:
This pull request integrates petitRADTRANS into the PROTEUS framework. In short, it takes the pressure, temperature, and gravity profiles from the atmosphere module as well as the atmospheric chemical composition from either the outgassing or the atmospheric chemistry modules and computes a transmission and an emission spectrum.

-There is now a 'get_prt.sh' script that clones the petitRADTRANS gitlab repository, installs petitRADTRANS and downloads the opacity data. It makes some small edits to the petitRADTRANS code related to MPI keys. It gets the names of the relevant opacity sources form the constants.py file.
-petitRADTRANS computes the transit radius using a reference radius, pressure and gravity and a hysrostatic equilibrium integrator. It assumes that the ideal gas law holds, the petitRADTRANS script sets these reference values using the midpoint in the atmospheric pressure, gravity and radius files, to avoid the effects of real gas behaviour in the deep atmosphere.
-The 'constants.py' file has been updated to include all the line, rayleigh and cia species that petitRADTRANS has access to. During installation, the opacity data for all of these species is downloaded. This is to avoid PROTEUS having to download opacity during its post-processing stage.

## Checklist

- [ ] I have followed the [contributing guidelines](https://proteus-framework.org/PROTEUS/CONTRIBUTING.html#how-do-i-contribute)
- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my code


tools/get_prt.sh
```diff
@@ -0,0 +1,64 @@
...
+
+# Download
+echo "Cloning from GitHub"
+if [ "$use_ssh" = true ]; then
+    uri="git@gitlab.com:mauricemolli/petitRADTRANS.git"
+else
+    uri="https://gitlab.com/mauricemolli/petitRADTRANS.git"
+fi
+echo "    $uri -> $workpath"
+git clone "$uri" "$workpath"
```
timlichtenberg, line 23: Since this script clones petitRADTRANS into `prt/` at install time, the `prt` entry committed at the repo root is redundant. It's currently a submodule gitlink (mode 160000, pointing at 0d9303cd) with no `.gitmodules`, so it's a dangling submodule reference, and it's also in `.gitignore`. Please drop it from the tree with `git rm --cached prt`.

tools/get_prt.sh
```diff
@@ -0,0 +1,64 @@
...
+
+# Change dir and install
+olddir=$(pwd)
+cd $workpath
+python -m pip install -U -e . --no-build-isolation
+
+sed -i '46c\    except Exception:' petitRADTRANS/__file_conversion.py
```
timlichtenberg: A few robustness points in this section:
- `use_ssh` is referenced above but never defined, so it always falls through to https. Worth defaulting it explicitly.
- These `sed -i '46c\...'` / `'30c\...'` edits patch petitRADTRANS by absolute line number, but the clone above is unpinned, so any upstream change shifts those lines and the sed silently rewrites the wrong one. Please pin the clone to a tag or commit and match on content instead of a line number.
- `sed -i` with no backup suffix is GNU-only; on macOS (BSD sed) it needs `sed -i ''`. PROTEUS supports macOS, so this will fail there.
- `PRT_INPUT_DATA_PATH=...` is assigned but never exported or used.

src/proteus/utils/coupler.py
```diff
@@ -369,11 +369,13 @@ def print_module_configuration(dirs: dict, config: Config, config_path: str):
...
-    if config.observe.synthesis == 'platon':
-        from platon import __version__ as platon_version
+    write = 'Observe module    %s' % config.observe.module
+    if config.observe.module == 'platon':
+        from platon import __version__ as obs_version
+    elif config.observe.module == 'petitRADTRANS':
+        from petitRADTRANS import __version__ as obs_version
 
-        write += ' version ' + platon_version
+    write += ' version ' + obs_version
```
timlichtenberg: `obs_version` is only assigned inside the platon and petitRADTRANS branches, so when `observe.module` is None (the default) this line raises `UnboundLocalError`. `print_module_configuration` runs on every start, so this breaks any run without synthetic observations. Either move the version append back inside the branches, or initialise `obs_version = None` and guard it.

src/proteus/config/_observe.py
```diff
@@ -14,24 +14,47 @@ class Platon:
...
-    synthesis: str
+    module: str
         Module to use for calculating synthetic spectra.
+    clip_vmr: float
+        Minimum VMR to include a species in radiative transfer.
     """
 
-    synthesis: str = field(
-        default='none', validator=in_((None, 'platon')), converter=none_if_none
+    module: str = field(
```
timlichtenberg: `module` no longer has a default, so it's now a required field. Configs that don't set `[observe] module` fail with `Observe.__init__() missing 1 required positional argument: 'module'`, which is what's failing the unit tests on dummy.toml and others. Please add `default='none'` so it stays backward-compatible, update the remaining input configs, and add a `synthesis -> module` rename to the config migration so existing v3 configs keep loading.

src/proteus/config/_observe.py
```diff
@@ -14,24 +14,47 @@ class Platon:
...
+        validator=in_((None, 'platon', 'petitRADTRANS')),
+        converter=none_if_none,
     )
+    clip_vmr: float = field(default=1e-8, validator=(gt(0), lt(1)))
```
timlichtenberg, line 60: Moving `clip_vmr` up to `Observe` and removing it from `Platon` leaves `platon.py` reading `config.observe.platon.clip_vmr` in two places (around lines 248 and 351), which no longer exists and will `AttributeError` on a real platon run. The mocked platon tests don't catch it because they mock the config. Please update platon.py to `config.observe.clip_vmr`.

src/proteus/observe/petitRADTRANS.py
```diff
@@ -0,0 +1,554 @@
...
+    return [gas for gas in gases if gas in petitRADTRANS_RAYLEIGH_SPECIES]
+
+
+def _get_supported_cia_species(gases: list[str], include_cia: bool) -> list[str]:
+    if not include_cia:
+        return []
+
+    available = []
+    for cia_species in petitRADTRANS_CIA_SPECIES:
+        components = cia_species.split('-')
```
timlichtenberg: The CIA species are double-dashed (e.g. `H2--H2`), so `split('-')` returns `['H2', '', 'H2']`. The empty-string component is never in `gases`, so `all(...)` is always False and every CIA pair gets dropped, meaning the continuum is silently never applied. Use `cia_species.split('--')`.

src/proteus/observe/petitRADTRANS.py
```diff
@@ -0,0 +1,554 @@
...
+        raise ValueError(
+            'Atmospheric composition has zero total mass fraction at one or more layers'
+        )
+
+    mass_fractions = {
+        gas: mass_contrib[i] / total_mass
+        for i, gas in enumerate(gases)
+        if gas not in petitRADTRANS_IGNORED_GASES
+    }
+    mean_molar_masses = total_mass / 1.0e-3  # pRT expects amu (g mol-1)
```
timlichtenberg, line 124: This sets the mean molar mass to the sum of x_i * M_i, which is only correct if the VMRs sum to 1. After `clip_vmr` drops trace gases and unsupported species are excluded, the sum is below 1 and the MMW is underestimated, which then feeds the hydrostatic scale height. The mass fractions just above are renormalised to 1, so the two are inconsistent. Normalise the VMRs (divide by their sum) before computing both.

src/proteus/observe/petitRADTRANS.py
```diff
@@ -0,0 +1,554 @@
...
+
+        # neglect trace gases
+        if np.amax(vmr) >= clip_vmr:
+            vmr_incl.append(vmr)
+            gas_incl.append(gas)
+
+    return gas_incl, vmr_incl
+
+
+def _construct_abundances(
```
timlichtenberg: `_construct_abundances` is never called in this module (the spectra use `_vmrs_to_mass_fractions` and `_interpolate_prt_profiles`). It looks carried over from platon.py. Please remove it.

src/proteus/observe/petitRADTRANS.py
```diff
@@ -0,0 +1,554 @@
...
+    return abundances
+
+
+def _get_ptr(atm: dict):
+    """
+    Returns the pressure, temperature and radius from the atmosphere data."""
+    prs = np.array(atm['pl'])  # Pa
+    tmp = np.array(atm['tmpl'])  # K
+    rad = np.array(atm['rl'])  # m
+    if prs[1] < prs[0]:
```
timlichtenberg: `_get_ptr` reverses `prs/tmp/rad` to ascending pressure here, but `_get_mix` returns the VMR arrays in the original `atm['pl']` order. They get paired in `_interpolate_prt_profiles` as `PchipInterpolator(log_prs, vmr)`. If a source delivers surface-first ordering so this reversal fires, the pressures and abundances are mismatched end-for-end and the spectrum is silently wrong; in the common TOA-first case it happens to be fine. Please reverse the VMRs together with the P/T/R triple, or sort all arrays in one place.

src/proteus/utils/constants.py
```diff
@@ -45,6 +45,13 @@
 # Supported gases
 vol_list = ['H2O', 'CO2', 'O2', 'H2', 'CH4', 'CO', 'N2', 'NH3', 'S2', 'SO2', 'H2S']
 vap_list = ['SiO', 'SiO2', 'MgO', 'FeO2']
+prt_gases = ["Al", "Al_+", "AlH", "AlO", "C2H2", "C2H4", "Ca_+", "CaH", "CH4", "CO",\
```
timlichtenberg: The backslash continuations aren't needed inside the brackets (implicit line continuation already applies), and the double quotes diverge from the single-quote convention used elsewhere in this file. This is what fails `ruff format --check` (along with `_observe.py`); a `ruff format` pass fixes both.

timlichtenberg (changes requested):
Thanks for putting this together, this is a genuinely useful capability and the structure of the synthesis module reads well. I'm requesting changes, though: CI is currently red across the board (ruff, both unit runners, both codecov gates), and a few issues go beyond what CI shows, including a couple of silent physics ones. Specific notes are inline; grouping them here.

**Blocking**
- `prt` is committed as a dangling submodule gitlink (mode 160000, no `.gitmodules`). Since `get_prt.sh` clones petitRADTRANS at install, it should be removed: `git rm --cached prt`.
- `print_module_configuration` raises `UnboundLocalError` when `observe.module` is None, which is the default, so it breaks normal runs.
- `observe.module` lost its default and is now a required field. That's what fails the unit tests on configs that don't set it. It needs a default plus a `synthesis -> module` config migration.
- CIA continuum is silently always empty because of `split('-')` on the double-dashed species names.
- Moving `clip_vmr` up to `Observe` breaks the existing platon path, which still reads `config.observe.platon.clip_vmr`.
- No tests for a 554-line module under `observe/`, which is on the physics-invariant list, and coverage drops below the 80% gate. Please add unit tests with a `physics_invariant` marker and at least one reference-pinned spectrum (a transit depth pinned against a published pRT result or an analytic limit).

**Worth a look (silent correctness)**
- The VMR arrays aren't reversed together with the P/T/R profile, so a surface-first source would pair abundances and pressures end-for-end.
- The mean molar mass assumes the VMRs sum to 1; after clipping they don't, so it comes out underestimated.

**Minor**
- `get_prt.sh` patches unpinned third-party source by absolute line number and uses GNU-only `sed -i` (fails on macOS), plus a couple of unused or undefined shell vars.
- `_construct_abundances` is dead code in the new module.
- `prt_rayleigh_species` lists He and H, which never appear in `prt_gases`, so only H2 can ever be used.
- The reference P/R/g uses the array index midpoint; worth confirming that's intended rather than a fixed reference pressure.
- A `ruff format` pass clears the style failures.

Happy to look again once CI is green. The physics structure is sound and most of these are small, targeted fixes.

nichollsh:
Thanks @rdc49, I'll review this today. As we discussed just now, I have imported the pRT repo from GitLab to the FWL GitHub page so that we have granular control: https://github.com/FormingWorlds/petitRADTRANS

src/proteus/config/_observe.py
```diff
@@ -14,24 +14,47 @@ class Platon:
     ----------
     downsample: int
         Downsample binning factor for the spectrum.
-    clip_vmr: float
-        Minimum VMR for a species to be included in the radiative transfer.
     """
 
     downsample: int = field(default=8, validator=ge(1))
-    clip_vmr: float = field(default=1e-8, validator=(gt(0), lt(1)))
```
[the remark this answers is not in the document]
nichollsh: This comment from copilot is not a problem since we are removing the PLATON backend.

src/proteus/config/_observe.py
nichollsh: Agreed w copilot. Set default to none.

src/proteus/config/_observe.py
nichollsh: Copilot's worry is not a problem if we are removing PLATON here.

src/proteus/config/_observe.py
```diff
@@ -14,24 +14,47 @@ class Platon:
...
+        Minimum VMR to include a species in radiative transfer.
     """
 
-    synthesis: str = field(
-        default='none', validator=in_((None, 'platon')), converter=none_if_none
+    module: str = field(
+        validator=in_((None, 'platon', 'petitRADTRANS')),
+        converter=none_if_none,
     )
+    clip_vmr: float = field(default=1e-8, validator=(gt(0), lt(1)))
```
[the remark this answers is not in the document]
nichollsh: This is a good suggestion from copilot

src/proteus/utils/constants.py
```diff
@@ -45,6 +45,13 @@
...
+             "CO2", "CrH", "Fe", "Fe_+", "FeH", "H2", "H2O", "H2S", "HCN", "K",\
+             "Li", "Mg", "Mg_+", "MgH", "MgO", "Na", "NaH", "NH3", "O", "O2", "O3",\
+             "OH", "PH3", "SH", "Si", "Si_+", "SiO", "SiO2", "SO2", "Ti", "Ti_+", "TiO", "V", "V_+", "VO"]
+prt_rayleigh_species = ['H2', 'He', 'H']
+prt_cia_species = ["CO2--CO2", "H2--H2", "H2--He", "H2O--H2O", "H2O--N2", "N2--H2", "N2--He", "N2--N2", "N2--O2", "O2--O2"]
```
nichollsh: These double dashes will cause problems, I think?

src/proteus/observe/petitRADTRANS.py
```diff
@@ -0,0 +1,554 @@
...
+    from proteus.atmos_clim.common import read_atmosphere_data
+
+    atm_arr = read_atmosphere_data(
+        outdir, [hf_row['Time']], extra_keys=['tmpl', 'pl', 'rl', 'gl', 'x_gas']
+    )
+
+    if (len(atm_arr) == 0) or (atm_arr[-1] is None):
+        log.warning(f"Could not read atmosphere data from '{outdir}'")
+        return None
+    return atm_arr[-1]
```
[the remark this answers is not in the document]
nichollsh: I suggest updating `read_ncdf_profile` to enable returning values for `gl`, since this would be useful functionality anyway.

src/proteus/observe/petitRADTRANS.py
```diff
@@ -0,0 +1,554 @@
...
+
+
+def _get_reference_prt_values(atm: dict) -> tuple[float, float, float]:
+    """Return the reference pressure [bar], radius [cm], and gravity [cgs] from the profile."""
+
+    mid_idx = len(atm['pl']) // 2
+    reference_pressure = float(np.asarray(atm['pl'], dtype=float)[mid_idx] / 1e5)
+    reference_radius = float(np.asarray(atm['rl'], dtype=float)[mid_idx] * 100.0)
+    reference_gravity = float(np.asarray(atm['gl'], dtype=float)[mid_idx] * 100.0)
+    return reference_pressure, reference_radius, reference_gravity
```
nichollsh: Why get the middle index? This seems like poorly constrained behaviour since the resultant reference pressure will be strongly tied to the number of layers.

Instead, it would be better to have the user provide an 'optimal' ref pressure in the config. The function would then find the index closest to this pressure. Use something like 10 bar as default.

src/proteus/observe/petitRADTRANS.py
nichollsh: Agreed

src/proteus/observe/petitRADTRANS.py
```diff
@@ -0,0 +1,554 @@
...
+
+
+def _copy_mass_fractions(
+    mass_fractions: dict[str, np.ndarray], gases_to_zero: list[str]
+) -> dict[str, np.ndarray]:
+    copied = {gas: np.array(values, copy=True) for gas, values in mass_fractions.items()}
+    for gas in gases_to_zero:
+        if gas in copied:
+            copied[gas] = np.zeros_like(copied[gas])
+    return copied
```
nichollsh: This function doesn't seem to be used anywhere?

src/proteus/observe/petitRADTRANS.py
```diff
@@ -0,0 +1,554 @@
...
+    X = [wl, de]
+    header = ''
+    header += str('Wavelength/um').ljust(14, ' ') + '\t'
+    header += str('None/ppm').ljust(14, ' ') + '\t'
+
+    # write file
+    X = np.array(X).T
+    log.debug('Writing transit depth spectrum')
+    np.savetxt(
+        get_transit_fpath(outdir, source, 'synthesis'),
```
nichollsh, line 548: Suggest using an output path of `observe/` rather than `synthesis/` to match the changes in the config.

src/proteus/observe/petitRADTRANS.py
```diff
@@ -0,0 +1,554 @@
...
+        star_radius=Rs_cm,
+        orbit_semi_major_axis=sep_cm,
+        frequencies_to_wavelengths=False,
+    )
+    stellar_flux = Radtrans.compute_star_spectrum(
+        star_effective_temperature=Ts,
+        orbit_semi_major_axis=sep_cm,
+        frequencies=np.array(wl_hz, dtype=float),
+        star_radius=Rs_cm,
+    )
```
nichollsh: Why not use the spectrum that PROTEUS already calculated? This would be more self-consistent than whatever pRT is doing here.

src/proteus/observe/petitRADTRANS.py
```diff
@@ -0,0 +1,554 @@
...
+    X = [wl, de]
+    header = ''
+    header += str('Wavelength/um').ljust(14, ' ') + '\t'
+    header += str('None/ppm').ljust(14, ' ') + '\t'
+
+    # write file
+    X = np.array(X).T
+    log.debug('Writing eclipse depth spectra')
+    np.savetxt(
+        get_eclipse_fpath(outdir, source, 'synthesis'),
```
nichollsh, line 715: As above: maybe use `observe/` rather than `synthesis/`.

src/proteus/observe/wrapper.py
```diff
@@ -28,10 +28,12 @@ def calc_synthetic_spectra(hf_row: dict, outdir: str, config: Config):
...
     """
 
-    if config.observe.synthesis == 'platon':
+    if config.observe.module == 'platon':
         from proteus.observe.platon import eclipse_depth, transit_depth
+    elif config.observe.module == 'petitRADTRANS':
+        from proteus.observe.petitRADTRANS import eclipse_depth, transit_depth
     else:
-        raise ValueError(f"Unknown synthesis module '{config.observe.synthesis}'")
+        raise ValueError(f"Unknown synthesis module '{config.observe.module}'")
```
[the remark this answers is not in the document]
nichollsh: Remove platon to avoid this

src/proteus/utils/coupler.py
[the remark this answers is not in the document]
nichollsh: Please address this

tools/get_prt.sh
```diff
@@ -0,0 +1,64 @@
...
+# Make room
+workpath="prt/"
+rm -rf $workpath
+
+# Download
+echo "Cloning from GitHub"
+if [ "$use_ssh" = true ]; then
+    uri="git@gitlab.com:mauricemolli/petitRADTRANS.git"
+else
+    uri="https://gitlab.com/mauricemolli/petitRADTRANS.git"
```
nichollsh:
```suggestion
    uri="git@github.com:FormingWorlds/petitRADTRANS.git"
else
    uri="https://github.com/FormingWorlds/petitRADTRANS.git"
```

tools/get_prt.sh
```diff
@@ -0,0 +1,64 @@
...
+cd $workpath
+python -m pip install -U -e . --no-build-isolation
+
+sed -i '46c\    except Exception:' petitRADTRANS/__file_conversion.py
+sed -i '30c\    except Exception:' petitRADTRANS/opacities/opacities.py
+
+echo "Installing petitRADTRANS opacity tables..."
+prt_input_data_path="$(pwd)/input_data"
+PRT_INPUT_DATA_PATH="$prt_input_data_path"
+rm -rf ~/.petitradtrans/petitradtrans_config_file.ini
```
nichollsh, line 60: This presumes I want the pRT data to be installed in my home folder. This should be set dynamically (esp. important on clusters with limited $HOME allocations, such as the IoA one).

I would suggest locating this in `$FWL_DATA/prt/` instead.

tools/get_prt.sh
```diff
@@ -0,0 +1,64 @@
...
+
+python - <<'PY'
+import os
+
+from petitRADTRANS.radtrans import Radtrans
+radtrans = Radtrans(line_species=['Ti_+'])
+PY
+
+sed -i "s|^prt_input_data_path = .*|prt_input_data_path = $(pwd)/input_data|" ~/.petitradtrans/petitradtrans_config_file.ini
```
nichollsh: Again, don't presume I want pRT data installed at `~/.petitradtrans`. I would suggest locating this in `$FWL_DATA/prt/` instead.

tools/get_prt.sh
```diff
@@ -0,0 +1,64 @@
...
+sed -i '/^\[Paths\]/i\
+opacities/lines/correlated_k/H2O/1H2-16O = 1H2-16O__HITEMP.R1000_0.1-250mu.ktable.petitRADTRANS.h5\
+opacities/lines/correlated_k/TiO/48Ti-16O = 48Ti-16O__McKemmish.R1000_0.1-250mu.ktable.petitRADTRANS.h5\
+opacities/lines/correlated_k/VO/51V-16O = 51V-16O__Plez.R1000_0.1-250mu.ktable.petitRADTRANS.h5\
+opacities/lines/correlated_k/Na/23Na = 23Na__Allard.R1000_0.1-250mu.ktable.petitRADTRANS.h5\
+opacities/lines/correlated_k/K/39K = 39K__Allard.R1000_0.1-250mu.ktable.petitRADTRANS.h5\
+opacities/lines/correlated_k/CH4/12C-1H4 = 12C-1H4__HITEMP.R1000_0.1-250mu.ktable.petitRADTRANS.h5\
+opacities/continuum/collision_induced_absorptions/H2--H2/H2--H2-NatAbund = H2--H2-NatAbund__BoRi.R831_0.6-250mu.ciatable.petitRADTRANS.h5\
+opacities/continuum/collision_induced_absorptions/H2--He/H2--He-NatAbund = H2--He-NatAbund__BoRi.DeltaWavenumber2_0.5-500mu.ciatable.petitRADTRANS.h5\
+' ~/.petitradtrans/petitradtrans_config_file.ini
```
nichollsh: Why these species?

prt
nichollsh, line 1: Remove this. We don't want to add pRT as a subproject. 

Once removed, add `prt` to the `.gitignore` file.

nichollsh (changes requested):
Thanks @rdc49 - this will be an important and useful addition to PROTEUS.

This is my first pass on this PR. I haven't tried running the pRT wrapper yet, but did follow the install script, which seems to generally work.

1. Perhaps this PR should also remove the PLATON backend, since we no longer plan to support it? This would then address #652.

2. The script should not install data files to `~/.petitradtrans` since it'll cause big problems for clusters with limited `$HOME` allocations. Instead, place these data in `$FWL_DATA/prt/` or something similar.

3. Add information into the documentation about how to get pRT using your script, and how to use it.

4. Add `*.h5` into the gitignore, since I can already forsee someone putting their massive pRT data files into the PROTEUS folder and uploading them to GitHub.

5. Other comments below.

timlichtenberg:
> Thanks @rdc49, I'll review this today. As we discussed just now, I have imported the pRT repo from GitLab to the FWL GitHub page so that we have granular control: https://github.com/FormingWorlds/petitRADTRANS

This is not a fork though because the original project sits on GitLab. How do we handle upstream updates?

nichollsh:
> This is not a fork though because the original project sits on GitLab. How do we handle upstream updates?

We have to do it manually using a local clone, but it's straightforward. The steps would be:

1. Clone FWL repo
2. Add original gitlab repo as an alternative remote ([docs](https://docs.github.com/en/get-started/git-basics/managing-remote-repositories))
3. Fetch from gitlab remote
4. Merge changes from gitlab remote master branch to github 'fork' master branch.

tools/get_prt.sh
rdc49: These species have multiple data files in the google keeper directory. In this case, the installation process prompts the user to select which option they prefer, which breaks the automated script. Setting these options manually in the config fixes this issue.

tools/get_prt.sh
rdc49: I think the second point can be addressed if we make these changes directly to our forked version of PRT.

tools/get_prt.sh
rdc49: PRT automatically creates the config file in the home directory, so I initially need to download some data into the prt folder in the home directory, which creates the config file. Then, I tell the config file to download data to a custom location. We could change this behaviour in our forked branch of prt?

tools/get_prt.sh
nichollsh: I see - makes sense. I think having 'default' options automatically selected is fine. Updating the config with `sed` is a good approach

tools/get_prt.sh
nichollsh: It's fine to have small files in the home directory. This means having a config file at `~/.petitradtrans` is okay. We should just ensure that large files (i.e >1 MB) are kept in the user-designated folder. Ideally, within `$FWL_DATA/prt/`.

rdc49:
Hi all, I've made changes to the branch in light of the issues flagged by the first review. There are some code coverage tests that are failing, but they're only 0.02% off the threshold, so hopefully they shouldn't be an issue.

nichollsh:
@rdc49 is this ready for another review, then? Please request one when ready.

src/proteus/atmos_clim/common.py
```diff
@@ -85,6 +85,8 @@ def read_ncdf_profile(nc_fpath: str, extra_keys: list = [], combine_edges: bool
     p = np.array(ds.variables['p'][:])
     pl = np.array(ds.variables['pl'][:])
 
+    g = np.array(ds.variables['gravity'][:])
+
     t = np.array(ds.variables['tmp'][:])
```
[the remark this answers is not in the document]
nichollsh: I don't like the suggestion from copilot of filling with NaN values here, since they will just propagate. @rdc49, instead, can you make this check for `gravity` being a valid key and, if it's not present, simply raise an error?

src/proteus/observe/petitRADTRANS.py
```diff
@@ -0,0 +1,575 @@
...
+    header += str('Wavelength/um').ljust(14, ' ') + '\t'
+    header += str('None/ppm').ljust(14, ' ') + '\t'
+
+    # write file
+    X = np.array(X).T
+    log.debug('Writing transit depth spectrum')
+    np.savetxt(
+        get_transit_fpath(outdir, source, 'observe'),
+        X,
+        delimiter='\t',
```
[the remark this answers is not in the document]
nichollsh: Agreed with this point by copilot - maybe just stick with "observe".

src/proteus/observe/petitRADTRANS.py
```diff
@@ -0,0 +1,575 @@
...
+    header += str('Wavelength/um').ljust(14, ' ') + '\t'
+    header += str('None/ppm').ljust(14, ' ') + '\t'
+
+    # write file
+    X = np.array(X).T
+    log.debug('Writing eclipse depth spectra')
+    np.savetxt(
+        get_eclipse_fpath(outdir, source, 'observe'),
+        X,
+        delimiter='\t',
```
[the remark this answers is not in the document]
nichollsh: As above, just stick with "observe" throughout the code and avoid the term "synthesis" since it's vague.

tools/get_prt.sh
```diff
@@ -0,0 +1,69 @@
...
+echo "    $uri -> $workpath"
+git clone "$uri" "$workpath"
+
+# Change dir and install
+olddir=$(pwd)
+cd $workpath
+python -m pip install -U -e . --no-build-isolation
+
+sed -i '0,/^    except ImportError:$/{s/^    except ImportError:$/    except Exception:/;}' petitRADTRANS/__file_conversion.py
+sed -i '0,/^    except ImportError:$/{s/^    except ImportError:$/    except Exception:/;}' petitRADTRANS/opacities/opacities.py
```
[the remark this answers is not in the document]
nichollsh: Please also implement this suggestion from copilot

docs/How-to/installation.md
```diff
@@ -446,13 +446,13 @@ written in Julia.
 bash tools/get_lovepy.sh
 ```
 
-### Synthetic observations (PLATON)
+### Synthetic observations (petitRADTRANS)
 
-[PLATON](https://platon.readthedocs.io/en/latest/intro.html) generates
+[PLATON](https://petitradtrans.readthedocs.io/en/latest/) generates
 synthetic transmission and secondary eclipse spectra.
```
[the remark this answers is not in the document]
nichollsh: Agreed - make sure the docs are consistent. It would be better to link to our GitHub fork here, actually.

tools/.nfs0000000001095c4100006b07
```diff
@@ -0,0 +1,69 @@
+#!/bin/bash
+# Download and install petitRADTRANS
+
+# Load shell-defined environment variables such as FWL_DATA.
+if [ -z "${FWL_DATA:-}" ] && [ -f "$HOME/.bashrc" ]; then
```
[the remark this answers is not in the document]
nichollsh: Yes, please remove these `.nfs*` files, and then add them to the `.gitignore`

nichollsh:
> but they're only 0.02% off the threshold,

The total coverage is decreased by this PR by -1% points, which is quite a lot actually. This is because the absolute coverage for the changed-lines in this PR is only 33.8%, which is too low. Please incorporate more tests to hit the required 80% patch coverage.

src/proteus/config/_observe.py
```diff
@@ -1,37 +1,52 @@
...
 
     Attributes
     ----------
-    downsample: int
-        Downsample binning factor for the spectrum.
-    clip_vmr: float
-        Minimum VMR for a species to be included in the radiative transfer.
+    input_data_path: str or None
+        Optional path to petitRADTRANS `input_data`. If `None`, the installed
+        package location will be used.
```
nichollsh: Is that right? I thought it looks for data based on the environment variable.

src/proteus/observe/petitRADTRANS.py
```diff
@@ -0,0 +1,575 @@
...
+if TYPE_CHECKING:
+    from proteus.config import Config
+
+log = logging.getLogger('fwl.' + __name__)
+
+petitRADTRANS_TLIMS = (100.0 + 0.5, 4000.0 - 0.5)
+petitRADTRANS_GASES = tuple(prt_gases)
+petitRADTRANS_RAYLEIGH_SPECIES = tuple(prt_rayleigh_species)
+petitRADTRANS_CIA_SPECIES = tuple(prt_cia_species)
+petitRADTRANS_IGNORED_GASES = set(prt_ignored_gases)
```
nichollsh: Why create these extra global variables and then only use them once? Why not just use the imported ones directly?

src/proteus/observe/petitRADTRANS.py
```diff
@@ -0,0 +1,575 @@
...
+FWL_DATA_DIR = Path(os.environ.get('FWL_DATA', platformdirs.user_data_dir('fwl_data')))
```
nichollsh: Probably better to get this using the `dirs` dictionary that's constructed when PROTEUS starts up.

tools/get_prt.sh
```diff
@@ -0,0 +1,69 @@
...
+opacities/lines/correlated_k/CH4/12C-1H4 = 12C-1H4__HITEMP.R1000_0.1-250mu.ktable.petitRADTRANS.h5\
+opacities/continuum/collision_induced_absorptions/H2--H2/H2--H2-NatAbund = H2--H2-NatAbund__BoRi.R831_0.6-250mu.ciatable.petitRADTRANS.h5\
+opacities/continuum/collision_induced_absorptions/H2--He/H2--He-NatAbund = H2--He-NatAbund__BoRi.DeltaWavenumber2_0.5-500mu.ciatable.petitRADTRANS.h5\
+' ~/.petitradtrans/petitradtrans_config_file.ini
+
+python - <<'PY'
+from petitRADTRANS.radtrans import Radtrans
+from proteus.utils.constants import prt_cia_species, prt_gases, prt_rayleigh_species
+
+radtrans = Radtrans(line_species=prt_gases, rayleigh_species=prt_rayleigh_species, gas_continuum_contributors=prt_cia_species)
```
nichollsh: This line triggers pRT to download all the gases in the array `prt_gases`. This is a lot of species. I would instead suggest:
* H2O
* H2
* CO2
* CO
* CH4
* SO2
* H2S
* O2
* N2
* NH3
* OH
* ... plus CIA combinations thereof

And then neglect all the refractories for now. At the moment, their `.h5` files are now just taking up space on my laptop but won't appear in the spectra.

rdc49:
It still says that the code coverage is less than 90%, but I added tests that address the lines flagged by codecov, so I'm not sure why the coverage is under 90%. Otherwise, the branch is ready for review.

nichollsh (changes requested):
Thanks, @rdc49. This is great. I have tested running PROTEUS with AGNI, and then it setting the config to automatically post-process with pRT. Works after getting through a few hours of pRT install difficulties (see below).

Overall this is really exciting to see! We can probably merge this with only a ~few more hours work on it. Please request a review when you're ready.

1. I have a suggestion for improving the functionality (see below).

2. You probably should also include pRT in one of the slow/nightly integration tests. At the moment, the integration test config files (in `tests/integration/*.toml`) are misformatted since they use the old syntax. This is one of the reason the integration tests are failing when I run them locally, but they are skipped by GitHub in the PR checks. Please incorporate pRT into the integration tests, or at least update the configs to use the new config-parser syntax.

3. Please include an example in the docs of what a PROTEUS run with pRT observations is expected to produce. Something like the plot below, but ideally after handling point (1) above.

Image

## Loop over gases

Copied from my message to you on Discord...

> It would be really useful if you did the loop over the different gases, removing one gas at a time from the pRT calculation, so we can see how the spectrum looks with/without each one included. This then tells us which is the origin of each molecular feature. I had this functionality in the PLATON implementation so it should be easy to restore it for pRT. Note that the output CSV file contains a column None/ppm which indicates no gases removed. You'd then have columns for each gas (e.g. H2O/ppm) which represents the whole transmission spectrum with all gases included except for H2O. So differences between the None/ppm and H2O/ppm depths are attributable to H2O absorption. 


## Installation problems

### Remove `sed` calls in setup script

I had a number of issues installing pRT with your script. However, when doing the main steps manually in a fresh conda environment, it seems to work fine. 

To get it working, I skipped the `sed` commands, which seem to be the main source of my problems. Why are they necessary? Does pRT not work without these for you?

### pRT install hangs

When installing pRT on the Groningen Habrok cluster, I get the following message and then it hangs forever

```
interactive2:pid1426166.python: Failed to get ens3 (unit 0) cpu set
interactive2:pid1426166.python: Failed to get ens4 (unit 1) cpu set
interactive2:pid1426166.python: Failed to get ens5 (unit 2) cpu set
interactive2:pid1426166.python: Failed to get ens6 (unit 3) cpu set
interactive2:pid1426166: PSM3 can't open nic unit: -1 (err=23)
interactive2:pid1426166.python: Failed to get ens3 (unit 0) cpu set
interactive2:pid1426166: PSM3 can't open nic unit: 0 (err=23)
```

_Can ignore this for now_, but will need to re-visit later and ensure that PROTEUS/pRT works on clusters reliably.

### Data download error

I get a strange error when the `get_prt.sh` script attempts to download some of the opacity files:

```
OSError: Unable to synchronously open file (truncated file: eof = 17711104, sblock->base_addr = 0, stored_eof = 128730792)
```

This seems to only happen on my laptop and would not appear to be a problem with your code here. I created a fresh conda environment and it seemed to resolve the problem. Just flagging for posterity. _Can ignore this for now_.

src/proteus/config/_observe.py
nichollsh: I left this set to "None" and it worked fine. Is this even a necessary config option?

tests/integration/TEST_SUMMARY.md
nichollsh, line 1: Please remove this file. It is tailored to your specific pRT and PROTEUS configuration. The test infrastructure is also described in the docs, [here](https://github.com/FormingWorlds/PROTEUS/blob/main/docs/Explanations/test_framework.md), rather than in loose markdown files.

src/proteus/observe/petitRADTRANS.py
```diff
@@ -415,43 +459,90 @@ def transit_depth(hf_row: dict, outdir: str, config: Config, source: str):
...
+
+            line_removed = _get_supported_line_species(gases_removed, input_data_path)
+            rayleigh_removed = _get_supported_rayleigh_species(gases_removed, include_rayleigh)
+            cia_removed = _get_supported_cia_species(gases_removed, include_cia)
+
+            wl_removed, depth_removed = _compute_transit_column(
+                gases_removed,
+                vmrs_removed,
+                line_removed,
+                rayleigh_removed,
```
nichollsh, line 532: Thanks for adding this functionality that I requested. This is great.

tools/get_prt.sh
```diff
@@ -27,8 +27,26 @@ olddir=$(pwd)
...
+needle = '    except ImportError:\n'
+replacement = '    except Exception:\n'
+
+for path in targets:
+    text = path.read_text(encoding='utf-8')
+    shutil.copy2(path, Path(str(path) + '.bak'))
+    idx = text.find(needle)
+    if idx != -1:
+        text = text[:idx] + replacement + text[idx + len(needle) :]
+    path.write_text(text, encoding='utf-8')
```
nichollsh, line 48: This is now much clearer. Thanks for changing this.

src/proteus/plot/cpl_spectra.py
nichollsh: Update the file name to indicate the data-source for the spectra. Currently, it's not clear whether these come from `profile`, `offchem`, etc.

nichollsh (changes requested):
Thanks again, @rdc49. I have a couple minor comments but otherwise this is looking to be in good shape!

* I have tested running the `proteus observe` command and it works as described. 
* I also ran `proteus offchem` and then `proteus observe`, to test the processing pipeline, and it also works well.
* It's unclear to me why some of the spectral features appear with transit depths above the "None" case. This would imply that CO2 has negative absorption? Please double check the way that they are removed from the pRT radiative transfer.
* See in-line comments below

Spectrum from _offchem_:

Image

Spectrum from _profile_ (e.g. differences in H2O at 800 nm, CO and 4 um):

Image

nichollsh:
@rdc49 seems like the 'observe' unit tests are now failing. However, I believe that they can be fixed with some minor changes. Please let me know when this is ready for a new review. Seems very very close to being ready for approval

rdc49:
The tests have all passed now, so the branch should be ready to merge.

nichollsh (approved):

nichollsh:
@rdc49 I fixed some final merge conflicts in the docs, and have now merged the branch into main.

PR #703, opened 2026-06-11
Fix spurious Codecov failures on PRs and keep the main baseline fresh

timlichtenberg:
## Description

Pull requests have been showing red `codecov/patch` and `codecov/project` checks even when their own coverage is fine. This fixes the cause and keeps Codecov's `main` baseline current.

There are two problems, addressed together:

1. The pull-request job restores the most recent nightly coverage report into `nightly-coverage/` (it needs it for the diff-cover union), then uploads to Codecov. `codecov-action` searches the workspace by default, so it also picked up that nightly `coverage.xml` and uploaded it under the `unit-tests` flag. The nightly report is aligned to an older commit, so its line numbers no longer match the PR head; Codecov overlaid them and mismarked covered lines as missing. That is what produced the spurious patch/project failures.

2. Codecov uses the latest `main` coverage as the PR comparison base, but the pull-request workflow only uploads coverage on `pull_request` events, so commits landing on `main` never refresh the base.

## Changes

- `.github/workflows/ci-pr-checks.yml`: set `disable_search: true` on the Codecov upload so only the head `coverage.xml` is sent under the `unit-tests` flag, not the stale nightly report restored for diff-cover. This is the fix for the spurious PR coverage failures.
- `.github/workflows/coverage-baseline.yml`: a workflow that runs the unit suite with coverage on every push to `main` (and on manual dispatch) and uploads it under the `unit-tests` flag, keeping the base current. It is hardened so it cannot silently stop refreshing the base:
  - the upload runs even when a unit test fails on `main`, so the base keeps advancing; a real test failure surfaces as a red status on a final step instead;
  - every push runs to completion (`cancel-in-progress: false`), so a rapid second merge cannot cancel the producer and leave the base on an older commit;
  - a missing `CODECOV_TOKEN` skips the upload visibly rather than running a tokenless no-op that stays green;
  - it uploads only the named `coverage.xml` (`disable_search: true`).

## Validation of changes

- Confirmed from a pull-request CI run log that the Codecov upload reported "Found 10 coverage files" including `nightly-coverage/coverage.xml` and an AGNI Julia file, all sent under the `unit-tests` flag, which is the source of the corrupted line mapping. `disable_search: true` restricts the upload to the single named file.
- Verified the baseline job reuses the same `setup-proteus` action and the same pytest selection and coverage flags as the pull-request unit job, so its report is directly comparable.
- Validated both workflow YAML files parse.
- The push trigger can only be exercised once this lands on `main`; `workflow_dispatch` is included so the baseline can also be refreshed manually.

Test configuration: macOS (Apple Silicon) with Python 3.12.

## Checklist


nichollsh (approved):
Good catch! Thanks for looking into this. Seems like we are now converging upon a codecov and PR status configuration that works reliably.

PR #704, opened 2026-06-13
Align dummy run status reference with termination priority order

timlichtenberg:
## Description

The model's termination criteria are evaluated in priority order, and the first satisfied criterion sets the run's exit status. The maximum-time criterion is checked ahead of the physical-completion criteria, so a run that reaches its configured target time on the same step that another criterion (for example volatile escape) also fires is recorded as "target time" (status 13). This keeps an age-targeted run labelled by the age it reached, rather than being relabelled and dropped from a sample.

The all-dummy integration run reaches its 10 Gyr ceiling on the same step its atmosphere escapes, so its expected end state is "target time" (13). Its committed reference still held the older "volatiles escaped" (15), which is what broke the nightly `test_dummy_run` check. This updates the reference to match the priority order, documents that order and the strict/non-strict exit logic in `check_termination`, and adds unit tests that pin the behaviour.

## Validation of changes

macOS, Python 3.12.

- `tests/integration/test_integration_dummy.py` passes (4/4); the run writes status 13, matching the updated reference.
- `tests/utils/test_terminate.py` passes (26/26), including tests that pin target time over escape and over solidification, and energy balance over escape.
- `tests/utils/test_helper.py` status-string tests pass.
- A full nightly run dispatched manually on this branch confirms the complete slow tier before merge.

## Changes

- Update `tests/data/integration/dummy/status` to the run's actual end state, "target time" (13).
- Document the termination-criteria priority order and the strict/non-strict exit logic in `check_termination`.
- Add unit tests pinning the criteria priority and the maximum-clock-runtime status string.

## Checklist

### DIFF
--- a/src/proteus/utils/terminate.py
+++ b/src/proteus/utils/terminate.py
@@ -218,66 +218,46 @@ def _check_keepalive(handler: Proteus) -> bool:
 
 
 def check_termination(handler: Proteus) -> bool:
-    """
-    Decide if the model should stop or not.
+    """**Decide if the model should stop or not.**
 
-    The model will stop when two sequential iterations satisfy the convergence criteria.
-    This occurs when finished_prev = True  and  finished_both = True.
+    This function checks numerical and physical criteria for model termination.
+    It is called at the end of each iteration, and returns a boolean
+    indicating whether the model should terminate.
 
-    Parameters:
+    The criteria are evaluated in priority order, and the first criterion that
+    is satisfied will trigger termination.
+
+    The check can be performed 'strictly' or 'non-strictly'.
+    In strict mode, the model must satisfy the termination criteria for two
+    consecutive iterations before it will terminate.
+
+    Parameters
     --------------
-        handler: Proteus
-            Proteus instance
+    - handler: Proteus
+        Proteus instance
 
-    Returns:
+    Returns
     --------------
-        fininshed: bool
-            Should the model terminate now?
+    - finished: bool
+        Should the model terminate now?
     """
 
     # Quantify model state at this iteration
     finished = False
     handler.finished_both = False
 
-    # Check if keepalive file has been removed
-    #    This means that the model should exit ASAP, regardless of the other criteria.
+    # ------------------------
+    # 1) Should exit immediately if keepalive file is removed,
+    #    regardless of the other criteria.
+
     if _check_keepalive(handler):
         handler.finished_both = True
         handler.finished_prev = True
         return True
 
-    # Criteria are evaluated in priority order. They are combined with a
-    # short-circuiting ``or``, so the first criterion that fires writes the
-    # status file and suppresses the checks below it. Physical completion
-    # reasons (solidified, energy balance, escaped, disintegrated) are
-    # therefore checked before resource-budget limits (maximum time, loops,
-    # clock): if a run both finishes escaping and reaches its time ceiling on
-    # the same iteration, the reported reason is the physical one rather than
-    # "target time".
-
-    # Stop simulation when planet is completely solidified
-    if handler.config.params.stop.solid.enabled:
-        finished = finished or _check_solid(handler)
-
-    # Stop simulation when at global energy balance
-    if handler.config.params.stop.radeqm.enabled:
-        finished = finished or _check_radeqm(handler)
-
-    # Atmosphere has escaped
-    if handler.config.params.stop.escape.enabled:
-        finished = finished or _check_escape(handler)
-
-    # Planet has disintegrated
-    if handler.config.params.stop.disint.enabled:
-        if handler.config.params.stop.disint.roche_enabled:
-            finished = finished or _check_separation(handler)
-
-        if handler.config.params.stop.disint.spin_enabled:
-            finished = finished or _check_spinrate(handler)
-
-    # Maximum time reached
-    if handler.config.params.stop.time.enabled:
-        finished = finished or _check_maxtime(handler)
+    # ------------------------
+    # 2) Check resource-based criteria, set by user according to the
+    #    available the machine / job. These are not physical criteria.
 
     # Maximum loops reached
     if handler.config.params.stop.iters.enabled:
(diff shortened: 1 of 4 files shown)


src/proteus/utils/terminate.py
```diff
@@ -246,9 +246,14 @@ def check_termination(handler: Proteus) -> bool:
...
         return True
 
-    # Maximum time reached
-    if handler.config.params.stop.time.enabled:
-        finished = finished or _check_maxtime(handler)
+    # Criteria are evaluated in priority order. They are combined with a
+    # short-circuiting ``or``, so the first criterion that fires writes the
+    # status file and suppresses the checks below it. Physical completion
+    # reasons (solidified, energy balance, escaped, disintegrated) are
+    # therefore checked before resource-budget limits (maximum time, loops,
```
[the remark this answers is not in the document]
nichollsh: Ignore this suggestion from copilot

tests/utils/test_terminate.py
```diff
@@ -329,6 +329,51 @@ def test_check_termination_strict_resets_if_condition_lost(monkeypatch, patch_st
...
+    assert terminate.check_termination(h) is True
+    codes = [code for _, code in patch_statusfile]
+    assert codes[-1] == 15
+    # Discrimination: max-time (13) is lower priority and is short-circuited,
+    # so it is never written, not merely overwritten.
+    assert 13 not in codes
+
+
+@pytest.mark.unit
+def test_solidification_outranks_max_time(monkeypatch, patch_statusfile):
```
nichollsh: This should be swapped. Time should outrank solidification.

tests/utils/test_terminate.py
```diff
@@ -329,6 +329,51 @@ def test_check_termination_strict_resets_if_condition_lost(monkeypatch, patch_st
     assert patch_statusfile[-1][1] == 13
 
 
+@pytest.mark.unit
+def test_physical_completion_outranks_max_time(monkeypatch, patch_statusfile):
```
nichollsh: Again, should be swapped.

nichollsh (changes requested):
I think there has been a misunderstanding here. The change in check-order for the termination criteria in my last PR was intentional. That this PR reverses those changes would be a mistake.

It makes more sense to check the "target time" physical criterion first (highest priority, after the max iteration/runtime checks) because it is more useful to compare against this one than the others. For example, when targeting a specific planet 'age', we can put this as the max time criterion. However, if the escape criterion (for example) were given priority, then we would erroneously exclude this simulation from analysis, even though it reached the intended time.

I agree we should include tests for regression and think about this carefully, but please do not undo the changes made in my last PR. We can update the tests accordingly, instead.

timlichtenberg:
> I think there has been a misunderstanding here. The change in check-order for the termination criteria in my last PR was intentional. That this PR reverses those changes would be a mistake.
> 
> It makes more sense to check the "target time" physical criterion first (highest priority, after the max iteration/runtime checks) because it is more useful to compare against this one than the others. For example, when targeting a specific planet 'age', we can put this as the max time criterion. However, if the escape criterion (for example) were given priority, then we would erroneously exclude this simulation from analysis, even though it reached the intended time.
> 
> I agree we should include tests for regression and think about this carefully, but please do not undo the changes made in my last PR. We can update the tests accordingly, instead.

Do you want to continue with this PR/branch then from here on? I think that would be more straightforward. Then I can check if I understand the logic better with your changes to the tests.

nichollsh:
@timlichtenberg I've re-ordered the termination criteria and updated the tests. I also added some substantial comments/docstrings to explain the logic in `check_termination()`. I also added another test for these behaviours.

Can you review this PR manually and let me know what you think? I can't request a review from you since you created it in the first instance.

timlichtenberg:
Thanks Harrison, you've convinced me on the ordering. Agreed that for an age-targeted run the max-time criterion should take precedence: otherwise a run that reaches the intended age but also trips escape would drop out of the sample, which is the opposite of what we want. Happy to drop my original physical-first framing, and the grouping plus docstring you added are a clear improvement.

Two things before this is ready to merge:

1. The nightly won't go green yet. The failing check is `test_dummy_run`, which byte-compares the run's `status` file against `tests/data/integration/dummy/status`. That reference still reads `15` (volatiles escaped), but under the new ordering the all-dummy run writes `13` (target time), since its atmosphere escapes on the same step it reaches the 10 Gyr ceiling. So the test still fails with the original error. The reference needs regenerating to `13`, which is the correct expected value under the new priority. I checked locally that with that one-line update the full dummy file passes (4/4) and the terminate unit tests stay green.

2. A question on the resource group: `maxiter` (12) and `clock` (11) sit above max-time and all the physical criteria. By the same argument you made for max-time, a run that escapes or solidifies on the exact step it hits the iteration cap would be labelled "max iterations" and dropped from an age-targeted sample. Should max-time, and arguably physical completion, sit above the pure compute-budget limits? Or did you intend the resource limits to always win? Low stakes since the coincidence is rare, but I want to make sure the grouping is deliberate rather than incidental.

nichollsh:
Thanks. I'll correct point (1) as you've suggested.

Good suggestion about point (2). Perhaps it would indeed make more sense, so I'll update this ordering such that the 'numerical' checks happen after the physical ones.

nichollsh:
I've now checked that the tests pass on my laptop:
- `pytest -k test_dummy_run`
- `pytest -k test_terminate`
- `pytest -k test_status`
- `pytest -m "unit and not slow"`

timlichtenberg:
Thanks Harrison, both changes look good to me. The dummy reference now matches the new ordering (`13 / Completed (target time)`), and moving the numerical checks below the physical ones resolves my second point. Locally the dummy integration file (4/4), the terminate unit tests (26/26), and the status tests all pass.

I've manually triggered the nightly on this branch to confirm the full slow tier is green rather than waiting for tonight's scheduled run: https://github.com/FormingWorlds/PROTEUS/actions/runs/27472971712. Once that comes back clean we can merge.

timlichtenberg:
Manual nightly came back green (all slow shards, including the two `misc` ones that were failing), so I've squash-merged this. Thanks for sorting out the ordering and the reference, Harrison.

PR #705, opened 2026-06-14
Add config tooling to check for orphan keys

### LINKED ISSUE #666: Ensure that all user-provided config options are meaningful and actioned
## Feature description
Currently, there's no functionality in PROTEUS to check that all config options are parsed+implemented correctly. I realised this could be a problem while working on PR #665.

For example, a user might write `rayleigh_enabled=True` in their config file and believe that they've enabled rayleigh scattering. However, the correct option is rayleigh=True. This could lead to substantial issues when running+interpreting simulation outputs.

This seems quite critical to address, in my opinion.

## Preferred solution
Warn the user when TOML file contains variables which are not mapped in the config parser routine. This is separate to having unused optional variables, which are fine to include.

## Relevant people
@timlichtenberg @egpbos @EmmaPostolec @planetmariana

nichollsh:
## Description

Currently, it is possible to include "orphan" keys in the config file which don't map to any thing in the Config object of PROTEUS.  This module provides helpers that recursively compare the raw dict, read from the PROTEUS toml-formatted config files, against the Config schema. 

E.g. if you were to pass `atmos_clim.agni.spectral_group="Honeyside"` it will do nothing to warn you, at the moment, but you could then write a paper saying you used these opacities.

These are important to identify because a user could set these with intent, but they
will be ignored if not mapped by the parser. I have been caught out by this before, so am making this PR to prevent it happening in the future.

Closes #666 

## Validation of changes

* add tooling in a new `validate.py` file which checks for orphaned keys
* add tests to check for this behaviour

Tested on Fedora 44, Python 3.13

## Checklist

### DIFF
--- a/auto/status
+++ b/auto/status
@@ -1,0 +1,2 @@
+20
+Error (generic case, or configuration issue)
--- a/auto/status
+++ b/auto/status
@@ -1,2 +1,0 @@
-20
-Error (generic case, or configuration issue)
--- a/src/proteus/config/__init__.py
+++ b/src/proteus/config/__init__.py
@@ -7,7 +7,7 @@ from pathlib import Path
 import cattrs
 
 from ._config import Config
-from .validate import check_for_unknown_keys
+from .orphans import check_config_orphan_free
 
 log = logging.getLogger('fwl.' + __name__)
 
@@ -41,10 +41,6 @@ def read_config_object(path: Path | str) -> Config:
             obj.escape.module,
         )
 
-        # Parsed config is now a Config object.
-        # Check for unknown keys in the raw dict (raises if any are found).
-        check_for_unknown_keys(cfg, outdir=obj.params.out.path)
-
         # Looks good! Return the structured config object.
         return obj
 
@@ -65,4 +61,4 @@ def read_config_object(path: Path | str) -> Config:
         ) from None
 
 
-__all__ = ['Config', 'read_config_object', 'read_config', 'check_for_unknown_keys']
+__all__ = ['Config', 'read_config_object', 'read_config', 'check_config_orphan_free']
--- a/src/proteus/config/__init__.py
+++ b/src/proteus/config/__init__.py
@@ -7,6 +7,7 @@ from pathlib import Path
 import cattrs
 
 from ._config import Config
+from .validate import check_for_unknown_keys
 
 log = logging.getLogger('fwl.' + __name__)
 
@@ -22,8 +23,13 @@ def read_config(path: Path | str) -> dict:
 
 def read_config_object(path: Path | str) -> Config:
     """Read and validate config into Config object."""
+
+    # Read config from TOML file in path as a raw dict.
     cfg = read_config(path)
+
+    # Attempt to structure config with cattrs.
     try:
+        # Structure the config
         obj = cattrs.structure(cfg, Config)
         log.debug(
             'Config structured: star.module=%s, interior_energetics.module=%s, '
@@ -34,7 +40,15 @@ def read_config_object(path: Path | str) -> Config:
             obj.atmos_clim.module,
             obj.escape.module,
         )
+
+        # Parsed config is now a Config object.
+        # Check for unknown keys in the raw dict (raises if any are found).
+        check_for_unknown_keys(cfg, outdir=obj.params.out.path)
+
+        # Looks good! Return the structured config object.
         return obj
+
+    # Catch validation exceptions
     except cattrs.errors.ClassValidationError as e:
         # Extract actionable error messages from the nested exception group
         messages = []
@@ -51,4 +65,4 @@ def read_config_object(path: Path | str) -> Config:
         ) from None
 
 
-__all__ = ['Config', 'read_config_object', 'read_config']
+__all__ = ['Config', 'read_config_object', 'read_config', 'check_for_unknown_keys']
--- a/src/proteus/config/orphans.py
+++ b/src/proteus/config/orphans.py
@@ -101,23 +101,20 @@ def _collect_orphan_keys(data: dict, cls: type, path: str = '') -> list[str]:
     return orphans
 
 
-def check_for_unknown_keys(raw_dict: dict, outdir: str | None = None) -> bool:
-    """Raise if *raw_dict* contains keys that Config schema doesn't define.
-
-    Raises a ValueError if one or more keys have no matching field.
+def check_config_orphan_free(raw_dict: dict, outdir: str | None = None) -> bool:
+    """Detect if *raw_dict* contains keys that Config schema doesn't define.
 
     Parameters
     ----------
     raw_dict:
-        Raw TOML dict as returned by :func:`tomllib.load`.
+        Raw TOML dict as returned by `tomllib.load`.
     outdir:
         Output directory path.
 
     Returns
     ------
     bool
-        True if looks good. False if unrecognised keys, but will raise first.
-
+        True if looks good. False if unrecognised keys are found.
     """
 
     # Identify orphaned keys in the raw_dict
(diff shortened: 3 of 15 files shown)


auto/status
```diff
@@ -0,0 +1,2 @@
+20
```
timlichtenberg: This file is committed by accident. Its contents are exactly what `check_for_unknown_keys` writes on the orphan path (status 20), so it came from a local validator run. Please `git rm auto/status`. The underlying cause is the status-write path; see my comment on `config/__init__.py`.

src/proteus/config/__init__.py
```diff
@@ -34,7 +40,15 @@ def read_config_object(path: Path | str) -> Config:
             obj.atmos_clim.module,
             obj.escape.module,
         )
+
+        # Parsed config is now a Config object.
+        # Check for unknown keys in the raw dict (raises if any are found).
+        check_for_unknown_keys(cfg, outdir=obj.params.out.path)
```
timlichtenberg: `obj.params.out.path` is the bare run name (default `"auto"`), not the resolved output directory. `UpdateStatusfile` runs `os.path.abspath` on it, so on an orphan error the status file is written to `<cwd>/auto/status` instead of `output/<name>/status`. That is what produced the committed `auto/status`. At this point in the flow the output dir is not resolved yet (`"auto"` is still the sentinel), so I would drop the status write from the validator and report status from the caller after `SetDirectories`. Worth noting too: this function is also called from `cli.py` and `grid/manage.py` and is otherwise side-effect-free, so writing files and creating dirs as a side effect of parsing is surprising.

src/proteus/config/validate.py
```diff
@@ -0,0 +1,153 @@
...
+
+    # List the keys
+    msg += '\nUnrecognised keys:\n' + '\n'.join(f'\t{key}' for key in orphans)
+    log.error(msg)
+
+    # Update the status file, if possible
+    if outdir is not None:
+        from proteus.utils.helper import UpdateStatusfile
+
+        UpdateStatusfile({'output': outdir}, 20)
```
timlichtenberg: This is the write that lands in the wrong directory. `outdir` here is `params.out.path` (a bare name like `"auto"`), and `UpdateStatusfile` does `os.path.abspath(dirs['output'])`, so it resolves against the current working directory rather than `output/`. I would remove the status write from this function and signal the error to the caller, which already owns the resolved `dirs['output']`.

src/proteus/config/orphans.py
```diff
@@ -0,0 +1,153 @@
...
+
+        UpdateStatusfile({'output': outdir}, 20)
+    else:
+        log.error('No `outdir` provided, so cannot update status file!')
+
+    # Raise error here
+    raise ValueError(msg)
+
+    # For completeness, but will raise before this is called.
+    return False
```
timlichtenberg, line 136: `return False` is unreachable; the `raise` above always fires. Drop it, and trim the docstring so it reflects that the function returns `True` or raises.

tests/config/test_validate.py
```diff
@@ -0,0 +1,311 @@
...
+    # that suppressed the key list would fail even though it still raised.
+    assert 'Unrecognised keys' in str(exc_info.value)
+
+
+def test_cfgvalid_check_for_unknown_keys_updates_status_when_outdir_provided(tmp_path):
+    """Status file is written with code 20 when outdir is supplied, on error."""
+    dirty = {**_MINIMAL_VALID, 'ORPHAN': 42}
+
+    with pytest.raises(ValueError):
+        check_for_unknown_keys(dirty, outdir=str(tmp_path))
```
timlichtenberg: This passes an absolute `outdir`, so `os.path.abspath` is a no-op and the test passes regardless of the relative-path issue above. To exercise the production path, `chdir` into `tmp_path`, pass a bare relative name like `outdir="auto"`, and assert where the file actually lands. That is the case that would have caught the committed `auto/status`.

timlichtenberg (changes requested):
Thanks for adding this. Orphan keys have caught me out before, so I'm glad to have a guard for them.

I ran the validator against every bundled TOML using the installed `Config` schema. Good news first: zero false positives across the full option surface. `input/all_options.toml` (which contains every valid key) and all of the standard, tutorial, and intercomp configs parse clean, which is strong evidence that the schema mapping is complete. The grid and inference files use a different top-level schema and do get flagged by the raw function, but they never reach `read_config_object` (grid reads the resolved `ref_config`, inference loads the `.infer.toml` separately), so no real workflow regresses.

Two things to fix before merge:

1. `auto/status` is committed by accident. Its contents are byte-for-byte what `check_for_unknown_keys` writes on the orphan path (status 20), so it was produced by a local validator run and slipped in. Please `git rm auto/status`.

2. The status-file write resolves to the wrong place, and it is what created that stray file. `read_config_object` passes `outdir=obj.params.out.path`, which is the bare run name (default `"auto"`, still an unresolved sentinel at parse time). `UpdateStatusfile` runs `os.path.abspath` on it, so the file lands at `<cwd>/auto/status` rather than the canonical `output/<name>/status` that the rest of the code builds in `SetDirectories`. I reproduced `<cwd>/auto/status` with identical contents in a sandbox. Beyond the wrong path, `read_config_object` is called from three sites and is otherwise a pure read/validate, so writing files and creating directories at parse time is a surprising side effect. Simplest fix: drop the status write from `check_for_unknown_keys` and let the caller report status once the directories are set up.

Smaller points are inline. Once the artifact and the status write are sorted out, this is good to go.

src/proteus/config/__init__.py
nichollsh: Agreed. Perhaps it does make more sense to just have the validator to validation, and write the status file elsewhere. I'll rework it this way instead.

src/proteus/config/validate.py
nichollsh: Resolved by change described above

auto/status
nichollsh: Resolved by change described below

src/proteus/config/orphans.py
nichollsh: No longer raises. Now just returns true/false and simply performs check. Write and raise done elsewhere.

tests/config/test_validate.py
nichollsh: Resolved by changes above

nichollsh:
Thanks for the constructive review, @timlichtenberg. I have adjusted the functionality in a few ways:

* consolidated phrasing to refer to 'orphans' rather than 'unknown keys' etc
* the main `check_config_orphan_free()` function now just returns true/false rather than raising
* handling the presence of orphan keys is now done in a more Pythonic fashion, separately to the check, in `proteus.py`
* updates the status file appropriately
* does not write mistaken `auto` file to output dir
* caught some issues that would arise on nightly integration runs (due to orphan keys in config files)
* fixed the failing tests

src/proteus/config/orphans.py
```diff
@@ -0,0 +1,138 @@
...
+            nested_cls = _extract_attrs_class(hints.get(key))
+
+            # Recursion on this function
+            if nested_cls is not None:
+                orphans.extend(_collect_orphan_keys(value, nested_cls, full_path))
+
+    return orphans
+
+
+def check_config_orphan_free(raw_dict: dict, outdir: str | None = None) -> bool:
```
timlichtenberg: Optional: `outdir` is now unused inside this function (the caller in `Proteus.__init__` owns the status write). I'd drop the parameter and its docstring entry so the signature reflects that this is a pure predicate.

src/proteus/proteus.py
```diff
@@ -46,6 +46,14 @@ def __init__(self, *, config_path: Path | str) -> None:
         self.directories: dict = None  # Directories dictionary
         self.init_directories()
 
+        # Check for orphan keys in the config
```
timlichtenberg, line 49: Two optional notes here:

1. `outdir=self.directories['output']` is now a no-op (the function ignores it); it can go once the parameter is removed.

2. Scope: moving the gate here means it covers `proteus start` but not `proteus grid`, which reads its base/ref config through `read_config_object` without the check. Per-point grid configs are written from a structured `Config` so they are orphan-free by construction, which means a typo in a grid *base* config is silently dropped and never flagged, the exact case this PR targets. Fine to leave as-is given it sidesteps the grid/infer different-schema issue, but worth a one-line acknowledgement if grid base configs are a concern.

timlichtenberg (dismissed):
Thanks for the rework, this addresses everything from my earlier pass and the result is cleaner than the first version.

The key change is the right one: `check_config_orphan_free` is now a pure predicate (no I/O, no directory creation), and the status-20 write plus the raise moved into `Proteus.__init__`, where the resolved `self.directories['output']` is available. That is why the status file now lands in `output/<name>/` instead of the working directory, and it removes the surprising side effect from config parsing. I re-checked the path resolution and ordering (`read_config_object` drops orphans silently, `init_directories` resolves the output dir, then the gate fires and writes status 20 before raising) and it is correct.

Also confirmed:
- `auto/status` is gone.
- Zero false positives across the full option surface (`all_options.toml` and every standard/tutorial/intercomp config parse clean).
- The three integration TOMLs are now orphan-free; nice to see the tool flush real dead keys (`author`, `condensation`) out of our own fixtures.
- The new `Proteus.__init__` tests exercise the real path and the status-file write, which is what the previous test could not do.

Approving. A couple of optional follow-ups inline, neither blocking.

src/proteus/config/orphans.py
nichollsh: Removed this parameter

src/proteus/proteus.py
nichollsh: Thanks. 

1. Removed this parameter.
2. I am leaving as-is, since the `proteus grid` functionality does still respect this behaviour, as it still depends on setting up a PROTEUS handler object when dispatching workers.

egpbos:
Very nice!

PR #711, opened 2026-06-16
Additional updates to BO scheme and stability for BL mantle model

nichollsh:
### Bayesopt scheme
* Now uses same acquisition function for initial sampling and main retrieval
* Support for LogPI acq function, which is analogous to 'standard' log-likelihood approach
* Can set number of samples equal to number of workers.
* More tests!

### Boundary layer model
* Improved stability by handling numerics when CBL is small
* Removed redundant parameter `phase_transition_width` since it duplicating behaviour

## Validation of changes

* More tests to cover changes to BO scheme and prevent regression
* Updated/fixed tests for BL model
* BO scheme works well for SN/SE/TR cases
* Tests pass on IoA cluster
* Tests pass on GitHub CI

## Checklist

### DIFF
--- a/.github/workflows/ci-pr-checks.yml
+++ b/.github/workflows/ci-pr-checks.yml
@@ -518,6 +518,12 @@ jobs:
           files: coverage.xml
           flags: unit-tests
           name: unit-tests-coverage
+          # Upload only the unit coverage.xml. The earlier diff-cover step
+          # restores the last nightly report into nightly-coverage/; without
+          # this, codecov-action also discovers and uploads that base-aligned
+          # report under the unit-tests flag, whose stale line numbers corrupt
+          # the head report and mismark covered lines as missing on the PR.
+          disable_search: true
           fail_ci_if_error: false
 
       - name: Upload HTML coverage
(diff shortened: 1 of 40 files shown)


nichollsh:
@rdc49 this is now ready for review. The tests were failing because of some minor syntax issues but I have resolved these. They won't impact the review.

rdc49:
Hi Harrison, these changes look good. My only question is: why define the minimum atmosphere mass and minimum boundary layer thickness in boundary.py, as opposed to having them as config parameters?

nichollsh:
Thanks, Robb. 

That's a good question. I defined these constants at the top of the file since they were previously just hard-coded deep within the file, and thought it best to consolidate them. I do think they should probably stay as constants in the file rather than config parameters, since they are very much numerical rather than physical.

What do you think? Would they be better suited as config options?

rdc49:
Thanks for clarifying Harrison. I'm not fussed about them being hard coded, so I'm happy for the branch to be merged as is. There are no potential sources of conflict with the edits I'm making the boundary module.

nichollsh:
Great to hear that it doesn't conflict - this was my main concern about this branch. Although overall this does raise the question about which parameters should be included in the config, and which (usually numerical) ones can be kept in the code. Maybe worth discussing in the next meeting.

Issue #712, opened 2026-06-18
Default `all_options.toml` is failing


nichollsh:
### Structure solve times-out (unclear how to fix)

The default config, `all_options.toml`, uses Aragog and Zalmoxis to solve for the interior structure using the PALEOS tables for mantle/core EOS. It seems to fail to obtain a solution for the Earth-sized planet. However, it's quite hard to tell what's going on because the logs have become very spammy and verbose.

### Oxygen dominated atmosphere (suggested solution below)

The model predicts an intially O2-dominated atmospheric composition. Maybe this physical, but it's not what we predict in CHILI or in most other simulations. I think it's because the initial surface temperature is very large (too large): 4238 K.

Since AGNI doesn't currently include O2 opacities, this means that the atmosphere iteration also fails to converge, because it's basically transparent.

## Expected behavior

The default config should run fine using the most 'reasonable' default options, without errors, for an Earth sized planet. It is also extremely slow to get past this structure-solve phase, and then slow to evolve the simulation.

## Evidence

I see multiple instances of:
```
Accepting timeout solution: mass error 2.9973% (threshold 2%), density_converged=True, pressure_converged=True.
```

Log file and config attached. 

[all_options.toml.txt](https://github.com/user-attachments/files/29098259/all_options.toml.txt)

[proteus_00.log](https://github.com/user-attachments/files/29098348/proteus_00.log)

## My computer

Python 3.12, Fedora 44, Julia 1.12.

Git hash for PROTEUS is `b7e68b02cfe940bf273f647ca85f34087ded82da`

Git hash for Aragog is `d051902b4e5f216ffcc1ac1799ff6c34eec8c289`.

Git hash for Zalmoxis is `c3781d7662641a117bcfe46f86a9ec5a7078a9c2`.

## Additional notes

I believe that @rdc noted similar issues recently.

## Relevant people

@timlichtenberg @maraattia

nichollsh:
If I instead set the initial-state parameters to the values below, we get much more reasonable behaviour. However, the initial structure solves are still pretty slow since it waits until "accepting the timeout solution".

```
temperature_mode          = "isentropic"
ini_entropy               = 3600.0
flux_guess                = 1e3 
```

The initial composition is then:
```
[ INFO  ]     CO2        = 67.77     bar (5.96e-01 VMR)
[ INFO  ]     SO2        = 21.25     bar (1.87e-01 VMR)
[...]

PR #713, opened 2026-06-18
Updates to Boundary Interior Module

rdc49:
This pull request contains updates to the boundary interior module in light of recent pull requests on the main branch. Here are the changes:

-There is now no longer a config check that prevents the boundary interior module being used alongside Zalmoxis.
-Various parameters in the boundary config were duplicated in the interior_energetics config. These parameters have now been removed from the boundary config: the module instead uses the parameters from the interior_energetics config.
-The boundary module was calculating the core radius using the core_frac parameter under the assumption that it was the radius fraction. This calculation has now been changed such that the core radius is calculated using either the core mass fraction or the core radius fraction as appropriate.
-The boundary module now uses the surface temperature defined in the planet config class.


nichollsh:
Thanks, @rdc49. Shall we come back to this after #702 is merged? Or would you rather do it first? Seems like they are mostly orthogonal changes.

rdc49:
> Thanks, @rdc49. Shall we come back to this after #702 is merged? Or would you rather do it first? Seems like they are mostly orthogonal changes.

The changes in either pull request are orthogonal to each other, so we can merge them in either order.

tests/interior_energetics/test_interior.py
```diff
@@ -705,14 +706,17 @@ def test_run_dummy_int_rf_depth_scaling():
...
     assert output['Phi_global'] == pytest.approx(0.5)
-    # RF_depth = Phi * (1 - R_core/R_int) = 0.5 * (1 - 0.5) = 0.25.
-    assert output['RF_depth'] == pytest.approx(0.5 * (1 - 0.5), rel=1e-10)
-    # Discrimination: the retired config.core_frac path would give
-    # 0.5 * (1 - 0.6) = 0.20, distinguishable from 0.25.
-    assert abs(output['RF_depth'] - 0.5 * (1 - 0.6)) > 0.02
-    # Boundedness: RF_depth is a normalised mantle fraction in [0, 1-R_core/R_int].
-    assert 0.0 <= output['RF_depth'] <= 0.5
-    assert output['RF_depth'] < output['Phi_global']
+    upper_mantle_radius = 0.5 * (hf_row['R_int'] + hf_row['R_core'])
```
nichollsh: If including this 'separation' of the mantle mass into the dry/wet parts, based here on the 50% radius shell, please at least make this number a config parameter.

src/proteus/interior_energetics/wrapper.py
```diff
@@ -1475,19 +1475,12 @@ def run_interior(
...
+        # every backend. For all backends T_surf shares the
         # T_magma budget.
         dT_delta_magma = config.interior_energetics.tmagma_atol
         dT_delta_magma += config.interior_energetics.tmagma_rtol * T_magma_prev
 
-        if config.interior_energetics.module == 'boundary':
-            dT_delta_surf = float(config.interior_energetics.boundary.Tsurf_event_change)
-        else:
-            dT_delta_surf = dT_delta_magma
+        dT_delta_surf = dT_delta_magma
```
nichollsh, line 1483: Why this change? Does it change the behaviour?

src/proteus/outgas/calliope.py
```diff
@@ -25,6 +25,26 @@
 mass_ocean = ocean_moles * molar_mass['H2']
 
 
+def _upper_mantle_mass(hf_row: dict) -> float:
```
nichollsh: Does making this change here mean that the dummy outgassing and atmodeller calculations won't respect the upper/lower mantle distinction?

nichollsh (changes requested):
Thanks for this, @rdc49. It's good that you caught this (somewhat interpretive) issue regarding the melt fraction. I also like that you've been able to consolidate some of the parameters. However, I have a few concerns about the changes in this PR.

1. There is a mix-up between the _kinematic_ viscosity $\nu$ and the _dynamic_ viscosity $\eta$. The former appears in the Rayleigh-number scaling law, but the latter is what's provided by the `const_log10visc` config parameter. We should make sure these are consistent since $\nu$ and $\eta$ generally differ by a factor of ~1e3.

2. It seems arbitrary to pick the 50% mantle radius as a reference point. Why 50%? It seems that leaving the calculation as it was before (taking the upper-mantle temperature to estimate the whole mantle melt fraction) might be simpler? How sensitive is the model to this 50% choice? Why not make it a config parameter, at least?

See in-line comments too.

nichollsh:
@rdc49 following our chat earlier, please let me know when this is ready for another review.

rdc49:
The branch is ready for review now.

.github/workflows/ci-pr-checks.yml
nichollsh, line 1: It should not be necessary to modify the workflows. Why is this change made and what does it do?

nichollsh (changes requested):
Overall looks good.

I ran the tests and they seem to pass on my laptop. I also tested manually changing the updated parameters, and it works as you'd expect it to in each case. Also allowed me to use Zalmoxis + boundary modules, although I find that Zalmoxis+PALEOS is very slow (see #712).

Not sure about the modification to the github workflow. Why was this modification was necessary?

.github/workflows/ci-pr-checks.yml
rdc49: I'm not entirely certain what the change is: the test_migrate_config_v2_to_v3.py test was failing, and copilot changed the workflow in response to this, which seemed to fix the issue.

nichollsh: Okay - please reverse this change, then, and we can see what happens on the nightly CI run. If it fails, we can address it then.

nichollsh:
Looks like there are still errors with the `migrate_config_v2_to_v3.py` script. Can you please resolve these? I think it's related to the moved/removed config options. Then, the tests should pass without having to update the GH workflow.

To resolve, these you probably just need to remove all instances of the following keys from all the `.toml` files in the PROTEUS repo (incl in the tests folder).

```
 E         Extra items in the right set:
E         'interior_energetics.boundary.Tsurf_event_change'
E         'interior_energetics.boundary.heat_fusion_silicate'
E         'interior_energetics.boundary.eta_solid_const'
E         'interior_energetics.boundary.T_p_0'
E         'interior_energetics.boundary.eta_constant'
E         'interior_energetics.boundary.rtol'
E         'interior_energetics.boundary.atol'
E         'interior_energetics.boundary.eta_melt_const'
```

nichollsh:
I've fixed the failing tests and added a few more. These check for the `core_frac_mode` option since various choices are now possible with the BL module when used as the `interior_struct` module.

The checks are failing because the Linux nightly reference is stale. I've run the tests locally and they pass. Merging now.

nichollsh (approved):

nichollsh:
@rdc49 ^

Issue #447, opened 2025-07-11
Track helium outgassing and escape


nichollsh:
Escape processes leave the potential for helium worlds.
https://iopscience.iop.org/article/10.1088/0004-637X/807/1/8/meta
https://arxiv.org/pdf/2503.05055

We should track helium in PROTEUS. This can be readily fitted into the current framework. A quick search shows a few papers which may have solubility data. We don't need to worry about opacity.

https://www.nature.com/articles/nature04215
https://doi.org/10.1016/S0012-821X(00)00215-6
https://doi.org/10.1016/0016-7037(86)90193-6

nichollsh:
## Helium worlds
This has become somewhat more relevant again in light of Colin Cherubim's discovery of He escape from a HZ super-Earth ([LHS 1140 b](https://en.wikipedia.org/wiki/LHS_1140_b?useskin=vector)) announced at Exoplanets 6. They used WINERED highres transit measurements to detect He absorption.

Obvious science question here: if such a planet has an atmosphere and exists in the HZ, is it actually habitable?

He-CH4 and He-H2 continua can generate enough opacity to solve for climate profiles.

## Noble gases as tracers
Also, noble gas (ratios) can trace devolatilisation more generally. This could be useful in the Solar System context.

https://www.nature.com/articles/s41586-022-04710-4

## To do

This is still not especially high priority, but worth considering for incorporation into the modelling used for Interra2027.

- [ ] Incorporate noble gas partitioning laws into PROTEUS' outgassing modules (CALLIOPE or Atmodeller?)
- [ ] Add support for these in PROTEUS
- [ ] Add support for these in AGNI
- [ ] Add support for these in Zephyrus


<img width="600px" alt="Image" src="https://github.com/user-attachments/assets/59a8c177-b4c0-455a-b7ab-ccafe5f0b574" />

@shorttle @timlichtenberg @EmmaPostolec

timlichtenberg:
Atmodeller can already do He, Ne, Ar, Kr, Xe using Jambon+1986. We can incorporate it into CALLIOPE. Can you deal with the necessary adjustments for AGNI? Or is it already doable in AGNI?

nichollsh:
Oh great! That's fantastic. Thanks.

We just need to incorporate support for this into PROTEUS then, e.g. in specifying their budget in the config and their abundance passing to the climate models. 

I think they are already supported in AGNI but adding them would be trivial, if not supported already. 

[...]

PR #706, opened 2026-06-14
Super-Earth structure and thermal evolution: interior coupling, molten initial conditions

timlichtenberg:
This branch backs the Super-Earth structure and thermal evolution paper. It couples the interior-structure and interior-energetics solvers into the production stack, reworks the fully molten initial condition, feeds the evolving interior temperature back into the structure solve, extends the energy budget, tracks atmospheric oxygen under buffered outgassing, and hardens the desiccation check and the restart path. The three paired module changes it depends on are merged and released (see Review and merge order), and this branch's version floors are set to those releases, so it is ready for review. The paper's production grid configurations and their generator are kept with the paper data, not in this repository. The production grids are still running, so the absolute energy balance and cooling timescales quoted in the paper are still being refined; that is a manuscript matter and does not affect the code under review.

## Resolves

**Resolves #712**: the default `all_options.toml` no longer fails. The molten `liquidus_super` initial condition replaces the over-hot adiabatic start that produced the O2-dominated atmosphere AGNI cannot converge, the coupling no longer livelocks, and `solver_max_steps` is raised so the atmodeller solve converges. The remaining structure-solve slowness is the JAX molten-initial-condition speedup tracked in #719.

**Resolves #720**: the elemental `H_kg_atm` reservoir-split over-count. Per-element reservoirs are rebuilt from a clean per-species sum, and the closure `atm + liq + sol == total` now holds to one per cent across the redox sweep, with a regression test in `tests/integration/test_integration_atmodeller_dummy.py`.

## Review and merge order

These four paired changes, all developed on branch `tl/superearth_structure`, are one coupled change set. The three module changes are merged and released:

1. **FormingWorlds/Zalmoxis#77**: merged, released as `fwl-zalmoxis` 26.7.3.
2. **FormingWorlds/atmodeller#1**: merged.
3. **FormingWorlds/aragog#20**: merged, released as `fwl-aragog` 26.7.4, which exposes the `step_dE_compression_J` field the interior-energetics code reads.

This PROTEUS change is the last of the set. Its `fwl-aragog` floor is set to `>=26.07.04` and its `fwl-zalmoxis` floor to `>=26.07.03`, so CI installs the released modules rather than the paired development branches, and the interior-energetics solves read `step_dE_compression_J` from released Aragog.

## What this does

[...]

PR #722, opened 2026-07-02
Track noble gas budgets through outgassing and mass balance

timlichtenberg:
## Description

This wires the noble gases He, Ne, Ar, Kr and Xe into PROTEUS, so a magma-ocean run can carry a noble gas inventory alongside the reactive C-H-O-N-S volatiles. It is the PROTEUS half of FormingWorlds/PROTEUS#447 (track noble gas partitioning and escape).

Both outgassing backends carry the noble gases. The atmodeller backend has shipped the Jambon, Weill and Braun (1986) noble gas solubility laws for a while, and this PR wires them through the PROTEUS atmodeller path. The CALLIOPE backend gains the same laws in FormingWorlds/CALLIOPE#23, and the two backends use the same calibration, so they agree on noble gas partitioning.

Each noble gas is opt-in per gas. A run with no noble budget is unchanged, and a gas contributes only when its inclusion flag in `[outgas.calliope]` is set and its budget in `[planet.elements]` is positive. A budget is set with a per-gas mode and value: `kg` (absolute), `ppmw` (relative to the volatile reservoir), or `solar` (a multiple of the protosolar noble gas to hydrogen mass ratio). The budget is resolved to element masses at IC (through the shared, backend-agnostic inventory routine) and passed to whichever backend is selected, which returns the atmospheric and dissolved masses. Those are tracked as element masses in the whole-planet mass balance and the escape reservoir.

The noble gases are tracked as elements, not yet as radiatively active gas species, so their atmospheric inventory is available as a mass but is not yet passed to the climate solver. That radiative coupling (the He-H2 and He-CH4 continua in AGNI) and mass or diffusion limited escape fractionation in ZEPHYRUS are the remaining pieces of #447, described in the issue.

### Dependency and CI

The IC step that turns a noble gas budget into element masses is a shared, backend-agnostic routine (`calc_target_elemental_inventories`), and it resolves the budget through CALLIOPE's `get_target_from_params`. So a full run with either backend needs a CALLIOPE that supports noble gases (FormingWorlds/CALLIOPE#23) to populate the noble element masses at IC. The released `fwl-calliope` predates that, so end-to-end noble gas support on released packages is unavailable for both backends until the `fwl-calliope` pin here is bumped (a follow-up commit once CALLIOPE#23 is released). The wiring on both sides is in place and tested; only the IC budget resolution is gated on the CALLIOPE release.

- The unit tests mock CALLIOPE and pass on the released version.
- The atmodeller wrapper is exercised end-to-end on CI (`tests/outgas/test_noble_gas_atmodeller_smoke.py`) by seeding the resolved element budget directly and running the real atmodeller solver, since atmodeller already ships the noble gas solubility laws.
- The CALLIOPE end-to-end smoke test (`tests/outgas/test_noble_gas_smoke.py`) detects whether the installed CALLIOPE has noble gas support and skips if not, so the PR gate stays green. It activates once CALLIOPE#23 is released and the pin is bumped.

### Changes

- `utils/constants.py`: the noble gases are added to `element_list`, and a `noble_solar_mass_ratio` table of protosolar X/H mass ratios is added.
- `config/_planet.py` and `config/_outgas.py`: per-gas budget modes (`He_mode`/`He_budget`, ...) and inclusion flags (`include_He`, ...), all defaulting off.
- `outgas/calliope.py`: the noble gas budget resolution (`_resolve_noble`, kg/ppmw/solar) and the wiring that passes the gases to CALLIOPE.
- `outgas/atmodeller.py`: the wiring that passes the pre-resolved noble gas element masses to atmodeller as species and mass constraints. This path also now takes per-species atmospheric mass from atmodeller's conserved total-mass output rather than a mole-fraction pressure formula, which was misweighting light species and not conserving the tracked inventory.
- `escape/wrapper.py`: noble gases are exempt from the major-volatile desiccation floor, so a trace noble inventory is not zeroed on the first escape step.
- Tests: unit tests for the budget resolution and wiring, capability-gated CALLIOPE and always-on atmodeller end-to-end smoke tests, a reference-pinned test for the protosolar ratios, and an escape-floor exemption test.
- Docs: the noble gas config fields in the planet config reference and the annotated `all_options.toml`.

## Validation of changes

macOS with Python 3.12.

- The full unit suite passes (2494+ tests), including the noble gas wiring, resolution, deferral, escape-floor, and reference-pinned tests.
- The atmodeller end-to-end smoke tests run the real solver and confirm a helium budget partitions between atmosphere and melt with the tracked helium mass closing against the budget to the solver tolerance (about 1e-6 relative), scaling with the budget, and staying absent when no noble budget is set. The atmospheric and dissolved reservoirs then sum to the tracked total exactly, since the wrapper takes the total from atmodeller's conserved output and derives the atmospheric share from it.
- The CALLIOPE end-to-end smoke test, run against a local CALLIOPE with noble gas support, confirms a ppmw budget becomes a kg target and closes mass through the real solver.
- A run with no noble budget is unchanged: the noble element masses stay zero through outgassing, the mass balance, and the escape distribution.
- The `solar` budget mode is an upper-bound reference (planetary bodies are noble-gas depleted relative to solar by orders of magnitude); this is documented in the planet config reference.

## Checklist


src/proteus/utils/constants.py
```diff
@@ -83,8 +83,16 @@
 vap_list = ['SiO']  # just one for now
 gas_list = vol_list + vap_list
```
nichollsh: Need to extend `gas_list` here with the `noble_gases` list.

src/proteus/outgas/calliope.py
```diff
@@ -121,9 +139,62 @@ def construct_options(dirs: dict, config: Config, hf_row: dict):
...
+        gas_kg = _resolve_noble(mode, budget, H_kg, M_reservoir, gas)
+        if gas_kg <= 0.0:
+            continue
+        solvevol_inp[f'{gas}_included'] = 1
+        solvevol_inp[f'{gas}_ppmw'] = 1e6 * gas_kg / M_mantle if M_mantle > 0 else 0.0
+
     return solvevol_inp
 
 
+def _resolve_noble(
```
nichollsh: Why is this extra function needed? Seems like it does the same thing as `resolve_element`?

src/proteus/outgas/calliope.py
```diff
@@ -204,6 +275,19 @@ def construct_guess(hf_row: dict, target: dict, mass_thresh: float) -> dict | No
         log.debug('    providing None, allowing CALLIOPE to guess')
         return None
 
+    # Noble gas partial pressures are not carried in the helpfile (the noble
```
nichollsh, line 251: Seems like an oversight, no? Why would we not track their partial pressures in the helpfile? It would be extremely valuable to record this information since (a) we might get NG dominated atmospheres and (b) they contribute to the total pressure.

src/proteus/outgas/wrapper.py
```diff
@@ -388,9 +394,13 @@ def run_outgassing(dirs: dict, config: Config, hf_row: dict):
...
 
     # calculate total atmosphere mass (from sum of volatile masses)
+    # M_atm is the total atmospheric mass, including the noble gases, so it
+    # stays consistent with the surface pressure and mean molar mass, which
+    # both count them. The crystallized-mantle escape debit reads M_atm, so a
+    # noble-free M_atm there would over-deplete the atmosphere.
     hf_row['M_atm'] = 0.0
-    for s in gas_list:
-        hf_row['M_atm'] += hf_row[s + '_kg_atm']
+    for s in list(gas_list) + list(noble_gases):
```
nichollsh: Again, this would be simplified if `noble_gases` simply featured in `gas_list` alongside the other species. Doesn't make sense to give them special treatment.

src/proteus/utils/coupler.py
```diff
@@ -544,8 +550,12 @@ def assert_mass_conservation(hf_row: dict, atol_frac: float = 1e-6) -> None:
 
     # Invariant 2: M_atm stays in sync with the per-species kg_atm fields it
     # is summed from. This guards against a future reordering that mutates a
-    # species kg_atm after M_atm is computed without refreshing M_atm.
-    summed = sum(float(hf_row.get(s + '_kg_atm', 0.0)) for s in gas_list)
+    # species kg_atm after M_atm is computed without refreshing M_atm. The
+    # noble gases are included because their atmospheric mass now counts toward
+    # M_atm (they contribute to the surface pressure and mean molar mass).
+    summed = sum(
+        float(hf_row.get(s + '_kg_atm', 0.0)) for s in list(gas_list) + list(noble_gases)
```
nichollsh: Similarly, this would be simpler if the NG were already in `gas_list`.

nichollsh (changes requested):
Thanks for working on this so promptly, @timlichtenberg.

I understand the motivation to (a) have noble gases disabled by default and (b) retain consistency with previous behaviours.

However, it seems that including NG could be a reasonable default behaviour in the future. There's no reason *not* to include them. Given this, why are they handled with such 'special treatment' throughout the code?

For example, why not include them in the main `gas_list` array? We are doing this for volatiles + vapours already, so why not keep `gas_list` complete in including all of the modelled species? This would also simplify the code substantially and avoid confusion in the future.

Because they are neglected from `gas_list`, most of the plots do not actually plot the NG output values (e.g. they do not appear in the 'plot_global.png' output file).

The gases are also missing pre-defined colours in the `plot.py` file's `_preset_colours` dict, which means they will all appear with the same colour (and thus be indistinguishable in the plots).

src/proteus/utils/constants.py
timlichtenberg: Done. gas_list is now vol_list + vap_list + noble_gases, so the noble gases are full gas species.

src/proteus/outgas/calliope.py
timlichtenberg: Agreed. I folded it into `_resolve_element`, which now also handles the 'solar' mode, and removed `_resolve_noble`. They only differed in the ratio mode.

src/proteus/outgas/calliope.py
timlichtenberg: Fixed. Now that the noble gases are in gas_list, their partial pressures, mixing ratios, and reservoir masses are all in the helpfile schema and written every step. The one subtlety is the element total: a noble gas shares one key for its species and its element, and that total is escape-owned, so it is not copied back from the backend (the same rule every non-oxygen element total follows).

src/proteus/outgas/wrapper.py
timlichtenberg: Done, that special-casing is gone. M_atm now just sums gas_list, which includes the noble gases.

src/proteus/utils/coupler.py
timlichtenberg: Done, same simplification. The mass-conservation check sums gas_list directly now.

timlichtenberg:
Thanks Harrison, this was the right call. All five inline points are addressed in the latest commits: the noble gases are now full `gas_list` members, so they carry helpfile columns (partial pressures, mixing ratios, reservoir masses), enter the surface pressure and mean molar mass, and flow through the mass balance with the per-site special-casing removed. `_resolve_noble` is folded into `_resolve_element`, and the noble gases get the colours from the CALLIOPE branch.

Two things worth flagging. First, a noble gas shares one key for its species and its element, and that element total is escape-owned, so it is not copied back from the outgassing backend (the same rule every non-oxygen element total follows); the schema emits each noble mass column once through the gas-species loop. Second, I checked AGNI: it already knows all five noble gas molar masses and tracks gases that are not in the spectral file as inert, so a run carries them without needing an AGNI change, which matches your point about handling zero abundance gracefully. I am running a coupled AGNI case to confirm end to end.

timlichtenberg:
Correction on the AGNI point from my last comment: I was wrong that AGNI carries the noble gases as-is. The coupled run showed AGNI rejects them during thermodynamic-data setup (`Gas 'Ne' contains unsupported element 'Ne'`), so listing them in the gas set fails before the atmosphere solve. It does know their molar masses, but it has no thermodynamic data for them.

The fix keeps AGNI untouched: since the noble gases are tracked as a mass reservoir and are not radiatively active, I exclude them from the composition handed to AGNI (and from its rainout pass) in `3876b5ff`. A coupled AGNI case now completes end to end with the noble gases carried as tracked mass only. Full radiative coupling would need AGNI to gain noble gas thermodynamic data, which is the remaining piece for a later change.

timlichtenberg:
Ready for another look. All your review points are addressed, and I ran a further review pass over the whole branch and fixed what it turned up (the atmodeller mean-molar-mass and escape-ownership handling, the AGNI composition and surface-pressure scaling, and the CALLIOPE-side helium constant and docs).

One CI caveat: the fast checks and the full unit, integration, and AGNI/aragog/janus slow tiers all pass, but the nightly coverage gate is red because two Zalmoxis slow tests time out. I traced that to the Zalmoxis structure solve running about twice as slow as on the last main nightly (main finishes the single solve in ~25 min; here the same solve does not finish in 60 min). The noble gas changes do not touch that path: the only interior_struct/zalmoxis.py line they affect is the one-shot dry-mass sum, which adds zero for the noble gases in a run without a noble budget, so the structure-solve inputs are identical to main. It looks like an environmental slowdown (a fresh-install dependency or a slow runner), which I am tracking separately from this PR. Codecov confirms the real coverage is up (+0.06%).

timlichtenberg:
@nichollsh thanks for the review. I've reworked this so the noble gases are no longer special-cased.

`gas_list` is now `vol_list + vap_list + noble_gases`, so they are full gas species: their partial pressures, mixing ratios, and reservoir masses are in the helpfile schema and written every step, and they flow through the same plotting loops as the other species. The one asymmetry that stays is the element total, which is escape-owned and so not copied back from the outgas backend, the same rule every non-oxygen element total follows.

I also folded `_resolve_noble` into `_resolve_element` (they only differed in the ratio/solar mode), dropped the `M_atm` and mass-conservation special-casing in `wrapper.py` and `coupler.py` so both just sum `gas_list`, and added distinct preset colours for He, Ne, Ar, Kr, Xe in `_preset_colours` so they are no longer indistinguishable.

Ready for another look when you have a moment.

timlichtenberg:
A note on the red CI, since it is not a code problem with this PR.

The only failing check is the estimated-total coverage gate on the Linux unit-tests job. The unit tests themselves pass and diff coverage on the changed lines is 100%, so this is not a test failure or a real coverage regression.

That gate estimates full-suite coverage by unioning this branch's unit coverage with the most recent nightly coverage artifact. The most recent successful nightly currently comes from another in-progress branch, so its line numbers do not align with this branch and the executable-line denominator is inflated, which pulls the estimate to 88% against the 90% target. The properly aligned codecov/patch check stays green.

This resolves with the super-earth structure work in #706: once that lands and the shared nightly coverage baseline is refreshed, I will rebase this branch on top and the gate realigns. No change is needed on this PR for it.

nichollsh (changes requested):
Thanks for incorporating my suggestions. The coupling with the updated CALLIOPE branch works well.

A few more things to change before this is ready...

1) I have updated AGNI to support all of the same noble gases as PROTEUS. This was trivial three-line change in the AGNI source code, so it makes sense to handle this behaviour here as part of the same PR, rather than make special adjustments (which themselves total >3 lines). We can remove the parts of the AGNI wrapper (e.g. in `construct_voldict`) which skip the NGs.

2) The NGs still do not appear in the output plots from simulations. E.g. in `plot_global.png`, `plot_chem_atmosphere.png`, and `plot_escape.png`. Please can you update the plotting functions?

3) The NGs are still not being saved to the helpfile.

4) When PROTEUS announces the list of included gases at the start of the run, it skips the NGs from the list (only includes vapours and volatiles currently).

I tested the AGNI support for NGs with PROTEUS coupling locally, and it works. See below.

<details>
<summary>Click here to open dropdown</summary>

```
[ INFO  ] Loop counters
[ INFO  ] current    init    maximum
[ INFO  ]      10       3       9000 
[ INFO  ] ---------------------------------------------------
[ INFO  ] Evolve interior...
[ INFO  ] Time-stepping intent: slow down
[ INFO  ] New time-step target is 1.00e+04 years
[ INFO  ]     T_magma    = 3568.026 K
[ INFO  ]     Phi_global = 1.000  
[ INFO  ]     RF_depth   = 0.471  
[ INFO  ]     F_int      = 2.28e+05 W m-2
[ INFO  ]     F_radio    = 0.00e+00 W m-2
[ INFO  ] ---------------------------------------------------
[ INFO  ] Evolve orbit and tides...
[ INFO  ]     Orb SMaxis = 1.00000 AU
[ INFO  ]     Orb eccent = 0.00000   
[ INFO  ]     Orb period = 365.25633 days
[ INFO  ] ---------------------------------------------------
[ INFO  ] Stellar flux management...
[ INFO  ]     F_ins      = 1.030e+03   W m-2
[ INFO  ]     F_xuv      = 2.222e-01   W m-2
[ INFO  ]     T_star     = 5722.151    K
[ INFO  ] Updated spectrum not required
[ INFO  ] ---------------------------------------------------
[ INFO  ] Running EL escape (ZEPHYRUS) ...
[ INFO  ] Bulk escape rate = 6.36e+04 kg s-1
[ INFO  ] Elemental escape fluxes:
[ INFO  ]      H = 1.88e+01 kg s-1
[ INFO  ]      O = 1.50e+04 kg s-1
[ INFO  ]      C = 9.75e+03 kg s-1
[ INFO  ]      N = 5.20e+03 kg s-1
[ INFO  ]      S = 1.04e+01 kg s-1
[ INFO  ]     He = 1.82e+03 kg s-1
[ INFO  ]     Ne = 4.52e+03 kg s-1
[ INFO  ]     Ar = 2.65e+04 kg s-1
[ INFO  ]     Kr = 5.44e+02 kg s-1
[ INFO  ]     Xe = 2.75e+02 kg s-1
[ INFO  ] ---------------------------------------------------
[ INFO  ] Not desiccated, H = 1.55e+20 kg
[ INFO  ] Solving outgassing...
[ INFO  ]     CO         = 53.05     bar (2.94e-01 VMR)
[ INFO  ]     Ar         = 50.65     bar (2.81e-01 VMR)
[ INFO  ]     He         = 34.80     bar (1.93e-01 VMR)
[ INFO  ]     Ne         = 17.10     bar (9.49e-02 VMR)
[ INFO  ]     N2         = 14.19     bar (7.87e-02 VMR)
[ INFO  ]     CO2        = 8.96      bar (4.97e-02 VMR)
[ INFO  ]     Kr         = 0.50      bar (2.75e-03 VMR)
[ INFO  ]     H2O        = 0.43      bar (2.38e-03 VMR)
[ INFO  ]     H2         = 0.29      bar (1.61e-03 VMR)
[ INFO  ]     Xe         = 0.16      bar (8.89e-04 VMR)
[ INFO  ]     O2         = 0.11      bar (6.12e-04 VMR)
[ INFO  ]     SO2        = 0.02      bar (1.13e-04 VMR)
[ INFO  ]     total      = 180.27    bar
[ INFO  ]     mmw        = 26.96871  g mol-1
[ INFO  ] ---------------------------------------------------
[ INFO  ] Solving atmosphere...
[ INFO  ] Using nonlinear solver to conserve fluxes
[ INFO  ] Attempt 1
[ INFO  ]     sol_type = 2,     conv_type = 1
[ INFO  ]     skin_d   = 1.0e-02 m, skin_k = 2.00 W K-1 m-1
[ INFO  ]     step  |res_med|    cost     flux_OLR   max(x)     max(|dx|)  flags
[ INFO  ]        1  1.598e+03  4.082e+04  2.282e+05  2.974e+03  3.500e-03  Cs-C2-Ls
[ INFO  ]        2  1.361e+03  3.478e+04  2.146e+05  2.963e+03  3.460e+01  Cs-C2-Ls
[ INFO  ]        3  1.168e+03  2.941e+04  2.026e+05  2.954e+03  3.460e+01  Cs-C2-Ls
[ INFO  ]        4  9.809e+02  2.469e+04  1.919e+05  2.945e+03  3.460e+01  Cs-C2-Ls
[ INFO  ]        5  8.241e+02  2.054e+04  1.825e+05  2.938e+03  3.460e+01  Cs-C2-Ls
[ INFO  ]     success in 5 steps
[ INFO  ]     conduct. skin flux = +1.26e+05 W m-2 
[ INFO  ]     total flux at TOA  = +1.82e+05 W m-2     
[ INFO  ]     total flux at BOA  = +1.47e+05 W m-2     
[ INFO  ]     global flux loss   = +3.57e+04 W m-2  (+1.96e+01 %) 
[ INFO  ]     final conv.value   = +2.05e+04 W m-2     
[ INFO  ]     surf temperature   = 2937.526  K        
[ INFO  ]     surf pressure      = 1.803e+02 bar      
[ INFO  ] Attempt 1 succeeded
[ INFO  ]     oceans area frac   =  0.000 %
[ INFO  ]     oceans max depth   =  0.000 km
[ INFO  ]     R_obs photosphere  = 6864.1 km
[ INFO  ]     Planet Bond albedo =  1.785 %
[ INFO  ] ---------------------------------------------------
[ INFO  ] Runtime info...
[ INFO  ]     Wall time  = 2026-07-03 08:04:44   
[ INFO  ]     Model time = 6.92e+02   yr
[ INFO  ]     T_surf     = 2937.526   K
[ INFO  ]     T_magma    = 3568.026   K
[ INFO  ]     P_surf     = 1.80e+02   bar
[ INFO  ]     Phi_global = 1.00e+00   
[ INFO  ]     F_atm      = 1.82e+05   W m-2
[ INFO  ]     F_int      = 2.28e+05   W m-2
[ INFO  ] Checking convergence criteria

```
</details>

nichollsh (approved):
Great! Thanks for implementing my suggestions. Happy for this to be merged.

PR #723, opened 2026-07-04
Publish live test-count badges from a workflow

timlichtenberg:
## Description

This adds a workflow that keeps the test-count badges live instead of hand-maintained. PROTEUS renders static badge SVGs into the docs at build time, and the testing page carries fixed numbers that drift out of date (it has been showing ~1007 while `main` actually collects ~2766). The new workflow regenerates the shields.io endpoint JSON (`tests`, `unit tests`, `integration tests`) from a full test collection on every push to `main` and publishes it, so the testing page can point at a live count.

`main` is protected by the branch ruleset, so the workflow cannot commit the JSON back onto `main`. It publishes the three files plus a short README to the root of a dedicated unprotected `badges` branch, and the testing page reads them from that branch's raw URL. This matches the pattern already used for the protected-main repos in the ecosystem.

Getting the count right needs a complete install: a test module that `importorskip`s a missing optional dependency is dropped at collection, so a bare install undercounts badly (~1007 instead of ~2766). The workflow reuses the same `setup-proteus` environment CI uses so every optional submodule imports, and it refuses to publish any count that falls below a full-suite floor, so an incomplete install fails loudly rather than shipping a misleading low number.

The badge generator (`tools/generate_test_badges.py`) already exists and is unchanged; this PR only adds the publish workflow.

## Validation of changes

Ran the generator in a full local environment and confirmed the collected counts: total 2766, unit 2517, integration 260, all far above the ~1007 bare-install undercount. Confirmed that every collection-gating optional dependency (torch, botorch, juliacall, jax, atmodeller, vulcan, and the fwl-* submodules) is a core dependency that `setup-proteus` installs, so the CI collection reflects the whole suite.

Exercised the publish step's git logic against a throwaway local bare remote: it creates the `badges` branch on first run, is a no-op when the counts are unchanged, and overwrites cleanly when a count changes, leaving only the three JSON files and the README on the branch.

Unit-tested the count-floor guard: it passes the real counts (2766/2517/260), and fails a bare-install undercount (1007), a catastrophic zero, and a degraded unit-only count.

Tested on macOS with Python 3.12.

## Checklist

### DIFF
--- a/.github/workflows/publish-test-badges.yml
+++ b/.github/workflows/publish-test-badges.yml
@@ -1,0 +1,199 @@
+name: Refresh test count badges
+
+# Regenerates the shields.io endpoint-badge JSON (tests-total, tests-unit,
+# tests-integration) from a full test collection and publishes it to the
+# dedicated ``badges`` branch. The testing page renders these as shields.io
+# endpoint badges, reading the JSON from this branch's raw URL. An accurate
+# count needs a complete install: a test module that importorskips a missing
+# optional dependency is dropped at collection, so this job builds the same
+# environment CI uses (setup-proteus) before running the generator, then
+# refuses to publish a count that falls below the full-suite floor.
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'src/proteus/**'
+      - 'tests/**'
+      - 'pyproject.toml'
+      - 'tools/generate_test_badges.py'
+      - '.github/actions/setup-proteus/**'
+      - '.github/workflows/publish-test-badges.yml'
+  workflow_dispatch:
+
+permissions:
+  contents: write
+
+concurrency:
+  group: badges-${{ github.ref }}
+  cancel-in-progress: false
+
+defaults:
+  run:
+    shell: bash -el {0}
+
+jobs:
+  refresh-badges:
+    name: Regenerate tests-*.json
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Guard against publishing from a non-main ref
+        # The push trigger is already main-only, but workflow_dispatch can
+        # be launched from any branch. Publishing another ref's test count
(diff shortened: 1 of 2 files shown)


nichollsh (dismissed):
Good with me - thanks.

I did notice that the number of tests (2695) is actually different to the number you quoted, and also different to those collected by `pytest` (2710). Doesn't seem like a big problem?

timlichtenberg:
Good catch on the mismatch, here is what is going on with the three numbers.

The badge number (2766) is the complete-suite count. It is produced in the documentation build, where every optional submodule is installed and importable, so no test module gets skipped at collection time and every test file is counted.

Your local number is lower (2710) because a normal checkout is usually missing one or more of the optional modules. When an optional dependency is absent, the test modules that import it are skipped at collection, so they never enter the count. That is the whole 2766 vs 2710 gap: it is which modules are present, not a discrepancy in the counting.

The last step, 2710 down to 2695, is the filter. The badge counts with `not skip`, so the handful of tests marked skip are left out. A plain `pytest --collect-only` still lists those skip-marked tests, which is why the raw collect number sits a bit above the filtered one.

So the published figure is the full suite as it collects in a complete environment with the skip-marked placeholders excluded, and the smaller local numbers follow from missing optional modules plus whether the skip filter is applied.

@nichollsh, I have grouped the ecosystem-wide test-count badge work into six PRs and would like your review on the whole set together: FormingWorlds/PROTEUS#723, FormingWorlds/VULCAN#20, FormingWorlds/SPIDER#10, FormingWorlds/MORS#45, FormingWorlds/JANUS#67, FormingWorlds/ZEPHYRUS#26. Each adds the badge generator and a docs Testing suite page, and publishes the count JSON to a dedicated badges branch. Re-requesting your review here.

Issue #736, opened 2026-07-11
Interior-energetics wrapper: skip the non-superliquidus table read and warn on outsized baseline remap


timlichtenberg:
## Context

Two small follow-ups in `interior_energetics/wrapper.py` were raised while reviewing #706. Neither changes results: the first is a performance fix on a common path, the second a diagnostic that would make a rare event visible.

### 1. Short-circuit the super-liquidus guard so non-superliquidus runs skip the table read

`_grid_extent_up_step` is computed before `_superliq`, so non-`liquidus_super` zalmoxis runs pay the `np.loadtxt` of `zalmoxis_output.dat` on every accepted re-solve even though both guard branches are unreachable for them. Evaluating `_superliq` first short-circuits it: `_is_up_step = _scalar_radius_up_step or (_superliq and _eos_grid_extent_up_step(...))`. Behaviour is identical and the common path skips the file read.

Comment: https://github.com/FormingWorlds/PROTEUS/pull/706#discussion_r3539272124

### 2. Warn when the one-time baseline remap shift exceeds `mesh_max_shift`

The default `liquidus_super` IC skips the one-time baseline entirely, so stock configs never reach it. For zalmoxis configs with a non-superliquidus `temperature_mode` that do reach it, the IC-versus-callable representation offset lands on Aragog in a single unclamped remap at the first evolution step. I do not have a measured typical baseline shift yet, so a warning when the actual shift exceeds `mesh_max_shift` would make an outsized one-step swap visible in the log.

Comment: https://github.com/FormingWorlds/PROTEUS/pull/706#discussion_r3539279214

## People

@timlichtenberg @maraattia

PR #740, opened 2026-07-13
Wire dissolved volatiles into the Zalmoxis structure solve

### LINKED ISSUE #734: Wire the wet-mantle (dry_mantle=false) structure solve: composition-driven radius, volatile_profile dispatch, gas-phase mass
## Context

#706 keeps the interior-structure and outgassing coupling strictly dry: `dry_mantle = false` is rejected at config load, so the wet-mantle path is unreachable. Three review points identified code that is correct today only because the wet path is gated off, and that must change together when the volatile-partitioning (wet mantle) wiring lands. Collecting them here so the wiring branch does not have to rediscover them. Related physics issues: #519, #566, #432.

### 1. Composition-driven radius increase is currently rejected as an artifact

The super-liquidus reject branch in `interior_energetics/wrapper.py` treats every non-thermal `R_int` up-step as a cross-table artifact, and the sentinel refresh advances `_last_w_{species}_liquid`, so a composition-triggered re-solve that legitimately raises `R_int` while the mantle cools is rejected and its trigger baseline is consumed, meaning it can never re-fire. On a wet mantle this is a physical signal rather than an artifact: dissolving more volatiles into the melt lowers the mean mantle density while `T_magma` keeps falling, so the structure stays composition-frozen for the whole molten epoch. Proposed escape, gated on the wet path so it stays identically inert for dry runs: `_composition_driven = comp_changed and not config.interior_struct.zalmoxis.dry_mantle`, accepted alongside `_interior_reinflating`.

Comment: https://github.com/FormingWorlds/PROTEUS/pull/706#discussion_r3539141810

### 2. The structure dispatch does not carry `volatile_profile`

The dispatch in `interior_struct/zalmoxis.py` reads `use_jax`, `temperature_arrays`, and the EOS registry, but not `volatile_profile`. On the pinned fwl-zalmoxis, `structure_model.solve_structure` forces the numpy path whenever a volatile profile is present, and that path consumes only the callable. So the moment `dry_mantle = false` becomes reachable, a wet solve with `use_jax = true` on a 2-phase PALEOS mantle withholds the callable for a JAX path that never runs: the numpy solve falls back to the internal hot-anchor T profile, and the `_drop_callable` rebuild then overwrites the wet-blended density with a dry recompute from the bare `mantle_eos`. The fix is one conjunct, and `volatile_profile` is already in scope: add `and volatile_profile is None` to the dispatch expression, which the rebuild, seed, and hint gates all inherit.

Comment: https://github.com/FormingWorlds/PROTEUS/pull/706#discussion_r3539150537

### 3. Gas-phase mass should come from the solver, not `total - liquid`

In `outgas/atmodeller.py`, for every species with a `total_mass` output the code overwrites the gas mass with `total_kg - liquid`. That equals the gas mass only while no species holds a solid-phase reservoir. Once solid partitioning exists (trapped melt, or a melt-to-solid partition coefficient), `total - liquid` silently books the solid share into the atmosphere. When solid-phase partitioning lands, treat the solver's own `gas_mass` as the authoritative gas-phase write and keep `total - liquid` as a closure check, and record that assumption in the comment. Also decide whether condensed O belongs in the buffered `O_kg_total` (today condensates enter per element, so the identity holds while graphite is the only condensate). See also #519.

Comment: https://github.com/FormingWorlds/PROTEUS/pull/706#discussion_r3539243940

## People

@timlichtenberg @maraattia

maraattia:
## Description

This makes `interior_struct.zalmoxis.dry_mantle = false` a supported configuration: the dissolved volatile inventory reported by the outgassing chemistry is blended into the mantle EOS through the per-shell `VolatileProfile`, so the structure solve carries the melt's water instead of treating the mantle as bare silicate. Closes #734.

The wiring consists of five pieces plus documentation:

* The config gate on `dry_mantle = false` is lifted. `build_volatile_profile` converts the dissolved reservoirs to per-phase mass fractions, `extend_mantle_eos_with_volatiles` adds the species to the mantle EOS string, and the wet mass target excludes only the atmospheric inventory, since the dissolved mass is already inside the wet EOS.
* `w_liquid` divides the dissolved mass by the structural melt reservoir `Phi_global * M_mantle`, the same reservoir the chemistry dissolved into. Using Aragog's EOS-based `M_mantle_liquid` diluted the melt concentration by the wet blend density deficit and dropped 6.4 percent of the dissolved water from the structure.
* The temperature-source dispatch gains a `volatile_profile` conjunct (#734 item 2): wet solves keep the external T(r) callable and stay off the arrays hand-off and the post-solve rebuild, which recompute density from the bare dry mantle EOS and would overwrite the volatile-blended column. The retry, seed, hint, and rebuild consumers all inherit the single hoisted decision.
* The super-liquidus monotonic-radius guard accepts a composition-driven up-step on wet mantles (#734 item 1): dissolving more volatiles into the melt legitimately raises `R_int` while `T_magma` keeps falling, and the previous reject path also consumed the composition sentinels, so the trigger could never re-fire and the wet structure stayed composition-frozen for the whole molten epoch. Dry runs keep identical control flow. The guard now also evaluates the super-liquidus predicate before the grid-extent probe, so non-superliquidus runs skip the `zalmoxis_output.dat` table read on every accepted re-solve (#736 item 1; the baseline-remap warning of #736 item 2 stays open).
* The atmodeller gas-phase write books the solver's own `gas_mass` when the output carries it, keeping `total_mass` minus `dissolved_mass` as the fallback tier and as a logged closure check (#734 item 3). The two derivations agree today because no gas species holds a solid-phase reservoir; the check keeps a future solid phase from silently inflating the atmospheric budget.

The wet path is scoped to dissolved H2O under CALLIOPE. The docs note added in this PR spells out why (a species needs both a dissolved mass from the chemistry and a structure-side EOS mapping, and H2O is the only species with both), how unmapped dissolved species are carried (inside the mass target at silicate density, mass exact, compressibility approximated), and which gates keep dissolved H2 out until its physics is verified.

## Validation of changes

Linux x86-64, Python 3.12, conda environment matching the current pins (fwl-zalmoxis 26.7.13, fwl-aragog 26.7.4, equinox 0.13.8).

* Full unit and smoke suites pass locally; `ruff check` and `ruff format --check` are clean; diff coverage on changed lines is 100 percent (25 of 25 lines).
* New unit tests pin each piece: the wet dispatch keeps the callable and skips the dry rebuild on a JAX-viable EOS (discriminating against the dry twin test, which withholds the callable on the identical EOS pair); the composition escape accepts the wet up-step and still rejects the identical dry scenario; non-superliquidus runs never call the grid-extent probe; the solver `gas_mass` wins over `total minus dissolved` when they disagree, with the closure gap surfaced in the log; `w_liquid` uses the structural melt reservoir with the per-phase masses as fallback.
* Coupled wet verification run (1 Earth mass, 8.3 wt percent dissolved H2O, Zalmoxis structure with Aragog energetics at 150 levels): the dissolved-water conservation residual between the chemistry reservoir and the structure-integrated water mass is -0.157 percent, at the level of the trapezoid integration noise of the structure integral itself. The melt fraction profile is fully molten end to end, matching the reported global melt fraction, and the per-shell dissolved-water fraction is uniform at the melt concentration the chemistry dissolved.

## Checklist


tests/outgas/test_from_o_budget_atmodeller.py
```diff
@@ -860,6 +861,62 @@ def test_noble_gas_passed_as_mass_constraint_and_uses_total_mass_output():
...
+    derivations disagree (the future solid-partitioning case) must book
+    the solver's own gas_mass into the atmosphere, keep the difference
+    out, and surface the closure gap in the log so an untracked reservoir
+    cannot silently inflate the atmospheric budget.
+
+    Discrimination: gas_mass (1.8e16 kg) and total - dissolved (2.0e16 kg)
+    differ by 10 sigma of every tolerance used, so passing on the wrong
+    convention is impossible.
+    """
+    import logging
```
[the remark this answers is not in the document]
maraattia: Removed in e13a157b; the module level import at line 24 covers the test.

maraattia:
This is ready for review @timlichtenberg, thanks.

timlichtenberg:
On it. Sorry, cmd+enter closing the PR....^^

docs/Reference/config/interior.md
```diff
@@ -42,7 +42,7 @@ Chabrier, or Seager 2007). It supports 2-layer (core + mantle) and 3-layer
 | `ice_layer_eos` | str or none | `none` | Ice/water layer EOS; `none` = 2-layer model |
 | `mushy_zone_factor` | float | `0.8` | Solidus depression $T_\mathrm{sol}=f\,T_\mathrm{liq}$ \[0.7, 1.0]; the default 0.8 is the Stixrude (2014) solidus-to-liquidus ratio for MgSiO$_3$. Applies to `PALEOS:` unified only; for `PALEOS-2phase:` it is treated as 1.0 |
 | `mantle_mass_fraction` | float | `0` | Mantle mass fraction for 3-layer models; `0` = auto ($1 - \mathrm{core\_frac}$) |
-| `dry_mantle` | bool | `true` | Structure EOS assumes a dry mantle. `false` (melt-fraction-aware dissolved-volatile mixing in the mantle density) is rejected at config load until the pinned Zalmoxis release supports per-shell volatile profiles |
+| `dry_mantle` | bool | `true` | Structure EOS assumes a dry mantle. Set `false` for melt-fraction-aware dissolved-volatile mixing in the mantle density (per-shell volatile profile). When `false`, only the atmospheric inventory is excluded from the dry-mass target; the dissolved mass stays inside the interior |
```
timlichtenberg, line 45: The reference configs didn't get this update. `input/all_options.toml` (around line 290) and `input/inference/example.toml` (around line 309) still read `dry_mantle = true   # keep true; false is not yet supported`, and the block above each still says `false` is rejected until the pin supports per-shell profiles. The `global_miscibility` comment in both files also still says it needs `dry_mantle = false`, which is no longer the reason it's rejected. Please bring all four spots in line with this row so the shipped configs don't contradict the docs.

src/proteus/interior_energetics/wrapper.py
```diff
@@ -2561,7 +2561,10 @@ def temperature_function(r, P):
         # structure decision. When the mantle is genuinely heating (T_magma
         # rising, or Phi rising as melt returns) the up-step is physical and is
         # accepted; the accept criterion is the observed thermal change, not an
-        # assumed mechanism. A reject restores the previous structure (hf_row keys AND the
+        # assumed mechanism. A wet mantle adds one more physical up-step: a
```
timlichtenberg, line 2567: Correct, but this now disagrees with the running-minimum comment at around line 2499 ("each accepted re-solve only lowers R_int, so the last accepted value IS the running minimum"). That only holds for dry mantles now: a composition-driven wet re-solve raises R_int and is accepted right here. A one-line note about the wet exception up there would keep the two comments consistent.

timlichtenberg (changes requested):
Read through this in full and ran the unit and smoke suites locally. The wiring looks right, and keeping the whole thing behind `dry_mantle = false` means nothing on the default path moves, which I'm glad about. One thing to fix before I merge, plus two minor notes.

The blocker is doc consistency. `input/all_options.toml` and `input/inference/example.toml` still tell people `false` is not supported, and still say `global_miscibility` requires `dry_mantle = false`. Both are wrong now. `all_options.toml` is the reference config everyone copies from, so it has to match the behaviour this PR ships and the `interior.md` row you did update. Specifics inline.

The rest checks out:

- The atmodeller `gas_mass` write is identical to `total - dissolved` for gas species today (atmodeller builds a gas species' `total` as `gas + dissolved`, with no condensed term), so it's a safe no-op in production and the closure warning can't fire yet. Fine as defensive code.
- The grid-extent probe reorder is behaviour-preserving; `_is_up_step` only feeds the super-liquidus branches, so skipping the table read when the run isn't super-liquidus changes nothing else.
- The composition escape is bounded and only fires on wet mantles; dry runs are byte-identical.

Tests pass, ruff and the structure check are clean. I did not reproduce the coupled conservation numbers (the -0.157% residual and the structural-mantle denominator argument); those rest on your validation run, and the reasoning is sound.

Sync the two config files and I'll merge.

docs/Reference/config/interior.md
maraattia: All four spots are synced in 0e008ff6: both `dry_mantle` blocks now describe the wet path as supported, matching the interior.md row, and both `global_miscibility` comments state the actual rejection reason (the H2 silicate binodal handoff, Zalmoxis #64) instead of the stale `dry_mantle` dependency.

src/proteus/interior_energetics/wrapper.py
maraattia: Added in 2727be01: the running-minimum comment now names the composition-driven wet exception and the invariant that survives it (the bound is the last accepted R_int either way).

maraattia:
Both reference configs are synced in 0e008ff6: the `dry_mantle` comment blocks now describe the supported wet path, and the `global_miscibility` rejection is re-anchored to its actual reason, the missing H2 silicate binodal handoff (Zalmoxis #64), in both files. The running-minimum comment gained the wet exception note in 2727be01. Thanks for the careful read.

timlichtenberg (approved):
Both points are addressed: the reference configs now match the docs and the running-minimum comment carries the wet exception. Nothing else outstanding from my side. Good to merge on your end whenever you're ready.

Issue #728, opened 2026-07-09
PR coverage gate reads the newest nightly artifact from any ref, causing spurious coverage-drop failures


timlichtenberg:
## Summary

The PR coverage gate downloads the newest successful `ci-nightly` artifact regardless of which ref produced it, so a PR's coverage pass/fail depends on whether the last nightly happened to run on `main` or on a feature branch, not on the PR's actual coverage. This produces spurious "coverage dropped beyond grace period" failures that clear on an unrelated event (a nightly re-running on the branch).

## Mechanism

The `Download last nightly coverage (for estimated total)` step in `.github/workflows/ci-pr-checks.yml` (~line 262) uses `dawidd6/action-download-artifact@v6` with:

```yaml
workflow: ci-nightly.yml
name: nightly-coverage
workflow_conclusion: success
```

There is no `branch:` pin, so the action resolves to the most recent successful `ci-nightly` run on **any** ref. The `Fail job if coverage dropped beyond grace period` step (~line 504) then unions the PR's unit coverage with whatever baseline that artifact happened to carry and compares the estimated total against the fixed ceiling. When the last successful nightly ran on a feature branch (branch-specific coverage) rather than `main`, the baseline shifts and the gate can flip even though the PR's own coverage has not changed.

## Symptom

Seen on PR #706: `Unit Tests (Linux)` fails on the coverage-drop gate (step "Fail job if coverage dropped beyond grace period") while every other check is green. Re-dispatching `ci-nightly` on the branch clears it, because a fresh branch nightly refreshes the artifact the gate picks up. The pass/fail is decided by which ref last ran a nightly, not by the PR.

## Fix direction

Pin the artifact source to a stable ref so the baseline is deterministic. Add `branch: main` (or `github.base_ref` on `pull_request` events) to the download-artifact step so the gate always unions against `main`'s nightly, which matches the documented intent ("PR unit coverage union with the latest nightly artifact"). Alternatively, tag the `nightly-coverage` artifact with the ref that produced it and have the gate select the base-branch baseline explicitly, failing loudly if only an off-`main` baseline is available.

## Impact

No correctness risk to the code; this is a CI-reliability issue. It produces spurious red on PRs, erodes trust in the coverage gate, and makes the reported "estimated total" non-reproducible across time. Pinning the baseline removes the race without changing the coverage ceilings.

PR #742, opened 2026-07-14
Delete quarantined snapshots and warn on outsized baseline swaps

### LINKED ISSUE #733: Delete quarantined .incomplete snapshot files so they are not baked into data.tar
## Problem

When a resume selects the last complete interior/atmosphere snapshot pair, the trailing files past the chosen row are quarantined by renaming them to `*.incomplete`. Nothing ever deletes these afterwards. `_snapshot_time` does not recognise the `.incomplete` suffix, so `remove_old` skips them and the rolling snapshots-only archive excludes them, but the final `archive.update(remove_files=True)` sweeps every loose file, so the truncated netCDFs get baked into `data.tar` as corrupt members and are restored to disk on every later extract. Once a successful selection has truncated the helpfile below those rows, the quarantined trailing files can never be used again.

## Proposed change

Either delete the quarantined files after a successful selection (keeping the rename-and-rollback only for the failure path), or teach `remove_old` and the final archive to drop `*.incomplete`.

## Related comment from PR #706

> Once a file is quarantined as `.incomplete`, nothing ever deletes it: `_snapshot_time` does not recognize the suffix, so `remove_old` skips it and the rolling snapshots-only archive excludes it, but the final `archive.update(remove_files=True)` sweeps every loose file, so the truncated netCDFs end up baked into `data.tar` as corrupt members and are restored to disk on every later extract. After a successful selection the quarantined trailing files can never be used again, since the helpfile is already truncated below them. Could the selector just delete them at that point (keeping the rename-and-rollback only for the failure path), or alternatively teach `remove_old` and the final archive to drop `*.incomplete`?

_Originally posted by @maraattia in https://github.com/FormingWorlds/PROTEUS/pull/706#discussion_r3539252397_

## People

@timlichtenberg @maraattia

### LINKED ISSUE #736: Interior-energetics wrapper: skip the non-superliquidus table read and warn on outsized baseline remap

maraattia:
## Description

Two small follow-ups in the resume and structure-baseline machinery, split out of the #706 discussion. Neither changes physics results. Closes #733; together with the guard short-circuit carried by #740, closes #736.

Please review and merge #740 before this one: item 1 of #736 rides there, so the `Closes #736` line here is only correct once #740 is in. This PR is otherwise independent of #740 (no shared hunks) and can be reviewed in parallel.

* Quarantined snapshot files are now deleted after a successful resume selection (#733). When `select_resumable_snapshot` picks a resume row, the trailing snapshot files past that row are renamed to `*.incomplete`; previously nothing ever removed them, and since `_snapshot_time` does not recognize the suffix they were invisible to `remove_old` and the rolling snapshots-only archive but still swept into `data.tar` by the final unfiltered `archive.update(remove_files=True)`, restoring corrupt netCDF members to disk on every later extract. Once a selection succeeds the helpfile is already truncated below the quarantined rows, so those files can never be used again; the selector now deletes them at that point and logs how many were removed. The failure path is unchanged: when no resumable pair exists, the quarantined files are still restored to their original names before the RuntimeError, so a failed resume attempt remains lossless. Of the two fixes proposed on the issue, this one keeps `_snapshot_time`, `remove_old`, and the archive filters untouched, which avoids giving the shared time parser a second meaning. The deleted files are corrupt truncations with no forensic value beyond their timestamps, which the resume log already reports.
* The one-time baseline structure solve now warns when its mesh swap exceeds `mesh_max_shift` (#736 item 2). The forced baseline bypasses the mesh-shift clamp by design (`_blend_max_shift` is infinite under `force`), so the IC-versus-callable representation offset lands on the interior solver in a single unclamped remap. `blend_mesh_files` already reports the pre-clamp shift, so the wrapper now logs a warning with the actual shift and the configured cap whenever the forced swap exceeds it. The default `liquidus_super` initial condition skips the one-time baseline entirely, so stock configs never see this path; the warning makes the rare outsized swap visible for the zalmoxis configs with a non-superliquidus temperature mode that do reach it. The warning cannot fire when no previous mesh exists, since the shift reports as zero then; the forced baseline normally runs after the initial-condition solve has written one.

## Validation of changes

Linux x86-64, Python 3.12, conda environment matching the current pins.

* `tests/utils/test_coupler.py` and `tests/interior_energetics/test_wrapper.py` pass at the unit tier (249 tests). `ruff check` and `ruff format --check` are clean on all touched files.
* The deletion fix flips four existing assertions that pinned the old leak (the corrupt-atmosphere, corrupt-interior, corrupt-SPIDER-JSON, and cross-row-collision fallback tests now assert the quarantined files are gone after a successful selection), and a dedicated failure-path test now pins the restore-not-delete split: after a failed selection every quarantined file is back under its original name with its original bytes, with no .incomplete strays. The new warning is pinned by a threshold-discriminating test: a forced swap above the cap produces exactly one warning carrying both percentages, an equal run below the cap stays silent, and the forced path is asserted to leave the incremental mesh-shift machinery untouched.
* Both changes are unit-tested only; no end-to-end resume of a real run directory was exercised.

## Checklist

- [ ] I have updated the docs, as appropriate (no user-facing docs describe the resume quarantine or the forced baseline; nothing to update)

### DIFF
--- a/src/proteus/interior_energetics/wrapper.py
+++ b/src/proteus/interior_energetics/wrapper.py
@@ -2826,6 +2826,17 @@ def update_structure_from_interior(
             not force and actual_shift > config.interior_struct.zalmoxis.mesh_max_shift
         )
 
+        # The unclamped baseline can jump far from the previous mesh; surface
+        # that so a large swap is visible in the log. With no .prev mesh on
+        # disk, blend_mesh_files returns 0.0 and this cannot fire.
+        if force and actual_shift > config.interior_struct.zalmoxis.mesh_max_shift:
+            log.warning(
+                'One-time baseline mesh swap shift %.1f%% exceeds mesh_max_shift '
+                '%.1f%% and lands on the interior solver unclamped in a single step',
+                actual_shift * 100,
+                config.interior_struct.zalmoxis.mesh_max_shift * 100,
+            )
+
         # Track convergence steps; give up after 20 consecutive blends
         # to avoid infinite rapid-update loops when Zalmoxis and SPIDER
         # persistently disagree (e.g. extreme mass / low CMF)
--- a/src/proteus/utils/coupler.py
+++ b/src/proteus/utils/coupler.py
@@ -1229,7 +1229,10 @@ def select_resumable_snapshot(
     (when ``require_atm``) snapshots both open cleanly, move any incomplete
     trailing snapshot files aside (``.incomplete`` suffix) so the modules'
     latest-file globs land on the complete pair, and return the helpfile
-    truncated to that row.
+    truncated to that row. Once a resumable row is found, the quarantined
+    files are deleted: the helpfile is truncated below their rows, so they
+    can never back a resume and would otherwise be swept into the final
+    data archive.
 
     Each half is probed with its own writer's filename convention. The
     interior name depends on the module: Aragog writes ``'%d_int.nc'``
@@ -1315,6 +1318,14 @@ def select_resumable_snapshot(
             'No complete interior+atmosphere snapshot pair found in the '
             'helpfile history; cannot resume from disk.'
         )
+    if quarantined:
+        # The helpfile is truncated below the dropped rows, so these files
+        # can never back a resume; delete them so the final archive sweep
+        # does not pick them up.
+        for dst, _original in quarantined:
+            if os.path.exists(dst):
+                os.remove(dst)
+        log.info('Deleted %d quarantined snapshot file(s)', len(quarantined))
     if not dropped:
         return hf_all, []
     return hf_all.iloc[: keep_idx + 1].reset_index(drop=True), sorted(dropped)
--- a/tests/interior_energetics/test_wrapper.py
+++ b/tests/interior_energetics/test_wrapper.py
@@ -245,6 +245,76 @@ def test_force_baseline_ignores_unmet_triggers():
     assert forced[0] == pytest.approx(50.0, rel=1e-12)
 
 
+@pytest.mark.unit
+def test_force_baseline_warns_when_swap_exceeds_mesh_max_shift(tmp_path, caplog):
+    """The unclamped baseline swap warns when its shift exceeds mesh_max_shift.
+
+    force=True disables the blend cap, so a baseline mesh far from the .prev
+    mesh lands on the interior solver in a single step. A 15% shift against
+    mesh_max_shift=5% must produce the warning with both numbers; the
+    companion 2% run through the same path must stay silent, pinning the
+    threshold rather than an unconditional log line. In both runs the forced
+    path must leave convergence tracking inactive (no clamped intermediate
+    exists for later steps to reconcile).
+    """
+    hf_row = {
+        'Time': 1100.0,
+        'T_magma': 3000.0,
+        'Phi_global': 0.8,
+        'R_int': 6.371e6,
+        'gravity': 9.81,
+    }
+    (tmp_path / 'data').mkdir(exist_ok=True)
+
+    def _run(shift):
+        config = _mock_config(mesh_max_shift=0.05)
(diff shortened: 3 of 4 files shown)


maraattia:
This is ready for review @timlichtenberg after you finish #740, thanks. They are independent though.

timlichtenberg (dismissed):
Both changes look right, and I checked them against the surrounding code and the tests. The deletion is correctly placed after a successful selection, the failure path still restores the quarantined files and stays lossless, and the baseline warning is log-only and fires only under `force`. Tests, ruff check and format all clean on my end.

Two non-blocking notes for whenever you next touch this:

- Nothing actually asserts the failure-path restore: the no-complete-pair test only checks that the RuntimeError is raised, not that the files come back at their original names. Now that success deletes and failure restores, a small test pinning that restore-not-delete split would be worth having.
- Minor: the description says the failure-path test asserts the lossless restore; it asserts the raise, not the restore. No code impact.

Good to merge on your side. As you noted, land #740 first, since the `Closes #736` line depends on it.

maraattia:
Both notes taken: 88ab1ac5 adds the failure-path test pinning the restore-not-delete split (every quarantined file back under its original name with its original bytes), and the description sentence is corrected to match. On the red codecov/project status: the head combine tracks about 759 more source lines than the pre #740 base report (hits +606, misses +127, partials +26) while this PR only adds a test, so it is the baseline mixing from #728 surfacing on the external status, not a real drop; codecov/patch passes and the status is not in the required set. I'll merge after you approve @timlichtenberg, thanks!

timlichtenberg (approved):
Reviewed both changes and they hold up.

The quarantined-snapshot deletion is correct: after a successful selection those files sit below the truncated helpfile, so they can never back a resume, and because `_snapshot_time` does not recognise the `.incomplete` suffix they slip past `remove_old` and the snapshots-only archive only to get swept into the final full `data.tar` and restored as corrupt netCDF on every later extract. Deleting them at the selection point closes that. The failure path still restores under the original names rather than deleting, so a failed resume attempt stays lossless. No downstream code depends on the `.incomplete` files persisting.

The forced-baseline warning is purely additive and fires only when the unclamped swap genuinely exceeds `mesh_max_shift`; it cannot trigger with no `.prev` mesh since `blend_mesh_files` returns 0.0 there, so the default superliquidus configs never hit it.

Tests are discriminating (both-sides threshold for the warning, unique-byte restore check for the failure path) and pass locally at the unit tier. codecov/project's -0.57% is the usual unit-only-versus-full-base artifact; patch coverage is fine.

PR #758, opened 2026-07-17
Divide the eclipse depth by the stellar surface flux

timlichtenberg:
## Description

An eclipse depth compares the planet's emergent flux with the star's flux at the stellar surface, scaled by the squared radius ratio. The denominator was taken straight from the stellar spectrum PROTEUS writes to `data/<iter>.sflux`, but that spectrum is the flux arriving at the planet: it is scaled to the top of the atmosphere before it is written, so it already carries the reduction by the orbital distance. Dividing by it left that reduction in the result on top of the geometric factor, so every synthetic eclipse depth came out too large by the squared ratio of separation to stellar radius.

At ten stellar radii that is a factor of 100. At 1 AU around a solar-radius star it is roughly 46000, which pushes thermal-infrared depths past unity. Transit depths are geometric and were never affected.

The spectrum is now carried back out to the stellar surface inside `eclipse_depth` before it divides the planet flux. The `.sflux` file itself is unchanged: the climate and photochemistry modules read it as the flux at the planet and need it that way, so the correction belongs to the one consumer that wants the stellar surface. `atmos_chem/vulcan.py` already does this same rescale for its own use of the file.

Three docstrings in that chain described something the code does not do, which is how the double reduction went unnoticed. `scale_spectrum_to_toa` gave its separation in AU while the body divides the astronomical unit by it and so needs metres; `write_spectrum` described fluxes at 1 AU while it is handed values already scaled to the planet; and the loader said its spectrum could be used directly as the eclipse denominator, which is the premise that was wrong. The local variable holding the loaded spectrum was called `stellar_surface_flux` while holding the flux at the planet, and now names what it holds.

One limitation worth stating: the rescale uses the separation on the current helpfile row. That is exact for a fixed orbit, which is the default. While an orbit evolves, the stored spectrum may have been written at an earlier separation and the ratio of the two is left behind. That residual is far smaller than the error it replaces, and it is the same staleness the climate module already sees from the same file.

## Validation of changes

Reproduced first, before changing anything: with the planet at ten stellar radii the computed depth was `3.33333333e-07` against a hand-computed `3.33333333e-09`, i.e. exactly 100x, which is the squared separation-to-radius ratio.

- New test `test_eclipse_depth_divides_by_the_stellar_surface_flux` pins the depth against a hand-computed value with the planet at ten stellar radii, so the separation and the stellar radius differ and the two candidate denominators sit two decades apart. It carries sign, discrimination and scale guards, and a limit-input case at one stellar radius where the rescale is the identity.
- The neighbouring eclipse pins place the planet at one stellar radius, where the correction factor is exactly 1, so they are unchanged and stay green.
- Checked the fix against deliberately broken copies of the code: removing the rescale, inverting it, dropping the square, mixing centimetres with metres, and corrupting the wavelength mapping each make the new test fail.
- `pytest -m "unit and not skip and not slow and not integration"`: 2659 passed, 13 skipped. `ruff` clean, `validate_test_structure.sh` passes, `check_test_quality.py` reports no regression.
- macOS with Python 3.12.

Any synthetic eclipse spectra generated since the observe module landed carry the inflation and would need regenerating; I found none on this machine.

## Checklist

### DIFF
--- a/src/proteus/observe/petitRADTRANS.py
+++ b/src/proteus/observe/petitRADTRANS.py
@@ -370,8 +370,9 @@ def _load_stellar_toa_flux(
 ) -> np.ndarray:
     """Load the PROTEUS-written stellar spectrum from the latest ``data/<Number>.sflux`` file.
 
-    The spectrum is already scaled to the top of the planet atmosphere, so it
-    can be used directly in the eclipse-depth denominator.
+    The spectrum is scaled to the top of the planet atmosphere, so it carries the
+    reduction by the orbital distance. A consumer that needs the flux at the stellar
+    surface, such as the eclipse-depth denominator, must undo that reduction.
     Selects the .sflux file with the largest numeric prefix in the filename.
     """
 
@@ -660,7 +661,15 @@ def eclipse_depth(hf_row: dict, config: Config, source: str, dirs: dict[str, str
         )
         wl_local = np.array(wl_cm_local, dtype=float) * 1e4
         stellar_wavelength_nm = np.array(wl_cm_local, dtype=float) * 1.0e7
-        stellar_surface_flux = _load_stellar_toa_flux(outdir, hf_row, stellar_wavelength_nm)
+        stellar_toa_flux = _load_stellar_toa_flux(outdir, hf_row, stellar_wavelength_nm)
+        # The eclipse depth compares the two emitting surfaces, so the denominator is
+        # the flux at the stellar surface. The stored spectrum is the flux arriving at
+        # the planet, still carrying the reduction by the orbital distance, so undo
+        # that reduction here; dividing by it as it stands leaves the reduction in the
+        # result on top of the geometric factor below. This assumes the stored spectrum
+        # was written at the separation this row holds, which is exact for a fixed orbit
+        # and leaves the ratio of the two separations behind while an orbit evolves.
+        stellar_surface_flux = stellar_toa_flux * (sep / Rs) ** 2
         depth_local = (
             np.array(planet_flux_local, dtype=float)
             / stellar_surface_flux
--- a/src/proteus/observe/petitRADTRANS.py
+++ b/src/proteus/observe/petitRADTRANS.py
@@ -11,6 +11,7 @@ from typing import TYPE_CHECKING
 import numpy as np
 from scipy.interpolate import PchipInterpolator
 
+from proteus.star.wrapper import scale_spectrum_to_stellar_surface
 from proteus.utils.constants import (
     prt_cia_species,
     prt_gases,
@@ -669,7 +670,7 @@ def eclipse_depth(hf_row: dict, config: Config, source: str, dirs: dict[str, str
         # result on top of the geometric factor below. This assumes the stored spectrum
         # was written at the separation this row holds, which is exact for a fixed orbit
         # and leaves the ratio of the two separations behind while an orbit evolves.
-        stellar_surface_flux = stellar_toa_flux * (sep / Rs) ** 2
+        stellar_surface_flux = scale_spectrum_to_stellar_surface(stellar_toa_flux, sep, Rs)
         depth_local = (
             np.array(planet_flux_local, dtype=float)
             / stellar_surface_flux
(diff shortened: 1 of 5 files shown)


src/proteus/star/wrapper.py
```diff
@@ -378,7 +378,7 @@ def scale_spectrum_to_toa(fl_arr, sep: float):
         fl_arr : iterable
             Stellar fluxes at 1 AU
         sep : float
-            Planet-star distance, in units of AU
+            Planet-star distance [m]
```
nichollsh, line 381: Good catch. The helpfile has `separation` in units of m, so there's no error in the previous behaviour here. The docstring was just incorrect, so thanks for updating it.

src/proteus/star/wrapper.py
nichollsh: Could you update the header here to include information about where the spectrum corresponds to? That way, we can avoid future issues/confusion, since the `.sflux` files will be self-describing.

src/proteus/observe/petitRADTRANS.py
```diff
@@ -370,8 +370,9 @@ def _load_stellar_toa_flux(
 ) -> np.ndarray:
     """Load the PROTEUS-written stellar spectrum from the latest ``data/<Number>.sflux`` file.
 
-    The spectrum is already scaled to the top of the planet atmosphere, so it
-    can be used directly in the eclipse-depth denominator.
+    The spectrum is scaled to the top of the planet atmosphere, so it carries the
+    reduction by the orbital distance. A consumer that needs the flux at the stellar
```
nichollsh, line 375: Agreed

src/proteus/observe/petitRADTRANS.py
```diff
@@ -660,7 +661,15 @@ def _compute_eclipse_column(
...
         stellar_wavelength_nm = np.array(wl_cm_local, dtype=float) * 1.0e7
-        stellar_surface_flux = _load_stellar_toa_flux(outdir, hf_row, stellar_wavelength_nm)
+        stellar_toa_flux = _load_stellar_toa_flux(outdir, hf_row, stellar_wavelength_nm)
+        # The eclipse depth compares the two emitting surfaces, so the denominator is
+        # the flux at the stellar surface. The stored spectrum is the flux arriving at
+        # the planet, still carrying the reduction by the orbital distance, so undo
+        # that reduction here; dividing by it as it stands leaves the reduction in the
+        # result on top of the geometric factor below. This assumes the stored spectrum
+        # was written at the separation this row holds, which is exact for a fixed orbit
+        # and leaves the ratio of the two separations behind while an orbit evolves.
```
nichollsh, line 672: You highlight this slight self-inconsistency in the PR description. Overall, I think this is not a big problem for now. It can be mitigated by writing the stellar spectrum more frequently, if desired, but thanks for recording the caveat here.

nichollsh (dismissed):
I have one optional suggestion (see comments below).

This suggestion is optional but will helping avoid confusion in the future. Thanks for catching this inconsistency, @timlichtenberg, I  am otherwise happy for this to be merged once changes from `main` are synced.

src/proteus/star/wrapper.py
timlichtenberg: Done. The `.sflux` header now names where the fluxes are: `Stellar flux at the top of the planet atmosphere (scaled from 1 AU by the orbital separation)`, alongside `t_star` as before. That makes the file self-describing, and anyone who wants the flux at the stellar surface can see straight away that they have to undo the orbital scaling first. I kept it to a single line so the loaders that skip one header row (climate, photochemistry, observation and plotting) are unaffected.

src/proteus/star/wrapper.py
timlichtenberg: Agreed, only the docstring was wrong. The helpfile stores `separation` in metres and the body divides `AU` by it, so the computation was always correct; I just corrected the text to match.

timlichtenberg:
Thanks Harrison. I took the optional suggestion: the `.sflux` header now names where the spectrum sits (top of the planet atmosphere, scaled from 1 AU by the orbital separation), so the file is self-describing. The branch is rebased on top of the main-sync merge, so it is current with `main`. Replied to the individual points inline.

nichollsh (dismissed):
Thanks for incorporating my optional suggestion. Happy for this to be merged.

PR #759, opened 2026-07-17
Make petitRADTRANS an opt-in extra and fetch its opacity tables directly

timlichtenberg:
## Description

Synthetic observations need petitRADTRANS and about 5.6 GB of opacity tables, and neither was reachable from a clean install. The package was not a dependency at all, and the only route to the tables was `tools/get_prt.sh`, which drove a headless browser at the library hosting them, reported success whether or not anything arrived, and was wired into nothing: not CI, not `install-all`, not the docs.

This makes petitRADTRANS an opt-in extra and fetches the tables the way every other PROTEUS dataset is fetched.

**Changes**

- `pip install "fwl-proteus[observe]"` installs petitRADTRANS, pinned to a commit because the fork publishes no releases. It builds Fortran, C and C++ extensions, so the host needs gfortran and a C/C++ toolchain; the install guide now says so.
- Nothing else depends on it. It is absent from the core install, no module imports it unless a config selects `observe.module = "petitRADTRANS"`, and a machine without it is a supported machine.
- `proteus.utils.prt_data` fetches the tables over plain HTTP from the library the petitRADTRANS team publishes, which stays the source of truth. Only the tables PROTEUS asks for are fetched, and only those not already present, so an interrupted run resumes.
- `PRT_DEFAULT_FILES` names the file used for each species. Several species publish more than one line list over different wavelength ranges and they do not give the same spectrum, so the choice is written down rather than left implicit in an install script. The entries reproduce what PROTEUS has been running with.
- The install guide gains the line-list reference for each species, read out of the tables themselves, so anyone publishing spectra can cite the groups whose data produced them.
- `tools/get_prt.sh` now only writes the configuration file.

**Two things worth a reviewer's attention**

The library answers a request for a path it no longer holds with a web page and an HTTP 200, so nothing raises. A table renamed upstream would have been saved under the name of the table it replaced and counted as installed, with petitRADTRANS failing much later while parsing it. Each table is now staged under a temporary name and moved into place only once it has arrived in full and been recognised as HDF5.

The observe integration tests could not fail. The fixture caught every exception and turned it into a skip, and the tests walked over whatever spectra a run produced, so a run producing none left them green having checked nothing. They now skip when there is nothing to check and say why, and a failure of the run is a failure. Seven of the eight report themselves as not-run until the fixture uses an atmosphere module that writes a profile (a separate change); the eighth passes on its own merits. That is a smaller green number than before and a truthful one.

## Validation of changes

macOS 15 (arm64), Python 3.12, petitRADTRANS 3.3.3 and the full 5.6 GB table set installed locally.

- All 20 modelled species fetch and load: petitRADTRANS constructs a `Radtrans` over the 19 line species it asks for, plus 3 Rayleigh scatterers and 10 collision pairs, from the configuration this code writes. `uncovered_species()` returns empty.
- The rejection path was reproduced against the live library: a renamed path previously wrote a 4411-byte web page to disk as an opacity table and passed every completeness check; it now raises and leaves nothing behind. A real table still downloads and verifies as HDF5.
- 2671 unit tests pass. 12 cover the new module, two of them driving the real transfer against a stubbed HTTP response rather than a stand-in, since the staging and clean-up are the behaviour in question.
- `tools/get_prt.sh` runs against the real tree and writes 30 named files.
- Every reference in the docs table was extracted from the tables with h5py and resolved against doi.org: 14 cited, 14 extracted, none invented. Where a table records no usable reference, or records petitRADTRANS in place of the data's own source, the docs say so instead of guessing.
- `ruff check`, `ruff format --check`, and `bash tools/validate_test_structure.sh` clean.

## Checklist

### DIFF
--- a/.github/data-manifest.yaml
+++ b/.github/data-manifest.yaml
@@ -38,6 +38,14 @@ datasets:
   mors_evolution_tracks:
     family: Spada
 
+  # petitRADTRANS opacity tables (~5.6 GB), fetched only for configs that select the
+  # optional observe module. These come from the library the petitRADTRANS team
+  # publishes, which stays the source of truth; PROTEUS names which file it wants per
+  # species in proteus.utils.prt_data and caches the result here. Bump the version to
+  # re-cut the cache after any of those names change.
+  prt_opacities:
+    version: 2026-07-17
+
   # DACE / Zeng-2019 reference catalogs for the cpl_population plots.
   exoplanet_archive:
     version: 2026-05
--- a/docs/How-to/optionalmodules_installation.md
+++ b/docs/How-to/optionalmodules_installation.md
@@ -43,17 +43,103 @@ written in Julia.
 bash tools/get_lovepy.sh
 ```
 
-## Synthetic observations (PetitRADTRANS)
+## Synthetic observations (petitRADTRANS)
 
-[PetitRADTRANS](https://petitradtrans.readthedocs.io/en/latest/) generates
-synthetic transmission and secondary eclipse spectra.
+[petitRADTRANS](https://petitradtrans.readthedocs.io/en/latest/) generates
+synthetic transmission and secondary eclipse spectra. It is selected with
+`observe.module = "petitRADTRANS"`.
+
+Installing it takes three steps.
+
+**1. Install the package.** It builds Fortran, C and C++ extensions from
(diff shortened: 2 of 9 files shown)


timlichtenberg:
Closing; out of scope.

PR #760, opened 2026-07-17
Give every test exactly one CI tier so none runs nowhere

timlichtenberg:
## Description
A group of tests was running in no CI tier at all, and nothing went red because they were never collected. This gives every test exactly one tier, wires the ones that were unreachable into the job that should run them, and adds a guard so the gap cannot reopen.

The tier filters are mutually exclusive: the pull-request tier selects `unit and not slow and not integration` and the nightly slow tier selects `slow and not unit and not smoke and not integration`. A test carrying two tier markers matches neither, so sixteen functions were silently dropped. Six others had the opposite problem: the unit and smoke filters do not exclude each other, so a smoke test in a unit module ran twice. Three files predicted this in their own comments and were overruled by a later edit that added the very module-level marker the comment warned against.

Investigating the invisible tests surfaced that several were also broken, which is how they stayed hidden. The inference-pipeline tests do not complete: each worker subprocess stalls after startup with no output. They error even to reach that point because both test base configs set `observe.synthesis`, a key the schema does not have, so PROTEUS rejects the config before running. The grid tests do run, but a single module-scoped fixture drives a whole parameter grid and takes about 27 minutes, so their 300 second budget could never pass.

## Why
Every one of these tests reads as covered on the dashboard while testing nothing. A regression in any code path they exercise would ship green.

## Changes
- Rename `observe.synthesis` to `observe.module` in the grid and inference base configs so the configs load. This duplicates the same one-line fix in the pending AGNI and SOCRATES pin update; whichever lands second is a trivial merge.
- Split every mixed-tier file so each carries one tier, moving shared helpers into modules beside them. The schema cross-product, marked slow, sweeps in about 2.5 seconds and stays at unit tier, which its own docstring already stated.
- Move the grid tests to the slow tier with a budget matched to their 27 minute runtime, in their own Linux-only shard.
- Add the two slow files that no shard named to the slow matrix; leave the integration tier scoped to its own directory.
- Skip the three inference tests with the stall recorded, to be re-enabled once the pipeline completes.
- Add `tests/tools/test_ci_tier_coverage.py`, a static guard that fails when any test carries no tier or more than one, and when a tiered file is not reachable by its CI job.
- Point the remaining test docstrings at the documentation pages that exist.

## Testing
macOS, Python 3.12.

- Confirmed the partition is exact with `pytest --collect-only` against the four real CI selectors and the slow matrix: zero tests in no tier, zero in more than one, down from sixteen and six on `main`.
- Measured each re-tiered test rather than trusting its marker: the grid tests pass in 27 minutes with a real budget, the schema cross-product runs in 2.4 seconds, superliquidus passes in 4m21s, the moved smoke and unit tests pass in seconds.
- Confirmed the inference tests stall: the worker log stops after its opening warnings and the run produces no output.
- The new guard passes and was mutation-tested: it catches a reintroduced double marker, a slow file dropped from the matrix, and a directory dropped from the integration tier.
- `ruff check`, `ruff format --check`, and `bash tools/validate_test_structure.sh` pass; the pull-request tier is green.

## Checklist


nichollsh (dismissed):
Very nice. I had wondered about the `observe.synthesis` issue this week too, since I noticed that those tests errored when running locally and had assumed that the CI was taking a different approach. I didn't realise that the CI was entirely missing them - thanks for catching this!

The additional AGNI tests are nice to have, especially since the grey RT scheme will keep them quick. The BO inference tests are now becoming important as the BO scheme sees more usage. We should revisit these after I merge my BO branch sometime in the coming ~weeks.

I've checked that the broken documentation links now resolve.

Happy for this to be merged once synced with the `main` branch, @timlichtenberg.

PR #761, opened 2026-07-17
Make the Earth tutorial run out of the box

timlichtenberg:
## Description

Makes a fresh install able to run the Earth tutorial from a clean checkout with no manual data steps. Three changes: the fwl-zalmoxis floor rises to 26.07.17, whose liquid-table entropy routing the super-liquidus (liquidus_super) interior initial condition needs (FormingWorlds/Zalmoxis#79; older versions return NaN for fully molten points and the tutorial fails at the interior structure solve); the data auto-fetch now downloads the Seager 2007 static EOS set for every multi-layer mantle family whose registry entry carries the Seager iron core fallback, which the start-of-run existence check requires (previously a fresh install running the tutorial config, PALEOS-2phase mantle with PALEOS iron core, failed that check with no automatic way to obtain the file, even without --offline); and the tutorial header now shows the plain auto-fetch start command with --offline noted for later runs.

## Why

I want the Earth tutorial to work out of the box for everybody who newly downloads the code. Running it on a machine whose data directory lacked the Seager tables reproduced the fresh-install failure and surfaced the fetch gap.

## Changes

- pyproject.toml: fwl-zalmoxis floor 26.07.13 to 26.07.17, with the pin comment stating what the minimum carries.
- src/proteus/utils/data.py: the Seager static fetch triggers for the mantle families that reference the Seager core fallback (new module constant SEAGER_FALLBACK_FAMILIES), not only for explicit Seager components or an empty core EOS.
- tests/utils/test_data.py: a parametrized test pairs every fallback family with a non-Seager core so dropping any family from the constant fails its case; a structural test parses the registry source and requires exact agreement with the constant, so future registry families cannot silently miss the fetch; two tests pin the tutorial's exact EOS combination and the API-backed 2-phase limit case.
- input/tutorials/tutorial_earth.toml: header run command drops --offline and explains the first-run download.

## Testing

macOS, Python 3.12, proteus conda environment.

* tests/utils/test_data.py passes in full (189 passed, 2 skipped), and the new tests discriminate: with the fetch condition reverted, the family tests fail; with a family removed from the constant, both its parametrized case and the registry-agreement test fail.
* Reproduced the fresh-install failure path: with the Seager tables absent, the tutorial aborts at the existence check even without --offline; after the fetch fix the config resolves all its tables.
* fwl-zalmoxis 26.7.17 is live on PyPI and the interior test directories pass at the unit tier against Zalmoxis at that release commit (586 passed, 7 skipped).
* The full tutorial is running to completion on this branch as the final verification of the out-of-the-box path.

## Checklist

- [ ] I have updated the docs, as appropriate (tutorial docs page already documents the auto-fetch flow; results refresh follows separately after the verification run)


nichollsh (approved):
Fantastic - I agree that it's critical for the tutorials to "just work". I ran this on my laptop for ~60 PROTEUS iterations and can see that it replicates your behaviour. 

plot_global_log

I also checked the interior plot, which showed some strange behaviour in Zalmoxis cases in previous tests, and it looks great.

plot_interior


I wonder if we could make this run faster? For example, by using larger time steps, since your plot shows a lot of small oscillations at later stages that need not be time-resolved. Something for a future PR, since it's critical to have it "just work" for now.

Thanks, @timlichtenberg!

PR #762, opened 2026-07-18
Inference framework changes used in BO paper at point of submission

nichollsh:
## Description

This small PR includes changes that align PROTEUS with the code version used for the Bayesian Optimisation paper, at the point of ApJ submission.

**Changes...**

* Updated the relevant config files
* Extract gravity at photosphere, since it acts as a constraining observable. This is now put into the helpfile as `g_obs`, to complement existing keys.
* Add tests for gravity helpfile key
* Replace deprecated PyTorch `seed` key with `rng` key
* Requires `scipy>=1.15.0`

## Validation of changes

* The configs run on my laptop and the IoA cluster
* Tests pass on my laptop (Fedora 44, Python 3.12, Julia 1.12.6)
* Tests pass on GitHub CI

## Checklist


nichollsh:
@timlichtenberg, I believe that this is now ready for review. Would you be able to take a look at it, please?

src/proteus/atmos_clim/janus.py
```diff
@@ -293,13 +293,16 @@ def RunJANUS(
     # observables
     p_obs = float(config.atmos_clim.p_obs) * 1e5  # converted to Pa
     r_arr = np.array(atm.z[:], copy=True, dtype=float) + hf_row['R_int']
+    g_arr = np.array(atm.grav_z[:], copy=True, dtype=float)
```
timlichtenberg: `atm.grav_z` is stale here. Line 267 calls `atm.write_ncdf()`, which reassigns `self.z` from `integrate_heights` (`atmosphere_column.py:394`), but `integrate_heights` computes gravity as a local and never writes back to `atm.grav_z`. So `r_arr` is built from the new heights and `g_arr` from the old ones, and the returned `g_obs` disagrees with `g_s (R_int/R_obs)^2` by 4.9 to 6.4 percent in the cases I tried.

Simplest fix is to derive gravity from the radius this function already returns rather than carry a second profile. Compare AGNI, `atmosphere.jl:3278-3280`, which takes radius, temperature and gravity at one index on one grid.

src/proteus/atmos_clim/dummy.py
```diff
@@ -151,6 +151,7 @@ def _resid(x):
     # For the dummy climate model, the observed temperature is the surface temperature.
     # Use the locally computed value to avoid requiring hf_row["T_surf"] in unit tests.
     output['T_obs'] = T_surf_atm
+    output['g_obs'] = hf_row['gravity']
```
timlichtenberg: Same inconsistency, much smaller. `R_obs` is `R_int` plus `height_factor` scale heights but `g_obs` is the surface gravity, about 1.9 percent apart at the default `height_factor = 3`. `hf_row['gravity'] * (hf_row['R_int'] / R_obs)**2` costs nothing and keeps the invariant true.

tests/atmos_clim/test_agni.py
```diff
@@ -255,6 +255,7 @@ def __init__(self):
         self.transspec_p = 1e2
         self.transspec_r = 6.371e6
         self.transspec_tmp = 295.0
+        self.transspec_grav = 9.8
```
timlichtenberg, line 258: This adds `transspec_grav` to the fake atmosphere so the existing tests keep passing, but nothing asserts anything about `g_obs`, and there is no dummy or JANUS test at all. `atmos_clim` is a physics module, so please assert one invariant: the inverse square relation between the returned `g_obs` and `R_obs`, for each of the three modules.

src/proteus/inference/gen_D_init.py
```diff
@@ -259,7 +259,7 @@ def sample_from_bounds(
     #     Each of the parameters are evaluated in space 0-1, normalised to the bounds
     #     This variable is 2D, with shape [nsamp, dims]
 
-    sampler = Halton(d=dims, seed=np.random.default_rng(seed), scramble=True)
+    sampler = Halton(d=dims, rng=np.random.default_rng(seed), scramble=True)
```
timlichtenberg, line 262: `rng` is not accepted before scipy 1.15.0, and `pyproject.toml:80` is a bare `"scipy"` with no floor, so a clean install can resolve an older one and die at the first sampling call. Please pin `scipy>=1.15.0`. It also records which scipy the published runs used, which I want written down anyway.

tests/inference/test_gen_D_init.py
```diff
@@ -204,7 +204,7 @@ def test_sample_from_bounds_caps_workers_and_saves(monkeypatch, tmp_path):
     captured = {}
 
     class FakeHalton:
-        def __init__(self, d, seed, scramble):
+        def __init__(self, d, rng, scramble):
```
timlichtenberg, line 207: `FakeHalton` was updated from `seed` to `rng` in step with the source and line 236 substitutes it for the real class, so the real signature is never exercised and the suite passes on versions where production fails. Either construct a real `Halton` in one small test, or assert the floor directly.

src/proteus/utils/coupler.py
```diff
@@ -1360,9 +1361,9 @@ def variable_is_logarithmic(varname: str) -> bool:
         'eccentricity',
         'params.stop.time.maximum',
         'planet.elements.H_budget',
-        'planet.elements.C_budget',
-        'planet.elements.S_budget',
-        'planet.elements.N_budget',
+        # 'planet.elements.C_budget',
```
timlichtenberg: Please delete these three rather than leave them commented, and move the choice into the inference config. `transforms.py` uses this to pick log versus linear sampling and `plot.py:491,623,625` uses it for axis scaling, so the search design and the diagnostic axes move together and silently for every caller.

As it stands `H_budget` keeps log scaling over 1.30 decades while `C_budget` and `S_budget` lose it over 1.60, so the wider parameters now get the treatment meant for the narrow one.

input/inference/example.infer.toml
```diff
@@ -7,63 +7,67 @@ seed = 42
...
 # helpfile: /data/hen28/Projects/PROTEUS/output/bayesopt_se/runtime_helpfile.csv
-# extracted at index: 25, time: 5.000e+09 years
-# [observables]
-# "R_obs" = 8.5317284082e+06
-# "T_obs" = 4.1254645744e+02
-# "atm_kg_per_mol" = 2.1585436833e-02
-# "C/O_atm" = 4.5518929163e-01
-# "S/O_atm" = 1.2828302065e+00
+# extracted at index: 24, time: 2.000e+09 years
+[observables]
```
timlichtenberg, line 55: This grows the observable set from five entries to seven, adding `O/H_atm` as well as `g_obs`, and the PR description mentions only `g_obs`. Since the objective is computed over this set, adding two changes the relative weight of every existing one; please say so in the description and confirm the paper's runs used all seven.

Separately, these values do not match the sub-Earth row of Table 1 on five of seven entries, while the SN and TR blocks match exactly and every SE input parameter matches. The SE is the case we report as failing, so I want to be certain it was scored against the numbers we publish.

input/inference/example.infer.toml
```diff
@@ -7,63 +7,67 @@ seed = 42
 logging = "INFO"
 
 # Path to output folder where inference will be saved (relative to PROTEUS output folder)
-output = "bayesopt_infer_sn_mat32_logei_10/"
+output = "bayesopt_infer_se_mat12_lei_5_100/"
 
 # Path to base (reference) config file relative to PROTEUS root folder
-ref_config = "output/bayesopt_sn/init_coupler.toml"
+ref_config = "output/bayesopt_se/init_coupler.toml"
```
timlichtenberg: `ref_config` points at `output/bayesopt_se/init_coupler.toml`, which only exists after a previous run, while the commented line below it would work anywhere. For the version we tell readers to reproduce, please make the working path the active one.

Same theme: selecting a case means hand editing twelve blocks across two files with no consistency check, so three config pairs or a single `case = "SE"` selector would help. Also `init_samps = 3` is a very small initial design for five free parameters, is that what the paper used? And the typo at line 20, "edrrors".

timlichtenberg (changes requested):
Thanks Harrison. The AGNI `g_obs` extraction is right, and the configs now match the time-stepping and parameter bounds in the manuscript. Since this is the version we point readers at, I am holding it to "a reader can reproduce the published numbers from a clean clone" rather than "it runs". Four things before it merges.

The JANUS `g_obs` is not the quantity its name claims. It reads gravity off a height profile that `write_ncdf` has already replaced, so the gravity and the radius describe different layers and disagree by about 5 percent. AGNI is what the paper used, so the published numbers are unaffected, but we are shipping this key for all three climate modules.

The new observable has no test anywhere. A single assertion tying `g_obs` to `R_obs` through the inverse square law would have caught the JANUS case, and I would rather that lived in the suite than in my head.

`rng` needs `scipy>=1.15.0` in `pyproject.toml`. It raises `TypeError` on 1.14.1, and the test cannot catch it because it substitutes a fake sampler for the real class.

`variable_is_logarithmic` is a shared utility, so commenting entries out of it changes the search design for every future retrieval, not just this study. Please make it a per study setting instead.

input/inference/example.infer.toml
nichollsh: I have now ensured that the manuscript and config are aligned

input/inference/example.infer.toml
nichollsh: I used three samples for the paper, which is described in the manuscript. I agree that it's small, but I wanted to perform a scaling analysis based on the different BO algorithms, so use only a limited number of initial guesses, instead allowing for more evaluations by the BO retrieval algorithm.

nichollsh: Also, indeed it's a somewhat manual process to switch between the cases, but this is just for the paper and doesn't matter here. I'll update the config to point to the default one, so it works out of the box.

src/proteus/utils/coupler.py
nichollsh: I want to keep this PR consistent with the version of PROTEUS used for the paper. I think we can re-visit this the future, potentially when addressing #772. I'll remove the dead comments.

tests/inference/test_gen_D_init.py
nichollsh: This is corrected by enforcing scipy>1.15

src/proteus/inference/gen_D_init.py
nichollsh: I have updated `pyproject.toml` to enforce `scipy>=1.15.0`.

tests/atmos_clim/test_agni.py
nichollsh: Thanks for catching this. I'll add several tests which check for the gravity behaviour from the various climate modules.

src/proteus/atmos_clim/dummy.py
nichollsh: Good call. It's better to keep them consistent, so I'll calculate gravity from the dummy atmosphere module by scaling from the pseudo-height calculation.

src/proteus/atmos_clim/janus.py
nichollsh: Thanks for catching this one, as well. I'll derive gravity from the radius following the analytic description you've outlined. It's worth noting that this neglects self-gravity, which JANUS does internally anyway, so it remains self-consistent.

nichollsh:
Thanks for this, @timlichtenberg. I've updated the code to address these points and responded to your specific comments above. Ready for another pass :)

timlichtenberg (changes requested):
Looks good, thanks for working through all of these. The gravity fix in both `janus.py` and `dummy.py` is right, the new `g_obs` tests cover the inverse-square invariant per module, and I checked the SE observables against Table 1, all seven match now. Fine to leave the log-scaling choice for #772.

One thing before you merge: `ref_config` on line 13 of `example.infer.toml` still points at `output/bayesopt_se/init_coupler.toml`, which only exists after a run, so the example won't run from a clean checkout. Flip it to the `input/inference/example.toml` line commented below it and it's good to go.

nichollsh:
Done! Thank you.

timlichtenberg (approved):
ref_config points at the working default now, thanks. All good, approving.

PR #766, opened 2026-07-20
Harden the nightly shard guard so caps clear each test's timeout

timlichtenberg:
## Description

The heavy nightly tests run one per shard under a raised job cap. The guard that protects that sharding only checked that a cap was set and stayed under the 360 min hosted-runner limit; it never checked that the cap actually sits above the selected test's own pytest-timeout. When it does not, the runner cancels the job before the test's own timer fires, which is the exact failure the one-test sharding exists to prevent. This hardens the guard to enforce that relationship and trims one shard's timeout to satisfy it.

Changes:

- The guard now asserts each heavy shard's cap sits at least a setup budget (25 min) above the selected test's effective timeout.
- It reads the function-level timeout, which overrides the module-level one, in both the positional and the keyword forms, treats a disabled `timeout(0)` as unbounded rather than zero, and resolves the timeout against the exact matched slow test rather than the select substring, so a non-slow or skip-marked sibling whose name is a superset of the select cannot supply a smaller timeout that passes the check for the wrong reason.
- It keys the duplicate check on `(select, os)`, so a dual-OS parity shard is allowed while a same-runner repeat is still caught; flags a file that would run both per-test and whole; catches a test selected by two distinct substrings; handles an annotated `pytestmark` assignment; and reads the 210 min fallback cap the messages cite from the workflow so it cannot drift out of sync.
- To satisfy the new rule for the coupled std-config shard, its per-test timeout drops from 335 to 300 min. The run measures about 180 min, so 300 min still leaves wide headroom, and the 355 min cap now clears the timeout plus the budget.

## Validation of changes

All seven tier-coverage guard tests pass locally. A mutation that reverts the std-config-multi timeout back to 335 min reddens the cap-versus-timeout assertion (`test_select_sharded_files_cover_every_slow_test`), confirming the guard catches the failure it targets rather than passing by construction. ruff, test-structure, and test-quality gates pass.

Test configuration: macOS with Python 3.12; the guard test is unit tier and parses the workflow and test files textually, so it needs no compiled modules.

## Checklist


nichollsh (approved):
Good catch; thanks for this. Checking the slow testing-shard maximum runtimes now seems fully self-consistent between the CI configuration and the tests' markers.

Confirmed that these pass on my laptop.

PR #767, opened 2026-07-20
Drop committed run outputs from the examples

timlichtenberg:
## Description

The examples folder carried committed simulation outputs (per-example `proteus_00.log`, `runtime_helpfile.csv`, `status`, and the `plot_*.png` gallery) that added roughly 12 MB to the repository and did not track the current code. This removes them and keeps only the example configurations, one `init_coupler.toml` per subdirectory.

It also removes `tools/make_example.sh`, whose only job was to copy those outputs from an `output/` run back into `examples/`. With the outputs no longer committed the script would only reintroduce them, and nothing else references it.

The readme now describes the folder as example configurations and points at `output/`, where a run writes its results in a subdirectory named by the config's `params.out.path`.

## Validation of changes

- Confirmed nothing in `docs/`, `mkdocs.yml`, `tests/`, `src/`, notebooks, or CI references any deleted output path; the only `examples/` reference in the test suite (`tests/config/test_config.py`) reads `examples/minimal/init_coupler.toml`, which is kept, and that test still passes.
- Verified the readme's output-path statement against `src/proteus/utils/coupler.py` (a run writes under `output/<params.out.path>/`).
- Checked that each example subdirectory retains its `init_coupler.toml` and that no non-output file was removed.
- Test configuration: macOS, Python 3.12.

## Checklist


timlichtenberg:
@nichollsh could you review this one when you get a chance?

The `examples/` folder has been carrying committed run outputs: the per-run `proteus_00.log`, `runtime_helpfile.csv`, `status`, and the full `plot_*.png` gallery, roughly 12 MB that bloats the clone and drifts out of sync with the code, since they are snapshots from whenever they were last regenerated. This PR drops those outputs and keeps only the three `init_coupler.toml` example configs, so `examples/` becomes a set of runnable configurations instead of stale results. I also removed `tools/make_example.sh`, which only existed to copy a run's outputs back into `examples/`, and rewrote the readme to point at `output/`, where a run actually writes.

Nothing in docs, tests, or CI references the deleted paths (the one test that touches `examples/` reads a config that is kept, and it still passes). I am flagging it to you mainly because it removes the committed example gallery, in case you or anyone leans on those images as a visual reference. Happy to keep a curated subset somewhere if you would rather.

nichollsh (approved):
Agreed. These were outdated and simply adding bulky files to the codebase. Happy for them (and the script) to be removed.

Issue #774, opened 2026-07-23
Aragog NetCDF files use %d rounding rather than %.0f rounding


nichollsh:
## Problem description

Output files are written into simulations' `data/` subfolders. These are formatted as `<years>_<domain>.nc` where `<years>` is the simulation time (in years) and `<domain>` is either `int` or `atm`. 

However, Aragog writes `<years>=%d` while AGNI/JANUS/SPIDER all write `<years>=%.0f`. This is inconsistent, and becomes more divergent for longer simulation integration times. It means that the NetCDF files between the atmosphere and interior are mismatching.

## Expected behavior
PROTEUS should format the files with `%.0f` consistently throughout. This means updating the `aragog.py` function `_write_output_ncdf` to format the files appropriately.

## Additional notes
Related to #717 and previously noted by @timlichtenberg

timlichtenberg:
Resolved by PR #872, which uses p as the decimal separator in sub-year snapshot filenames (e.g. 884p700_atm.nc instead of 884.700_atm.nc), eliminating the ambiguity with file extensions.

PR #775, opened 2026-07-23
Write atmosphere NetCDF files consistently and improve reliability of VULCAN offline chemistry

### LINKED ISSUE #717: Atmosphere NetCDF files sometimes not written, and atmosphere plots sometimes not created
## Problem description

There are potentially two problems here, but they might have a common origin.

**Missing NetCDF**
Some of my recent changes to the `atmos_clim` wrapper aimed to bring its behaviour in-line with the `interior_energetics` wrapper, by writing NetCDF files according to `write_mod`. That is, not necessarily at every iteration. However, it seems that this sometimes causes the final iteration's NetCDF file to be skipped. This happens even when `write_mod=1`.

**Missing plot**
I also noticed that the `plot_atmosphere.png` files are not always created. This could be because the NetCDF files are missing. However, it could also be because the `cpl_atmosphere()` function isn't being called.

## Expected behavior

Even when `write_mod>1`, we should ensure that the final file is always written. Especially when `write_mod = 0` or `= 1`. In addition, PROTEUS should make the atmosphere plot when using AGNI or JANUS.

## My computer

Fedora 44, Python 3.12.13, Julia 1.12.6

- PROTEUS `90de97ffd42be268d4de1ac22704ace6525eb42d`
- AGNI 1.11.13
- netCDF-Fortran 4.6.2

### LINKED ISSUE #773: VULCAN offline chemistry crashes ('NoneType' object is not subscriptable)
## Problem description

VULCAN offline chemistry crashes in some scenarios, rather than running successfully (ideally) or exiting gracefully (worst case).

I suspect that this might be related to #717, since PROTEUS attempts to provide VULCAN with TP profiles from these NetCDF files. I don't think it's caused by the atmosphere having escaped, since the same error occurs for non-escaped cases in my grid of models.

```log
[ INFO  ] ===> Atmosphere has escaped! <===
[ INFO  ] 
[ INFO  ] Termination criteria satisfied
[ INFO  ] Writing data
[ INFO  ]  
[ INFO  ] Running atmospheric chemistry...
[ INFO  ]     Using vulcan module, offline
[ ERROR ] Could not find NetCDF file '/data/hen28/Projects/PROTEUS/output/l9859c_grid/case_000000/data/45113622_atm.nc'
[ WARNING ] One or more NetCDF files could not be found
[ CRITICAL ] Uncaught exception
  File "/data/hen28/miniforge3/envs/py312/bin/proteus", line 6, in <module>
    sys.exit(cli())
             ^^^^^
  File "/data/hen28/miniforge3/envs/py312/lib/python3.12/site-packages/click/core.py", line 1514, in __call__
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/hen28/miniforge3/envs/py312/lib/python3.12/site-packages/click/core.py", line 1435, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/data/hen28/miniforge3/envs/py312/lib/python3.12/site-packages/click/core.py", line 1902, in invoke
  File "/data/hen28/miniforge3/envs/py312/lib/python3.12/site-packages/click/core.py", line 1298, in invoke
  File "/data/hen28/miniforge3/envs/py312/lib/python3.12/site-packages/click/core.py", line 853, in invoke
    return callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/hen28/Projects/PROTEUS/src/proteus/cli.py", line 200, in start
    runner.start(resume=resume, offline=offline)
  File "/data/hen28/Projects/PROTEUS/src/proteus/proteus.py", line 1242, in start
    run_chemistry(self.directories, self.config, self.hf_row)
  File "/data/hen28/Projects/PROTEUS/src/proteus/atmos_chem/wrapper.py", line 72, in run_chemistry
    success = _run(dirs, config, hf_row)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/hen28/Projects/PROTEUS/src/proteus/atmos_chem/vulcan.py", line 104, in run_vulcan
    atmos = read_atmosphere_data(
            ^^^^^^^^^^^^^^^^^^^^^
TypeError: 'NoneType' object is not subscriptable
```

## Expected behavior

PROTEUS not to crash when providing data to VULCAN.

## Evidence

PROTEUS git hash `9f585d2a8184fff5a12a923faf4aeaa8dafdf863`.

See log file (excerpt above), config, helpfile here: [crash.zip](https://github.com/user-attachments/files/30296793/crash.zip)

## My computer

IoA cluster Red Hat Linux. Python 3.12.3 and Julia 1.12.6.

nichollsh:
## Description

This pull request introduces a few different improvements and fixes.

### Robust Atmospheric Data Handling

* Added `find_latest_atmosphere_time` in `common.py` to locate the most recent available atmosphere NetCDF snapshot, and updated `run_vulcan` in `vulcan.py`. Improving resilience to missing files.
* Refactored AGNI and JANUS climate models to use dedicated `write_atmos_ncdf` helper functions for writing NetCDF files
* JANUS writes files with `%0f` rather than `%d` string formatter, bringing it in line with AGNI. This makes locating and writing netCDF files consistent between atmosphere modules.
* Closes #717 

### Output Writing and Configuration

* Clarified the meaning of `dt_write_rel` parameter in documentation and config files
* This option now acts as an additional time-based write trigger combined with `write_mod` via logical OR, rather than logical AND.

### Chemistry Model 

* Adjusted VULCAN chemistry model time-stepping parameters for improved convergence and stability
* Calls function to update the atmosphere chemistry plot after running offline chemistry routine
* Resolves #773

## Validation of changes

* More tests added, which pass
* Existing tests pass, but Claude identified a new issue #791 that exists on `main`.

Tested on IoA Cluster (RHEL) with Python 3.12 and Julia 1.12

## Checklist


timlichtenberg (changes requested):
Thanks Harrison. The `boundary` case in the profile-time selection is a real fix (a boundary interior was intersecting against an empty list, so it plotted no profiles at all), the end-of-run atmosphere write is the right place for #717, and the new tests carry proper discrimination guards. Three things I would like changed first.

**JANUS resume loses its newest snapshot.** The JANUS writer now rounds, but `_atm_snapshot_names` in `src/proteus/utils/coupler.py` still probes JANUS with truncation, so for any row whose fractional year is at or above 0.5 the resume path looks for a file the writer never wrote, rejects the row, and deletes the interior half with it. I drove the real `select_resumable_snapshot` over a directory written by this branch; evidence and script in #795. The fix is to drop the JANUS branch in `_atm_snapshot_names` and update the three tests in `tests/utils/test_coupler.py` that pin the truncation.

**The convention is still split between atmosphere and interior.** Aragog writes `'%d_int.nc'`, so a JANUS plus Aragog run now disagrees on about half its snapshot names and `UpdatePlots` intersects those times away. That is #774, and I would rather it landed with this so the two halves move together instead of diverging in between.

**`dt_write_rel` going from AND to OR removes the throttle.** With the shipped `input/all_options.toml` (`write_mod = 1`, `dt_write_rel = 1e-3`) the iteration criterion is always true, so the parameter is inert and nothing can hold writes below the `write_mod` cadence any more. In the regime it exists for, a timestep floored at 1e-3 yr with elapsed time already at 1e6 yr, `main` writes nothing across 5000 iterations while this branch writes 5000 helpfile rows plus their interior and atmosphere netCDFs. Separately, `write_mod = 0` is documented as "wait until completion" but now writes periodically whenever `dt_write_rel > 0`. If dropping the throttle is deliberate, please say so in the description and in the config comment; otherwise a floor and a ceiling can coexist.

Smaller points:

- `src/proteus/proteus.py`: the comment above `last_write_time` still says `dt_write_rel` suppresses high-frequency writes, and the one above the final interior write says it may have suppressed the last write. Neither holds now. The new comment in the loop also lists "the final iteration is always written" as a criterion of `is_write_snapshot`, which that function does not implement.
- `src/proteus/atmos_chem/vulcan.py`: `year_int` is computed before the fallback reassigns `year`, so an online run that falls back writes `vulcan_<requested>.csv` holding chemistry from a different snapshot. Moving the filename block below the fallback fixes it.
- `find_latest_atmosphere_time` parses the time with a bare `float()`, so a single unexpected `*_atm.nc` in `data/` turns the graceful path into a `ValueError`.
- Em-dash in `tests/utils/test_coupler.py` line 3002.

The 389 unit tests across the touched modules pass here and formatting is clean. `ruff check` flags three import-order errors in `tests/escape/test_escape_boreas.py`, which are already on `main` and not yours, but PR lint stays red until someone clears them.

nichollsh:
Thanks for this, @timlichtenberg. I've addressed your issues in the last commit and am requesting a new review. I addressed the major and minor comments, where possible, although I am not sure the VULCAN one is a problem. Responses to major comments:

> JANUS resume loses its newest snapshot.

I've now updated all the snapshot read/write/search functions to ensuring homogeneity between AGNI and JANUS. This includes the tests.

> convention is still split between atmosphere and interior

All modules now write with `%.0f` - including Aragog. Updated PR description.

> dt_write_rel going from AND to OR removes the throttle. 

The intended change here is that `dt_write_rel` and `write_mod` act separately. Either can trigger a write, so this indeed means that the default config will write every iteration of `write_mod=1`. I believe that having them act separately is more intuitive; updated comments/docstrings to clarify.

nichollsh:
Fixed one broken test, should be ready now @timlichtenberg

timlichtenberg (changes requested):
Thanks Harrison. The three things I raised are all resolved, and I checked each against the branch rather than against the description.

The JANUS resume path works now: I drove the real `select_resumable_snapshot` over a directory named by this branch's writers, with helpfile times 10, 20.4, 30.7 and 40.5, and it keeps every row including the newest, where before it dropped everything from the first fractional row onward. Aragog moving to `%.0f_int.nc` closes the atmosphere/interior split, so the `UpdatePlots` intersection no longer throws half the snapshots away, and #774 goes with it. On `dt_write_rel`, your reading is reasonable and the docs and comments now say what the code does, so I am happy to take it as an OR.

Two small things I would like fixed, then I am happy to approve.

**The new float-parse guard crashes.** `find_latest_atmosphere_time` calls `UpdateStatusfile(output_dir, 22)`, but `UpdateStatusfile` expects the `dirs` dict and does `dirs['output']`, so passing the path string raises `TypeError: string indices must be integers` from inside the `except ValueError` handler. I reproduced it on the tip with a `data/` directory holding `30_atm.nc` and `final_atm.nc`. The guard was meant to make an unparseable name skippable and instead makes it fatal, inside the VULCAN fallback that exists precisely to degrade gracefully. I suggest dropping the `UpdateStatusfile` call and keeping only the warning: a stray filename in `data/` is not an atmosphere-model error, and this helper also runs during offline post-processing, where it would stamp "Error (Atmosphere model)" over a run that had already completed. Neither of the two new tests in `tests/atmos_clim/test_common.py` puts an unparseable `*_atm.nc` in the directory, so nothing catches this; one extra file in the existing fixture would.

**Every dummy-atmosphere run now ends on a warning.** `write_atmosphere_snapshot` checks `atmos_o._atm is None` before it checks the module, and nothing ever assigns `_atm` on the dummy path, so the end-of-run call logs `Cannot write atmosphere; atmosphere struct unallocated` on every dummy run. I confirmed it by calling the function with a real `Atmos_t()` and `module = "dummy"`. Moving the module check above the None check would make dummy return silently while keeping the warning for the case it was written for, an AGNI or JANUS run that never allocated. `test_write_atmosphere_snapshot_dummy_module_writes_nothing` passes a non-None `object()` as `_atm`, so it exercises a state a dummy run never reaches.

Smaller points, all optional:

- `tools/chili_postproc.py:127` still builds `f'{int(t)}_int.nc'` under a comment saying Aragog truncates. With the new writer it silently returns NaN viscosity for every time whose fractional part is at least 0.5. `f'{t:.0f}_int.nc'`, and dropping the comment, fixes it.

- `src/proteus/proteus.py:740` still says `dt_write_rel` is there "so that dt_write_rel can suppress high-frequency writes during rapid early evolution", which is the old AND behaviour.

- `src/proteus/config/_params.py:48`: "wait until completio".

- Three docstrings in `tests/utils/test_coupler.py` still describe the old split conventions. `test_atm_snapshot_names_single_convention` talks about the active module although the function no longer takes one; `test_select_resumable_snapshot_accepts_both_atm` still describes a truncated JANUS name; and `test_interior_snapshot_names_track_each_writer_convention` gives `%.0f.json` as the Aragog format too. In the second of those, the closing block repeats the identical `select_resumable_snapshot` call with the same arguments and re-asserts `dropped == []`, so it cannot fail independently of the assertion six lines above it.

- `_atm_snapshot_names` raises on negative time and has a test for it; `_interior_snapshot_names` raises too but does not.

One thing worth a line in the description rather than in code: output directories written before this change carry the old truncated Aragog names, so resuming one is no longer clean. With helpfile rows 10, 30.2 and 30.7, and a legacy `30_int.nc` / `30_atm.nc` pair that was actually written at 30.7, the selector rejects 30.7, falls back to 30.2, and resumes that row against a snapshot holding the state from 30.7. Saying plainly that older runs should be restarted rather than resumed would save someone a confusing debugging session.

nichollsh:
Great - thanks. I implemented the fixes for the major issues you identified. Also updated the docstrings and removed the redundant tests. 

I am leaving the CHILI one for now.

Ready for another review @timlichtenberg

timlichtenberg (approved):
Thanks, both fixed. I checked the tip directly: an unparseable snapshot name now returns cleanly with a warning and no status file, and a dummy run is silent while an unallocated AGNI struct still warns.

Approving.

PR #777, opened 2026-07-24
Configurable output root via PROTEUS_OUTPUT_PATH (revives #692)

egpbos:
## Description

Lets the output root directory be set with the `PROTEUS_OUTPUT_PATH` environment variable. By default every run is still written under `<proteus>/output`; when the variable is set to a non-empty value, that path is used as the output root instead, so simulation results can live off the code tree (cluster scratch storage, or a shared output area used by several checkouts). Leaving the variable unset (or empty) keeps the previous behaviour exactly.

This revives #692, which added the same `PROTEUS_OUTPUT_PATH` override but was closed as stale pending tests and documentation. Both are included here.

What changed:
- New helper `get_output_root(root_dir)` in `utils/coupler.py` resolves the root: `<root>/output` when unset, otherwise the env value. Empty or whitespace-only values are treated as unset (this avoids sending output to a path rooted at the empty string), and `~` and `$VAR` references are expanded.
- `get_proteus_directories` routes the `output` and `output/*` keys through that root, so every per-run subdirectory (`data`, `observe`, `offchem`, `plots`) follows it. Source-tree keys (`proteus`, `input`, submodule paths) are unaffected.
- `inference/gen_D_init.py` resolved the init-grid path from a hard-coded `<proteus>/output`; it now goes through the same output root, matching the idiom already used elsewhere in that file, so init grids are found under a relocated root too.
- Usage docs gain a short "Relocating the output root" section.

Naming: the variable is `PROTEUS_OUTPUT_PATH`, keeping the name from #692.

## Validation of changes

Tested on macOS with Python 3.12.

- Unit tests for `get_output_root` (unset default, set override, blank/whitespace fallback, `~` and `$VAR` expansion) and for `get_proteus_directories` (all output subdirs routed through the override; code-tree keys unaffected; unset case matches the legacy `<root>/output/<run>` layout). `tests/utils/test_coupler.py`: 124 passed.
- Updated `tests/inference/test_gen_D_init.py::test_create_init_routes_to_sample_from_grid` to pin the grid path through the output root: passes.
- `ruff check` and `ruff format --check` clean on the changed files; `tools/check_test_quality.py --check` reports no new violations; `tools/validate_test_structure.sh` passes.
- Not run locally: `smoke`/`integration`/`slow` tiers and a full end-to-end run, which need the compiled binaries (SOCRATES, AGNI, SPIDER) and full data. One pre-existing failure in `tests/inference/test_gen_D_init.py::test_sample_from_grid_builds_and_saves_dataset` reproduces on an unmodified `main` in this environment (a local NumPy/torch mismatch) and is unrelated to this change.

## Checklist

### DIFF
--- a/docs/How-to/usage_running.md
+++ b/docs/How-to/usage_running.md
@@ -100,6 +100,16 @@ To make every available plot:
 proteus plot -c input/all_options.toml all
 ```
 
+## Relocating the output root
+
+By default every run is written under the `output/` folder inside your PROTEUS installation, so `params.out.path` names a subfolder of `<PROTEUS>/output/`. Set the `PROTEUS_OUTPUT_PATH` environment variable to write runs somewhere else instead, for example onto fast scratch storage on a cluster or into a shared area used by several checkouts:
+
+```console
+export PROTEUS_OUTPUT_PATH=/scratch/$USER/proteus_output
+```
+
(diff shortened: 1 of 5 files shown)


timlichtenberg (approved):
Really nicely done. Clean, well-scoped change, and the tests are exactly the rigour we want to see: the discrimination guards and the back-compat layout pin especially. Approving, thanks.

egpbos:
Great! Should I merge myself or wait for additional reviews?

PR #781, opened 2026-07-24
Test: reproduce/guard data.tar inflation over the archive lifecycle

egpbos:
## Description

Phase 1 of an investigation into `data.tar` archive inflation. Production profiling of a 1.1 TiB, 54-run dataset found ~97–98% of `data.tar` bytes were byte-identical duplicates. This PR pins the behaviour with tests **before** any redesign (Phase 2, separate); no production code is changed.

Adds two end-to-end tests to `tests/utils/test_archive.py` that replay the full `proteus.py` archive lifecycle (rolling `update` + `remove_old` cycles, then the end-of-run `update(remove_files=True)` full archive) on a realistic mini `output/data/`, and measure inflation **by member content (SHA-256)**, not just by name:

- **`test_lifecycle_final_pass_stores_reference_data_once`** — regression guard that the static EOS table (`spider_eos/`) and fixed-name runtime files enter `data.tar` **exactly once** across the whole lifecycle. Extends the rolling-only coverage of `test_update_snapshots_only_bounds_tar_growth_across_cycles` (#706) to the final `remove_files=True` pass, and adds a content-level check that a fixed-name file which legitimately changes each cycle (`zalmoxis_output.dat`) is preserved as distinct versions rather than mislabelled as duplication.

- **`test_lifecycle_snapshot_reappend_no_inflation`** — encodes the *target* contract (store each snapshot once, inflation ≈ 1.0). Marked **`xfail(strict=True)`**: on `main` every timestamped snapshot lands in `data.tar` **~2×**, because the rolling archive keeps snapshots loose (`remove_files=False`) so each is re-appended on the next cycle before `remove_old` prunes it, and once more by the final full-archive pass (observed inflation ~1.45× on this fixture). `strict` so it flips to a hard failure — forcing the marker's removal — once the archive redesign removes the duplication.

Also adds a `_tar_content_stats` helper (gross/distinct bytes, per-name member and distinct-content counts). Pre-existing archive tests are unchanged.

Verdict recorded (measured on this fixture, not assumed): EOS/fixed-name files are **fixed on `main`** by #706's `snapshots_only=True` guard (green guard); `.nc`/`.json` snapshots **still inflate on `main`** (~2× per snapshot, strict `xfail`). The ~1.45× figure is fixture-specific and is not a claim about the production ratio.

No linked issue yet — Phase 1 of the archiving-inflation investigation.

## Validation of changes

- `pytest tests/utils/test_archive.py -q` → **21 passed, 1 xfailed**, no regressions in the pre-existing archive tests.
- `pytest ... --runxfail` confirms the xfail is a genuine `assert 2 == 1` (snapshot stored twice), not an incidental error masquerading as expected-fail.
- Quality/lint gates clean: `tools/check_test_quality.py --check`, `ruff check`/`format`, `tools/validate_test_structure.sh`.
- Test configuration: macOS (Darwin) with Python 3.12; CI additionally runs Unit Tests on Linux and macOS — all green on this PR.

Open question for reviewers (feeds the redesign, not asserted here): the `snapshots_only=False` path also retains multiple *distinct* versions of `zalmoxis_output.dat` (append-only tar keeps every version under one name). Before deciding whether a redesign must preserve that version history, confirm whether any resume/extract path ever reads an *older* in-tar version or whether an overwrite was intended (`tarfile.extractall` lets the last member win).

## Checklist

- [ ] I have updated the docs, as appropriate (n/a — test-only change)
- [x] I have checked that all dependencies have been updated, as required (n/a — no dependency changes)

### DIFF
--- a/tests/utils/test_archive.py
+++ b/tests/utils/test_archive.py
@@ -13,6 +13,7 @@ Testing standards:
 
 from __future__ import annotations
 
+import hashlib
 import os
 import tarfile
 from collections import Counter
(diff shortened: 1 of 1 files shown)


timlichtenberg:
@egpbos Please can you use the PROTEUS PR templates when creating the PRs?

egpbos:
> @egpbos Please can you use the PROTEUS PR templates when creating the PRs?

Done.

PR #783, opened 2026-07-25
Move the Bayesian-optimisation stack into an optional extra

timlichtenberg:
## Description

Every PROTEUS install currently pulls `torch`, `botorch` and `gpytorch`. Nothing outside `src/proteus/inference/` imports any of them, and that package is reached only through `proteus infer`, so a forward model, a grid sweep and the plotting path all pay for several hundred megabytes of dependencies they never touch. On Linux `torch` additionally pulls in the CUDA runtime packages.

They now ship as an `inference` extra:

```console
pip install "fwl-proteus[inference]"
```

**This is a breaking change for anyone who was getting torch transitively from `pip install fwl-proteus`.** It is worth calling out in the release notes for whichever version ships it.

Without the extra, `proteus infer` stops with a message naming the missing package and the install command. The classifier behind that message insists on all three of: the failure is a module-not-found, its name resolves to one of the three distributions, and the import system agrees the package is absent. Anything else propagates untouched, so a version mismatch inside the stack or a corrupted install is never disguised as "you forgot to install it".

## Changes

- `pyproject.toml`: `torch`, `botorch` and `gpytorch` move from `[project.dependencies]` to an `inference` extra.
- `src/proteus/cli.py`: `proteus infer` reports a missing extra with the package name and the install command; every other import failure propagates unchanged.
- `tests/inference/*`: each module guards the whole stack with `pytest.importorskip`, so an install without the extra skips instead of failing collection.
- `tests/test_optional_dependencies.py` (new): no module outside `src/proteus/inference/` may import the stack, directly or through `proteus.inference`; the CLI's distribution list and the test-quality linter's optional-dependency set are pinned to the extra declared in `pyproject.toml`.
- `.github/actions/setup-proteus`: CI installs `.[develop,vulcan,atmodeller,inference]`, alongside the existing optional backends, so the inference tests keep running there and coverage is unchanged.
- `tools/check_test_quality.py`: the three distributions join the optional-dependency set.
- Docs: a section in the optional-modules page (including PyTorch's CPU index for anyone who does not want the CUDA runtime), a requirement note at the top of the inference guide, and the extras needed to match CI's test selection in the testing guide, the manual install guide and the quick reference.

## Validation of changes

macOS 15 (Apple Silicon), Python 3.12, conda.

With the three packages blocked at import, simulating an install without the extra:

- unit tier: 2609 passed, 25 skipped, no collection errors
- `import proteus` and the CLI load
- `proteus infer` exits 1 with the missing-package message

With the extra installed:

- unit tier: 2708 passed, 11 skipped, unchanged from before the change
- inference tier: 105 passed, 3 skipped, unchanged

Packaging and partial installs:

- the built wheel lists all three only under `extra == "inference"`; no core `Requires-Dist` mentions them
- with only `gpytorch` removed, the inference directory produced 9 collection errors before this change and 11 skips after it

Each decision in the classifier and each structural guard was checked by mutation: dropping the module-not-found narrowing, the installed check, the membership check, the dotted-name resolution, the core-import guard, the `proteus.inference` guard, the pyproject-to-CLI link, or the linter entry each fails a specific test.

## Checklist

### DIFF
--- a/.github/.claude/rules/proteus-tests.md
+++ b/.github/.claude/rules/proteus-tests.md
@@ -188,7 +188,7 @@ pytest.importorskip('boreas')
 
 Optional deps that have hit this trap on `tl/interior-refactor`: `hypothesis` (three times), `boreas`, `atmodeller`, `lovepy`, `mors`, `vulcan`, `zalmoxis` (when not installed via editable).
 
-The lint script (`tools/check_test_quality.py`) enforces this. Rule key `missing_importorskip`: any module-top `import <optional_dep>` or `from <optional_dep> import ...` that is not preceded by a module-scope `pytest.importorskip('<optional_dep>')` is flagged. The check covers `hypothesis`, `boreas`, `atmodeller`, `lovepy`, `mors`, `vulcan`, `zalmoxis`.
+The lint script (`tools/check_test_quality.py`) enforces this. Rule key `missing_importorskip`: any module-top `import <optional_dep>` or `from <optional_dep> import ...` that is not preceded by a module-scope `pytest.importorskip('<optional_dep>')` is flagged. The check covers `hypothesis`, `boreas`, `atmodeller`, `lovepy`, `mors`, `vulcan`, `zalmoxis`, `torch`, `botorch`, `gpytorch`.
 
 ---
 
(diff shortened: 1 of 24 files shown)


nichollsh:
This is a good call. These dependencies take up a lot of disk space. Thanks, @timlichtenberg .

Issue #407, opened 2025-05-08
Parametrised mass loss from giant impacts


nichollsh:
https://arxiv.org/abs/2505.04343

> ...we present a scaling law that precisely approximates loss from 0.35 to 5.0 Earth mass planets with 5% mass fraction H-He atmospheres for any combination of impactor mass, impact velocity, and angle. 

Such a scaling law could be included in PROTEUS. It would potentially account for the stochasticity of such events.

> ... while individual impacts rarely cause significant loss (10%) from roughly Earth-mass planets with such massive primary atmospheres, the cumulative effect of multiple impacts can be substantial (40-70% loss).

timlichtenberg:
Anna Grace Ulses, PhD student in Josh KT's group will visit us this summer and work with Tadahiro Kimura and me on this project, starting with allowing remelting of the mantle and timing of GIs.

PR #789, opened 2026-07-26
Grow the planet by giant impacts: the accretion module and Morrigan

### LINKED ISSUE #168: Volatile delivery
We now have the initial components of an escape model integrated into PROTEUS. However, it is possible that accretion is also contemporaneous with magma ocean evolution. It would be interesting to explore what role this plays in the changing volatile inventory/distribution of young planets.

It should be relatively easy to "add" additional volatile mass to the inventory as the planet evolves, in the same manner in which escape removes it. We would need some estimate of the rate at which this occurs.

@timlichtenberg is a good idea and is there a nice way to estimate this?

timlichtenberg:
## Description

This adds `accretion`, a module that grows the planet by giant impacts while the magma ocean evolves, and couples Morrigan into PROTEUS as its dynamical implementation. Closes #168. Related to #407, which asks for a parametrised impact mass-loss scaling; the loss used here is the Kegerreis et al. (2020) law evaluated by ZEPHYRUS, not the scaling that issue proposes.

An impact is applied at the end of the timestep that lands on its time, so the orbit and structure of that step already use the grown planet. Each collision adds the impactor's rock and re-solves the interior structure at the new mass, delivers the impactor's volatiles while the same collision fraction strips both bodies' atmospheres, re-melts the mantle by re-applying the run's temperature-mode initial condition, and moves the orbit by the impact's change in semi-major axis and eccentricity, clamped to a bound orbit. The whole-planet mass closes to before plus rock plus delivered minus stripped, which can be a net shrink when a small impactor blows off a heavier atmosphere.

Three implementations:

- `morrigan` follows a system of embryos through the giant-impact phase with the semi-analytical Monte Carlo model of Kimura et al. (2025) and hands PROTEUS the impact history of one survivor.
- `timeline` replays an impact history from a file, for reproducing a published history or driving PROTEUS from one computed elsewhere.
- `dummy` builds a history from scaling laws: exponential approach to an asymptotic mass, evenly spaced impacts, Noack & Lasbleis (2020) radii, momentum-conserving mergers.

Impactors are dry by default, can carry the planet's own formation abundances, or can carry per-element budgets set in the config. Impact atmosphere loss is off by default, and can be a fixed fraction or the ZEPHYRUS erosion law.

Morrigan is an optional dependency, installed with `pip install "fwl-proteus[morrigan]"` or `tools/get_morrigan.sh`, and pinned by its PyPI floor like VULCAN. The floor is release 26.07.27, which carries the accretion coupling documentation and the tutorials that walk through the configuration below.

The two documentation figures now show the module. The code-architecture diagram gained an Accretion stage in the coupling loop and an accretion module group, and is generated from a model in `tools/figures` rather than drawn by hand, so its light and dark variants come from one source and it rebuilds with pdflatex. Both figures' labels are real vector text, so a still export shows the same wording as a browser.

## What review would help most with

The physics is being tested and developed by Anna Grace, Tadahiro and me, so review effort is best spent on the **software engineering and the framework integration** rather than on the impact physics itself.

That said, this touches a lot of the framework, and a look at those parts would be very welcome:

- the new module's shape against the rest of `src/proteus/`: the wrapper and dispatch pattern, the configuration schema and its validators, and whether anything here should have been shared rather than added;
- where the step sits in the coupling loop, and what it mutates: `config.planet.mass_tot`, the structure re-solve, the volatile budgets, the solidification latch, and the orbit;
- the resume path, since the impact history has to survive a restart and is written to the run directory for that reason;
- the interaction with modules it did not previously touch, in particular the interior re-melt and the ZEPHYRUS collision call;
- test coverage and whether the seams are tested at the right level;
- the documentation and the figures, which now carry a generated source.

## Validation of changes

**Unit and integration tests.** New tests cover the impact accounting, the volatile delivery and strip, the re-melt guards, the timeline resolution and resume path, and the analytical module's scaling limits.

**Coupled capstone run.** A full run with Aragog, AGNI, CALLIOPE, ZEPHYRUS and Zalmoxis, starting from 1 M⊕ at `temperature_mode = "liquidus_super"`, with Morrigan seed 7 over eight embryos and `atmloss_module = "zephyrus"`:

| Quantity | Result |
|---|---|
| Impacts applied | 2, at 912 yr (+0.70 M⊕) and 1939 yr (+0.50 M⊕) |
| Final planet mass | 1.30e25 kg (2.18 M⊕) |
| Impact re-melt heat | 3.49e30 J and 6.11e30 J |
| Solidification | 2.014 Myr at global melt fraction 0.05 |
| Steps | 3772 |

The run logged no clamp or strip warnings, so those paths did not fire in this configuration. It ran from the branch at "Depend on the published fwl-morrigan release instead of a commit"; the commits after it correct the impact accounting and the orbit update, and were assessed against this run rather than re-run, so the numbers above are not a validation of the branch tip unqualified.

**Figures.** The architecture diagram was reproduced from the previous one before anything was added, and checked against it: 98.6 % of label lines within one pixel, and 0.009 % of non-text pixels differing by more than a quarter step, all of it corner antialiasing. Every region that only moved was then checked as a pure translation. All 42 links in the figure resolve to a real file, and every line anchor into `proteus.py` lands on the call site of the stage it labels. A geometric check over both figures, which flattens every stroked shape and allows for paint order, confirms that no line crosses a label and no two labels overlap.

**Golden trajectory.** This branch adds two helpfile columns, `M_accreted_rock` and `step_dE_impact_J`, and writes them on every run rather than only when accretion is on, so the recorded trajectory that `test_integration_golden_run.py` compares against no longer covered everything the run produces. It is recorded again here, on top of the reference #812 recorded for the four columns #809 added, so this one carries both sets rather than either alone. Nothing else about it moved: the run still produces 56 rows, the count goes 761 to 763, all 761 columns already recorded are identical to the bit with none differing, the configuration is untouched, and both new columns are zero down every row.

What that does and does not establish is worth stating, because re-recording a reference is exactly the step that can bury a regression. The golden configuration runs with accretion off, so the comparison never enters `apply_impact`, the stripping and delivery arithmetic, or the re-melt energetics. It establishes that this branch leaves a run without impacts completely unchanged, which is what protects everyone else's runs from it, and it now pins both new columns at zero on that path with zero tolerance. It is not a validation of the impact physics; that rests on the accretion unit tests, the smoke and slow accretion tests, and the coupled run reported above.

**Test configuration.** macOS 15 (Apple silicon), Python 3.12.13, Aragog 26.7.4, fwl-morrigan 26.7.27.

## Checklist


nichollsh:
@timlichtenberg we can discuss this at the meeting, but would you like me to review this PR at some point? What is the timescale?

timlichtenberg:
@nichollsh Yes, this would be good actually, thank you! I would like someone to have an independent eye on this. Not super urgent, we are basically testing this right now. The major focus should be on the framework integration, not physics, as we are still testing this and are probably going to add a few changes. But this PR basically brings in the whole scheme, and follow-up PRs would be much more confined.

PR #790, opened 2026-07-27
Check that a fixed run still produces the trajectory it produced

timlichtenberg:
## Description

A change that is meant to leave behaviour alone can only be shown to have left it alone by running the same configuration before and after and finding the same numbers. Nothing in the suite currently notices when a change alters the numbers a run produces without breaking an assertion, so this adds that check: a fixed all-dummy configuration, the trajectory it produces, and the comparison between the two.

Everything lives under `tests/`, alongside `tests/helpers/_smoke_invariants.py`, which is the same shape of thing: a verification library the suite imports and unit-tests in place.

While working on a change, running the check reports which quantities moved, at which row, and by how much relative to the tolerance. It takes a few seconds:

```bash
pytest tests/integration/test_integration_golden_run.py
```

Once a change to the trajectory is the intended outcome, record the reference again in the same commit, so the diff of `golden_run.tsv` shows what moved:

```bash
pytest tests/integration/test_integration_golden_run.py --record-golden
```

What it covers is the coupling framework: the main loop, the timestep controller, the module handshakes, the helpfile round-trip and every configuration field the run reads. It does not cover the physics modules, which run on their dummy backends here.

Complements #784, which checks that stopping and resuming a run changes nothing. That comparison is between two runs made in one process; this one is against a trajectory recorded earlier, so it also holds across commits.

## Changes

- `tests/helpers/_trajectory.py`: runs a configuration, records a trajectory, reads one back, and compares two column by column, reporting which columns disagree, where, and by how much. Also decides the two outcomes the check acts on, so a missing reference and a reference belonging to another configuration are each reported as themselves rather than as a wall of differing columns.
- `tests/integration/golden_run.toml`: the fixed configuration, with everything that steers the run written out rather than left to a schema default.
- `tests/integration/golden_run.tsv`: the recorded trajectory, one line per helpfile column so a quantity that moves shows as one changed line in a diff.
- `tests/integration/test_integration_golden_run.py`: the comparison, integration tier.
- `tests/helpers/test_trajectory.py`: unit tier. The checks that need only the recorded file, that it belongs to the committed configuration and that it records a run where quantities are moving, live here so they run on every pull request.
- `tests/conftest.py`: registers `--record-golden`.
- `docs/How-to/testing.md`, `docs/Explanations/test_framework.md`: how to run the check and record the reference again, and why a parity comparison is written the way it is.

Points worth a reviewer's attention:

- **The comparison covers every helpfile column**, not a chosen list, and treats an undefined value as state: two NaNs at the same row agree, a NaN against a number does not. Screening non-finite values out first would skip precisely the columns a change corrupted.
- **The tolerance is relative with an absolute term scaled to each column's own range.** A purely relative tolerance demands exact agreement wherever a quantity passes through zero, which is where cancellation makes agreement least exact. A column that is zero at every row gets no tolerance and has to stay exactly zero, which is the behaviour to want: those are quantities the configuration never switches on.
- **The reference carries a digest of the configuration it came from**, so a configuration edited without recording again is reported as that rather than as an unexplained difference across hundreds of columns.
- **Everything that sets a step size is pinned in the configuration**, which is not only `[params.dt]`: the interior clamps the step against its own temperature tolerances, and the observed radius depends on a scale-height factor. Each of those moves the trajectory when its schema default changes.
- **Recording reports what it accepted through the skip reason** rather than printing it, so it is visible under the suite's ordinary output capture instead of needing `-s`.
- **`run_trajectory` redirects `TMPDIR`** for the duration of a run. The framework allocates its scratch directory there while the runner is being constructed, before the configuration has been checked, so a configuration with one unrecognised key raises out of the constructor with the directory already on disk and nothing holding a reference to it. Redirecting is what makes that removable.

## Validation of changes

Configuration: macOS 15.7 (arm64), Python 3.12, conda environment with the pinned module set, `proteus doctor` all-green.

- Unit tier 2881 passed, 11 skipped. Integration tier 205 passed, 4 skipped, run locally in full since this adds a file there.
- **Reproducibility, measured.** Three runs of the configuration agree bit for bit on all 757 state columns; only wall-clock time differs. Values round-trip through the reference file exactly, so the file is not a term in the tolerance.
- **Sensitivity, measured.** Scaling `dTdt` in `src/proteus/interior_energetics/dummy.py` by `1 + eps` and running the comparison: `1e-6` is reported at 49 times the tolerance, `1e-7` at 4.9 times, `2e-8` sits inside it. So the threshold is just above `2e-8`. The column that reports it is not the one perturbed: a change to the interior is amplified on its way into the volatile partitioning, since dissolved volatiles depend steeply on melt fraction near solidification.
- **The tests were checked against deliberate defects**, not only run. Every guard in the comparison was reverted in turn and the suite required to fail. One was written to survive: with the interior step tolerances now pinned, changing their schema defaults must no longer move the trajectory, and it does not.
- **Discriminating power of the fixture, measured.** 106 of the 757 recorded columns move over the run. Switching escape off makes the elemental totals, the per-element escape rates, the cumulative escaped mass and the planet mass constants, which is why escape is on here.
- **Physics of the recorded run, checked independently.** Per-element mass closure holds to 2e-16, cumulative escaped mass matches the elemental drawdown to 1.7e-9, per-element escape rates sum to the total to 1.5e-16, and temperature, pressure, melt fraction and atmospheric mass stay in bounds throughout.

## Known limits

- **The reference was recorded on macOS/arm64 and has not been reproduced on Linux.** The tolerance is set for the difference between one platform's maths library and another's, and nothing on this configuration's path iterates to a tolerance (the surface state is fixed rather than solved for), so there is no convergence tolerance to vary. The first nightly is the test of that; if it fails, the report names the columns and by how much.
- The comparison needs a real run, so it is nightly rather than per-pull-request.
- A field the configuration does not name takes its schema default, and changing that default moves the trajectory with the configuration's bytes unchanged. That is the intended reading: the default is part of the behaviour under test.
- **`MIN_COLUMNS_COMPARED` has to move when the helpfile grows.** It is a floor on how much of the helpfile the comparison covers, and a floor set against a narrower helpfile stops discriminating as columns are added. It sits at about 86% of what the configuration writes today.
- Found while writing this and left alone as pre-existing: a `Proteus()` construction that raises after `init_directories()` leaks its scratch directory, since no reference to the object escapes `__init__`. Worth its own issue.

## Checklist


nichollsh:
Nice idea - thanks, @timlichtenberg. I can certainty see a lot of utility and value in being having built-in functionality for checking that PROTEUS runs are deterministic and reproducible. 

However, it's not clear to me why this exists within the main PROTEUS source code. Specifically: why did you decide that this should be a new set of functions in `src/proteus/utils/trajectory.py` rather than part of the test suite? Or maybe as a tool in `tools/`? Currently, this functionality is split between all three 'parts' of the codebase. 

Do you see wider applications for this beyond the test suite and documentation? For example, maybe this relates to recent Morrigan developments, which (as I understand it) may lead to multiple PROTEUS runners.

timlichtenberg:
> why did you decide that this should be a new set of functions in `src/proteus/utils/trajectory.py` rather than part of the test suite? Or maybe as a tool in `tools/`? Currently, this functionality is split between all three 'parts' of the codebase.

You are right, and I have moved the whole thing into `tests/`. There is no longer a leg of this in `src/` or in `tools/`.

The precedent decided it. `tests/helpers/_smoke_invariants.py` is a verification library of about the same size, imported by seven integration tests and unit-tested in place, and this is the same shape of thing. My reason for `src/` was that `tools/` is not importable, which is true but answers the wrong question: `tests/helpers/` is importable, and it is already where the suite keeps this kind of code.

Recording is a pytest option now rather than a script, so the recorder leg is gone rather than relocated:

```bash
pytest tests/integration/test_integration_golden_run.py                  # compare
pytest tests/integration/test_integration_golden_run.py --record-golden  # record
```

On wider applications, I no longer think they justify keeping the library outside the suite. The comparison is general in principle and Morrigan comparing two runners is the obvious use, but nothing outside the test suite calls it today. If a second consumer appears it can move to `src/` behind a `proteus` subcommand, and two callers will have settled the API by then. Building for that now was the mistake you were pointing at.

Two things worth a look. Recording reports what it accepted through the skip reason rather than printing it, since a `print` is swallowed unless the caller also passes `-s`. And the branch is rebased onto current `main`: the rock-vapour work took the helpfile from 459 columns to 756, so the reference is recorded again. Every column that existed before reproduces exactly and the run still takes 56 steps, so only the schema is wider, not the trajectory. `MIN_COLUMNS_COMPARED` moved with it, since at its old value it would have accepted the comparison losing nearly half the helpfile.

docs/Explanations/test_framework.md
```diff
@@ -189,3 +189,44 @@ assert result == pytest.approx(expected, rel=1e-3)
...
+  against itself passes forever. The golden-run test carries a guard that
+  perturbs the trajectory by a stated amount and requires the comparison to
+  report it; the resume test instead asserts that its first leg really did
+  stop short, so the two runs it compares are not the same run.
+
+The tolerance sits in a band, not at a point: loose enough to absorb the
+last-bit differences between one platform's maths library and another's,
+tight enough that no change of behaviour fits underneath. Tests pin both ends
+by size rather than as multiples of the tolerance, since a test written in
+terms of the constant follows it wherever it is moved and pins nothing.
```
nichollsh: Great to introduce this new type of test. Some suggested phrasing changes...


```suggestion
The core of PROTEUS' testing suite compares what the model is found to do, given some current state, against some pre-defined expectation. Here, we outline an additional type of testing: 'trajectory-parity'. These perform a relative test between what the code did _previously_ and what it does _now_. 

There are currently two main trajectory-parity tests. 

1. The resume test asserts that a run stopped and restarted follows the same trajectory as one that ran straight through, which is the strongest thing that can be said about resuming. 
2. The golden-run test asserts that a fixed configuration follows the trajectory recorded beside it. This ensures that any changes to the code which claim to have 'changed nothing' actually verifiable.

The second is built on the comparison in `tests/helpers/_trajectory.py`; the first compares its two runs directly. Three properties make these comparisons worth having:

- **The tests cover everything a PROTEUS run writes.** A curated list of quantities (in remainder of the test main suite) covers only the failures somebody anticipated. Comparing every output columns catches other problems.
- **Undefined values are a state, not noise.** Screening non-finite values out before comparing skips precisely the columns a change corrupted into NaN. The resume test screens its reference side instead, which is sound since it compares a resumed run against an uninterrupted one, so screening the uninterrupted side leaves a corrupted resumed value to fail the comparison rather than skipping it.
- **It has to be able to fail.** A parity assertion that compares a run against itself passes forever. The golden-run test implements a guard that perturbs the trajectory by a stated amount and requires the comparison to report it; the resume test instead asserts that its first leg really did stop short, so the two runs it compares are not the same run.

These are implemented to a set tolerance, defined to be loose enough to absorb the numerical differences between platforms, but tight enough to catch changes of behaviour.
```

tests/integration/golden_run.tsv
nichollsh, line 1: The formatting of this file is very strange. Why not use the usual helpfile format? Not a blocking problem, since there's no "correct" way to store data, but it's just odd.

Why are some "const" and others "series"?

nichollsh (changes requested):
Thanks for this, @timlichtenberg. I think moving this into the test suite really is the right move - it would be great to keep this here, even with further integration of Morrigan. But, of course, happy to keep discussing this in the future.

I've run the commands
```bash
pytest tests/integration/test_integration_golden_run.py --record-golden
pytest tests/integration/test_integration_golden_run.py
```
and confirmed that the new test types are passing on my laptop.

Choosing `Tmagma` is the perturbed column seems like a reasonable choice.

I also manually enabled plotting for the golden_run config, and checked the plots that it produces. These seem to show reasonable behaviour, so am happy to confirm that this test is appropriate for probing the wider suite. Images below (also for posterity).

Image

Image

---

This all being said, I do have one suggestion and a separate question below.

timlichtenberg:
Thanks for running it both ways and for checking the plots; good to have them on the thread.

Docs suggestion taken. I fixed three small things in it while applying: "changed nothing' actually verifiable" reads "are actually verifiable", "in remainder of the test main suite" reads "in the remainder of the main test suite", and "every output columns" is singular. I also wrapped it to the width the rest of the page uses.

On the file format, the question is fair and the honest answer is that the file failed to explain itself, so I have made it do that. The header now says why it is laid out this way.

It is the transpose of the helpfile. The reference is a file we read in diffs, and that is the only thing the layout is chosen for. Laid out as a helpfile, a quantity that shifts changes all 56 rows, and each of those rows is 756 comma-separated numbers, so the reviewer's job is to find which field moved. Transposed, the same change is one line, and the line names the quantity.

`const` against `series` is the same argument applied to size. 650 of the 756 columns hold a single value for the whole run, so writing them once instead of 56 times takes the file from 275 kB to 127 kB. Nothing is lost: a column that varies at all, including one that becomes undefined part way through, is written as `series` with a value per row.

Happy to switch it to the helpfile format if you would still rather have the familiar one; it is a small change and the comparison does not care. I would rather keep this one for the diff behaviour, but it is your call.

nichollsh:
Makes sense! Thank you for clarifying this; as I now understand it, the `const` and `series` flags are effectively a form of data compression. This is fine with me - no need to change to the helpfile csv format.

Thanks for implementing my doc suggestions.

nichollsh (dismissed):

timlichtenberg:
Rebased onto main, which cost your approval, @nichollsh. Sorry about that. Here is everything that changed since you looked.

#802 landed while this was open and added `bol_scale` to the helpfile, so the run writes a column the recorded reference does not have. Running the comparison before touching the reference reported exactly one difference, `bol_scale` produced by the run and absent from the reference, with the other 756 columns agreeing over 56 unchanged rows. The schema grew by one column and no value moved. The reference now carries it, constant at 1.0, since nothing scales the bolometric flux on this configuration.

Two comments were wrong and are corrected. The escape module's comment in the configuration quotes how many columns move with escape on and off, measured against a narrower helpfile than the one written today; re-measured by running it both ways, it is 84 to 106 out of 757. And the paragraph above the perturbation guard ran two different measurements together: scaling `dTdt` is reported on a different column from the one perturbed, while the guard moves a recorded column directly and asserts the difference lands on that same column, so the first sentence read as a claim about the guard that the guard asserts the opposite of.

Editing the configuration changes its digest, so the reference is recorded once more. All 757 columns reproduce and the only other line to change is the digest.

Three fresh runs still agree bit for bit on all 757 state columns.

nichollsh:
No worries about the force-merge. I appreciate you clarifying these differences, though. Exciting to see more development!

Issue #708, opened 2026-06-16
logger is not configured at all times, logs get lost


egpbos:
When running PROTEUS through CLI (e.g. `proteus start`), the logger only gets initialized inside `PROTEUS.start()`, but before that the logger already gets used via the directory setting functions called from `PROTEUS.__init__()`. This is one example, I imagine there could be others. The main point is that logging should be activated project-wide at every entry point.

nichollsh:
This also seems to impact the `grid-summarise` and `grid-pack` commands, since they do not explicitly setup the logger.

PR #792, opened 2026-07-28
Close the fwl logger handlers after each test (#791)

### LINKED ISSUE #791: TestSetupLogger fails when TMPDIR is on an NFS mount
## Problem

`TestSetupLogger` in `tests/utils/test_logs.py` fails on machines whose `TMPDIR` points to an NFS network file system mount. This causes some of the  current tests to error. The failure is in `tempfile.TemporaryDirectory` cleanup, not in an assertion, so every affected test reports as failed even though its checks passed.

```
OSError: [Errno 39] Directory not empty: '/data/hen28/tmpdir/tmpz84o3cmd'
  .../python3.12/shutil.py:662: in _rmtree_safe_fd
```

This is because each test calls `setup_logger(logpath=..., logterm=False)` inside a
`with tempfile.TemporaryDirectory() as tmpdir:` block and never closes the
`FileHandler` that `setup_logger` attaches to the `fwl` logger. The logfile is
therefore still open when the context manager exits.

On a local filesystem, unlinking an open file is invisible to `rmdir`. On NFS the
server cannot drop an open file, so the client renames it to a `.nfsXXXX`
entry that persists until the descriptor closes. `rmdir` then sees a non-empty
directory and raises `ENOTEMPTY`.

CI does not catch this because the runners use local-disk temp space.

## Suggestion

Close and detach the handlers `setup_logger` registers before the temporary
directory is removed. Add a fixture that tears down the `fwl` logger after each test:

```python
@pytest.fixture(autouse=True)
def _close_fwl_handlers():
    yield
    logger = logging.getLogger('fwl')
    for handler in list(logger.handlers):
        handler.close()
        logger.removeHandler(handler)
```

## People

@timlichtenberg @egpbos

egpbos:
## Description
`TestSetupLogger` in `tests/utils/test_logs.py` errors on machines whose `TMPDIR` is an NFS mount. The tests call `setup_logger(logpath=..., logterm=False)` with a path inside an inline `with tempfile.TemporaryDirectory()`, but the `FileHandler` that `setup_logger` attaches to the shared `fwl` logger is never closed, so the logfile is still open when the directory is torn down.

On a local filesystem an unlinked-but-open file is invisible to `rmdir`. On NFS the server keeps the open file as a `.nfsXXXX` entry until the descriptor closes, so `rmdir` sees a non-empty directory and raises `OSError: [Errno 39] Directory not empty` during cleanup. Because the failure is in `TemporaryDirectory` teardown rather than an assertion, every affected test reports as failed even though its checks passed. CI does not catch this because the runners use local-disk temp space.

Fix: add a module-level `tmpdir_that_closes_logger_files` fixture that owns the `TemporaryDirectory`, yields it to the test, and after the test returns closes and detaches the `fwl` handlers before its own `with` block removes the directory. Every `setup_logger` / `bootstrap_logger` test that writes a logfile now takes this fixture instead of opening its own `TemporaryDirectory`. This covers the `TestSetupLogger` tests, the two `fmt` tests, and the two `TestBootstrapLogger` tests that write a logfile (the latter three were added by #708 / #776).

Note on the design:
- A plain post-test teardown fixture does not fix this: an inline `with tempfile.TemporaryDirectory()` inside a test removes the directory when the method returns, which is before any fixture teardown runs. The fixture therefore has to own the directory so the close happens between the test and the removal.
- `TestBootstrapLogger`'s existing `pristine_fwl` fixture is kept as-is. It provides a handler-free logger at setup for that class's preconditions; it cannot fix the cleanup itself (same too-late ordering), so its two file-writing tests use the shared fixture for the directory.

This branch is rebased onto `main` after #776 merged, so it applies the fix to the tests that PR added as well.

Closes #791

## Validation of changes
- `tests/utils/test_logs.py`: 49 tests pass locally; `ruff check` and `ruff format --check` clean.
- Reproduced and fixed on real NFS storage (a Kapteyn workstation with `$HOME` on NFS), with `TMPDIR` pointing at the NFS mount, running `TestSetupLogger` and `TestBootstrapLogger`:
  - Pre-fix (`main`, post-#776): `17 failed, 5 passed`, with `OSError: [Errno 39] Directory not empty` in `TemporaryDirectory` cleanup. The failures include the two `TestBootstrapLogger` file-tests added by #776. The 5 passes are the tests that never open a FileHandler (the invalid-level test and the console-only bootstrap tests).
  - With this fix: `22 passed` on the same NFS `TMPDIR`.

Test configuration: reproduction on a Linux NFS mount (the reported condition); local development checks on macOS with the PROTEUS conda environment (Python 3.12).

## Checklist

### DIFF
--- a/tests/utils/test_logs.py
+++ b/tests/utils/test_logs.py
@@ -28,6 +28,29 @@ from proteus.utils.logs import (
 pytestmark = [pytest.mark.unit, pytest.mark.timeout(30)]
 
 
+@pytest.fixture
+def tmpdir_that_closes_logger_files():
+    """Yield a temp dir, then close the 'fwl' file handlers before removing it.
+
+    setup_logger attaches a FileHandler to the shared 'fwl' logger. If the logfile
+    is still open when the directory is torn down, cleanup fails on an NFS TMPDIR:
+    the server keeps the open file as a .nfsXXXX entry, so rmdir raises OSError
+    [Errno 39]. Owning the TemporaryDirectory here and closing the handlers after
+    the test yields but before this with-block removes the directory releases the
+    descriptors first, so cleanup succeeds on NFS as well as local disk. A plain
+    post-test teardown fixture would be too late: an inline with-block
+    TemporaryDirectory inside a test removes the directory when the method returns,
+    before any fixture teardown runs. Shared by the setup_logger and
+    bootstrap_logger tests that write a logfile. See issue #791.
+    """
+    with tempfile.TemporaryDirectory() as tmpdir:
+        yield tmpdir
+        logger = logging.getLogger('fwl')
+        for handler in list(logger.handlers):
+            handler.close()
+            logger.removeHandler(handler)
+
+
 class TestStreamToLogger:
     """Test suite for StreamToLogger stream-to-logger redirection.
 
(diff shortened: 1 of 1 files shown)


nichollsh:
This depends on https://github.com/FormingWorlds/PROTEUS/pull/776 being merged first

egpbos:
Merged #776, will now rebase on main and update to also cover the added tests there.

egpbos:
Done, ready for review!

nichollsh (approved):
Great! Works for me on the IoA cluster, where the issue with NFS was first noted. Happy for this to be merged.

nichollsh:
@egpbos with #794 being merged, is this now already ready?

egpbos:
Yes! It seems I didn't finish the comment above where I wanted to say I would rebase on main and then apply the fix to the added tests from the other PR too. That's what I did in the new commit.

egpbos:
Oh wait you referred to the faster CI PR, I was referring to the previous logging one. The faster CI one is not related to this one.

Issue #793, opened 2026-07-28
speed up "fast PR checks"


egpbos:
The "fast PR checks" CI workflow seems to have been degrading in speed over the past months; whereas halfway May it still ran in 4-5 minutes sometimes (see https://github.com/FormingWorlds/PROTEUS/actions/workflows/ci-pr-checks.yml?page=11&query=is%3Asuccess) it now needs 13-14 minutes on both Mac and Linux, and most of that seems to be in downloading data, which I wonder whether this needs to be tested across both Mac and Linux at all. Is something going wrong with caching? Do we need all those downloads at all (see this latest run https://github.com/FormingWorlds/PROTEUS/actions/runs/30345820266)?

I can look into this, please let me know if there's anything you know about this that I should take into account.

egpbos:
I dug into run [30345820266](https://github.com/FormingWorlds/PROTEUS/actions/runs/30345820266) and compared it against a mid-May run (26018924667, 2026-05-18). Findings below.

## Where the time goes

Almost all of the growth is in the "Set up PROTEUS environment" composite step. The unit tests themselves are still ~90 s.

| Phase | May 18 | Jul 28 | Δ |
|---|---|---|---|
| **Linux job total** | 5m39s | 16m30s | +10m51s |
| Set up PROTEUS env | **4m32s** | **14m11s** | **+9m39s** |
| Run unit tests | 32s | 94s | +62s |
| **macOS job total** | 5m51s | 11m50s | +5m59s |

Inside the Jul 28 Linux setup, the growth is three items (all read off the step logs):

| Sub-step | May | Jul | What it is |
|---|---|---|---|
| **AGNI spectral-data fetch** | absent | ~226 s | Always-run step added in `4e96c3cd`. Dominated by a single `gases.zip` (~146 s). |
| **"Download smoke data"** | 19 s | ~207 s | 58 reference files fetched one at a time through fwl-io (57 mass-radius from `zenodo.15727899` + 1 DACE exoplanet), ~3 s latency each = ~190 s. |
| **apt system packages** | 19 s | ~195 s | Grew 10x. This one is uncertain: it could be runner/mirror variance or `needrestart`, and it is only one data point, so I would not bank on it as a structural cause yet. |

Both runners pay all three: the macOS job shows the same 58 downloads and the same AGNI fetch.

## Two things are going wrong with caching

[...]

PR #794, opened 2026-07-28
Stop re-downloading cached FWL_DATA and AGNI data on every CI run

egpbos:
## Description

Two cheap caching fixes for the fast PR checks, addressing part of #793. Both target the "Set up PROTEUS environment" step, which is where nearly all of the recent slowdown lives (Linux setup grew from ~4.5 min in May to ~14 min now; the unit tests themselves are still ~90 s). Both the Linux and macOS legs were affected.

**FWL_DATA cache (solution 1 in #793).** The cache key hashed only `.github/data-manifest.yaml`, which has not changed since the exoplanet and mass-radius reference data were moved to fwl-io (#782). With a constant key, `actions/cache` exact-hits and never re-saves, so the 58 fwl-io reference files (~190 s of serial per-file downloads) were absent from the frozen tree, re-downloaded on every run, and thrown away at job end. This PR hashes `src/proteus/data/proteus_manifest.toml` (the real fwl-io pins) into the key as well, so:

- a Zenodo re-pin now busts the cache and forces a fresh save, instead of silently re-downloading forever against a frozen exact-hit cache;
- the key changes once with this PR, which seeds a fresh, complete cache; the `restore-keys` prefix lets that save build on the previous tree rather than re-downloading everything.

**AGNI spectral-data fetch (solution 2 in #793).** The fetch ran unconditionally, re-pulling ~200 s of data (dominated by the thermodynamics archive) even when the warm AGNI cache already held a complete `res/` tree. It is now guarded by the step's existing `res_complete` check: skip when every required artifact is present and non-empty, and fall through to the existing retry/repair loop otherwise. The repair path the unconditional fetch was added for (a partial or zero-byte `res/` from a green-but-incomplete download) is preserved, because such a tree fails `res_complete` and still triggers the fetch.

Solutions 3 (skip the real-binary toolchain for the mocked unit job) and 4 (decouple downloads from the OS matrix) from #793 are deliberately out of scope here; they need verification and design first.

## Validation of changes

Confirmed on CI (both runners): a warm re-run downloads 0 reference files and skips the AGNI fetch, cutting the job from 16m30s to 6m39s on Linux and 11m50s to 5m19s on macOS. Full evidence and the two-run explanation are in [this comment](https://github.com/FormingWorlds/PROTEUS/pull/794#issuecomment-5105593962).

Test configuration: CI-config only (no Python touched); the two guarding unit tests (`test_ci_setup_installs_every_declared_extra`, `test_ci_config_never_pins_user`) pass locally.

## Checklist

### DIFF
--- a/.github/actions/setup-proteus/action.yml
+++ b/.github/actions/setup-proteus/action.yml
@@ -527,11 +527,20 @@ runs:
     # ------------------------------------------------------------------
     # 9. FWL_DATA cache + smoke / full data download
     # ------------------------------------------------------------------
+    # The key hashes BOTH dataset descriptors: .github/data-manifest.yaml (the
+    # older download() datasets: spectral files, stellar spectra, tracks) and
+    # src/proteus/data/proteus_manifest.toml (the fwl-io pins for the exoplanet
+    # + mass-radius reference data). Keying on the fwl-io manifest as well means
+    # a Zenodo re-pin busts the cache and forces a fresh save. Without it, files
+    # pinned after the cache was first frozen re-download on every run: the key
+    # stays constant, actions/cache exact-hits and never re-saves, so the stale
+    # tree can never pick them up. The restore-keys prefix still lets a fresh
+    # save build on the previous tree instead of re-downloading everything.
     - name: Cache FWL_DATA
       uses: actions/cache@v5
       with:
         path: ~/fwl_data
-        key: fwl-data-${{ runner.os }}-${{ hashFiles('.github/data-manifest.yaml') }}
+        key: fwl-data-${{ runner.os }}-${{ hashFiles('.github/data-manifest.yaml', 'src/proteus/data/proteus_manifest.toml') }}
         restore-keys: |
           fwl-data-${{ runner.os }}-
 
--- a/.github/actions/setup-proteus/action.yml
+++ b/.github/actions/setup-proteus/action.yml
@@ -387,6 +387,18 @@ runs:
           fi
           return $rc
         }
+        # Skip the fetch when a warm cache already holds a complete res/ tree.
+        # res_complete asserts every required artifact is present AND non-empty
+        # (a zero-byte file from a green-but-empty fetch fails the -s test), so a
+        # tree that passes needs no repair and re-pulling it is pure download
+        # cost (~200 s, dominated by the thermodynamics archive). An incomplete
+        # or partially-written tree fails the check and falls through to the
+        # retry loop below, which overwrites and repairs it, so the repair path
+        # the unconditional fetch was added for is preserved.
+        if res_complete 2>/dev/null; then
+          echo "AGNI basic data already complete in res/ (warm cache); skipping fetch."
+          exit 0
+        fi
         fetch="$GITHUB_WORKSPACE/AGNI/src/get_data.sh"
         delays=(10 30 60)
         attempts=$(( ${#delays[@]} + 1 ))
--- a/.github/data-manifest.yaml
+++ b/.github/data-manifest.yaml
@@ -1,9 +1,10 @@
 # PROTEUS data manifest -- explicit version pins for the datasets CI caches.
 #
 # The CI composite action (.github/actions/setup-proteus/action.yml) hashes
-# this file to key the FWL_DATA cache. The cache invalidates only when this
-# manifest changes; CI runs that don't change the manifest reuse the
-# previously cached FWL_DATA tree.
+# this file together with src/proteus/data/proteus_manifest.toml (the fwl-io
+# dataset pins) to key the FWL_DATA cache. The cache invalidates when either
+# descriptor changes; CI runs that change neither reuse the previously cached
+# FWL_DATA tree.
 #
 # Bump a version string in this file when:
 #   * a new dataset becomes a hard requirement of a tier the CI runs;


nichollsh:
Thanks for this, @egpbos. Could you clarify what happens when the AGNI data are updated? If the AGNI `res/` folder is populated with older files, restored from a cache, does this still re-trigger the download or do we need to wait until the `fwl-io` does this for us?

That is: what happens when this record is updated? https://zenodo.org/records/21390786

egpbos:
## Confirmed on CI (both runners)

Ran this branch twice: a seeding run that reseeds the cache once, then a warm re-run that is the real test.

| Evidence | Linux | macOS |
|---|---|---|
| FWL_DATA cache (warm run) | exact hit on new key `068e80…` | exact hit |
| Reference-file downloads (warm run) | **0** | **0** |
| AGNI fetch | skipped (`res/` verified complete) | skipped |
| Job wall time, before | 16m30s | 11m50s |
| Job wall time, warm run | **6m39s** | **5m19s** |

The warm re-run passed all checks (green), so the tests still pass on the warm-cache path, not just the setup step.

### What the two runs show

- **The frozen-cache bug is fixed and now self-heals.** Seeding run: the new key had no exact match, so it restore-key-matched the old stale tree, downloaded the 58 fwl-io reference files once, and logged `Cache saved with key: fwl-data-Linux-068e80…`. Warm run: exact hit on that key, and `Download smoke data` reported 0 downloads.
- **fwl-io is genuinely idempotent on a populated tree.** This was the open question from #793. The 0-download warm run answers it: files already present with a matching checksum are skipped. So the cache-key fix alone was sufficient and no change to the fetch code was needed.
- **The AGNI guard works.** It skipped the fetch even on the seeding run, because the warm AGNI cache already held a complete `res/`. This removes ~226 s per run per runner (mostly the thermodynamics archive).

Net: fast PR checks are back to roughly the pre-June baseline, saving about 10 min per run on Linux and 6.5 min per run on macOS, on every PR.

egpbos:
> Thanks for this, @egpbos. Could you clarify what happens when the AGNI data are updated? If the AGNI `res/` folder is populated with older files, restored from a cache, does this still re-trigger the download or do we need to wait until the `fwl-io` does this for us?
> 
> That is: what happens when this record is updated? https://zenodo.org/records/21390786

Good question. It doesn't trigger on that. The cache action right now only looks at two files: .github/data-manifest.yaml and src/proteus/data/proteus_manifest.toml. On main it only looks at .github/data-manifest.yaml, but it looks like that one is not kept up to date, which is the reason the cache was no longer being updated. I think indeed if all data starts using the fwl-io manifest then the CI cache should always be up to date. But I must admit that I'm not sure whether that is the plan (and even if it is, whether it will happen soon).

Is it necessary for the fast PR tests to have the _latest_ data? Or the further reaching question: is (all) data necessary at all for fast PR tests?

nichollsh:
> is (all) data necessary at all for fast PR tests?

No, it isn't. If we are skipping a lot of the physics for the fast PR tests then we likely don't need all of the data. 

>  think indeed if all data starts using the fwl-io manifest then the CI cache should always be up to date

So in that case, perhaps a lot of the problems highlighted here will be resolved/fixed when fwl-io is fully implemented across all modules? Do we want to address this issue now, then, or wait a while?

egpbos:
> So in that case, perhaps a lot of the problems highlighted here will be resolved/fixed when fwl-io is fully implemented across all modules? Do we want to address this issue now, then, or wait a while?

Well, the first part of this PR (particularly [this line](https://github.com/FormingWorlds/PROTEUS/pull/794/changes#diff-20ccc7688611e9f7131407be98be42556b9570438f1baf486ea462d33ba4726eR555), I should have made this two commits, sorry for that) makes sure the manifest triggers cache-refresh, so I think we can address that part now already for sure. The second part (preventing AGNI redownload if the cache is fine) we could leave out for now, although it does save 200+ seconds every CI run if we do it now.

egpbos:
I split the commit in two now. To clarify my previous comment:

- 0e06eb5 I think is a fix that we should do anyway, it fixes the currently broken cache.
- d7a4ed9 is the AGNI part that we need to decide on doing now or waiting on AGNI compatibility with the fwl-io manifest.

egpbos:
Discussed this offline with @timlichtenberg. The conclusion we reached on the AGNI commit is that it's probably best to merge it like it is now. Win: speeds up CI now by 3-4 minutes, quite noticeable. Costs: a bit of maintenance (just remember to remove it after AGNI data has switched to fwl-io) and it will not automatically update to new Zenodo releases. Mitigation for that: delete the cache manually, that will always trigger a new download. Do you think this is workable @nichollsh? Also this part would only be until AGNI data has switched to fwl-io.

Btw, also good to know: fwl-io development is currently blocked because Tim is waiting for admin rights to make our Dataverse mirrors public (they are private by default it seems). This may take until after the summer, hard to say.

nichollsh:
That sounds good to me! I think if we can already benefit from the changes you've made now, then that's great. We can make further developments in the future and they're all iterative/positive. 

It's still not entirely clear to me how AGNI/Obliqua can best utilise fwl-io. We should discuss this in more detail, since I'd like to get it right.

egpbos:
> It's still not entirely clear to me how AGNI/Obliqua can best utilise fwl-io. We should discuss this in more detail, since I'd like to get it right.

What I understand is that @timlichtenberg will get around to this at some point, he's including things into fwl-io step by step. I think the plan is to just figure out step by step how to do it? Your comment https://github.com/nichollsh/AGNI/issues/185#issuecomment-5055626031 will probably be addressed that way, but maybe good if we put it on the Interra board to make sure we discuss it briefly next week?

I still need an approval on this PR before merging, if you or @timlichtenberg have the time :)

nichollsh (approved):
Apologies for the delay in reviewing this. Thanks for identifying this performance issue @egpbos. This PR seems to take a good step toward addressing the problem, and we will just need to manually delete the cache if the AGNI reference data update. This shouldn't be too often/numerous before we get around to the complete fwl-io implementation.

I am good with this being  merged as-is, and we can continue to iterate on the CI performance (particularly relating to the AGNI data) in future PRs.

PR #799, opened 2026-07-30
Apply the orphan-key check wherever a configuration is loaded

timlichtenberg:
## Description

Builds on the orphan-key tooling from #705. That added the walk that compares a config against the schema; this puts it where every config load goes through, and adds the one fault it cannot see.

Config parameters all have defaults and the parser drops any key it does not recognise, so a misspelled or outdated option silently does nothing and the run carries on with the default. The check from #705 was called from `Proteus.__init__` only, so `proteus grid`, the download commands and the install and update paths all bypassed it. It now lives in `read_config_object`.

The grid is the worst of it, and not just because it was one more caller. Case files are written out from the parsed base config, so an unrecognised key in the base is erased before any case file exists and the per-case check sees clean files. Misspell `maximum` in `tests/grid/base.toml` and every case in the ensemble runs on the 1e7 default instead of the 3e7 asked for, silently.

Two other things:

- The key names never reached the user. They went out through `log.error` calls that run before a handler is attached, so the `RuntimeError` said only that unknown keys existed.
- `[[planet]]` instead of `[planet]` is spelled correctly, so comparing names finds nothing wrong, while the whole section is dropped and every parameter inside it reverts to its default. Caught now, at any depth.

## Validation of changes

macOS, Python 3.12. Unit tier green.

- Every config in `input/` and `tests/` still loads, so nothing anyone already has stops working. Grid and inference definition files use their own schemas and do not go through this loader.
- Configs written back out by `Config.write()` and by `Grid.write_config_files()` reload under the new rule, so grids and saved run configs are unaffected.
- New tests cover both faults at the loader, at the runner and through the CLI.

One thing worth a look: the check runs where `read_config_object` runs, which is not everywhere. `proteus infer` only checks that its reference config exists, and `tools/migrate_config_v2_to_v3.py` validates without it. I will file those two separately.

## Checklist

### DIFF
--- a/src/proteus/config/orphans.py
+++ b/src/proteus/config/orphans.py
@@ -10,13 +10,25 @@ will be ignored if not mapped by the parser.
 from __future__ import annotations
 
 import logging
+import types
 import typing
+from pathlib import Path
 
 import attrs
 
 from ._config import Config
 
 log = logging.getLogger('fwl.' + __name__)
+
+
+class UnknownConfigKeyError(ValueError):
+    """Raised when a config file carries keys outside the schema.
+
+    A distinct type so that callers can present this to the user as a
+    configuration problem without also catching unrelated failures. It derives
+    from ValueError, which is what every other config rejection raises, so a
+    caller that does not care about the distinction needs no change.
+    """
 
 
 def _extract_attrs_class(hint: type) -> type | None:
@@ -45,6 +57,33 @@ def _extract_attrs_class(hint: type) -> type | None:
     return None
 
 
+def _expects_single_table(hint: type) -> bool:
+    """Whether *hint* puts exactly one table at a field, rather than several.
+
+    A field holding one nested class is written as ``[name]`` and nothing else
+    is valid there. A field holding a container of them would be written as
+    repeated ``[[name]]`` tables, where a list is the intended shape rather
+    than a mistake. The schema declares no such field today, so this only
+    keeps a later one from being refused.
+
+    Parameters
+    ----------
+    hint:
+        Type hint to inspect.
+
+    Returns
+    -------
+    bool
+        True when a single table is the only shape the field accepts.
+    """
+    origin = typing.get_origin(hint)
+    if origin is None:
+        return isinstance(hint, type) and attrs.has(hint)
+    if origin in (types.UnionType, typing.Union):
+        return any(isinstance(arg, type) and attrs.has(arg) for arg in typing.get_args(hint))
+    return False
+
+
 def _collect_orphan_keys(data: dict, cls: type, path: str = '') -> list[str]:
     """Recursively collect TOML keys that have no matching field in *cls*.
 
@@ -64,9 +103,32 @@ def _collect_orphan_keys(data: dict, cls: type, path: str = '') -> list[str]:
         *data* but are not declared fields of *cls* or any nested attrs class.
     """
 
+    return _collect_key_problems(data, cls, path)[0]
+
+
+def _collect_key_problems(data: dict, cls: type, path: str = '') -> tuple[list[str], list[str]]:
+    """Recursively collect the keys in *data* that *cls* cannot accept.
+
+    Parameters
+    ----------
+    data:
+        Raw TOML sub-dict to inspect.
+    cls:
+        attrs-decorated class to compare against.
+    path:
+        Dotted key prefix for building human-readable paths in error messages.
+
+    Returns
+    -------
+    tuple[list[str], list[str]]
+        Two lists of dotted key paths, in file order. The first holds names the
+        schema does not declare. The second holds sections the schema declares
+        as a table but which the file supplies as something else.
+    """
+
     # If the class is not an attrs class, then we cannot inspect it.
     if not attrs.has(cls):
-        return []
+        return [], []
 
     # Get the field names of the attrs class
     field_names = {f.name for f in attrs.fields(cls)}
@@ -80,6 +142,7 @@ def _collect_orphan_keys(data: dict, cls: type, path: str = '') -> list[str]:
         hints = {}
 
     orphans: list[str] = []
+    mistyped: list[str] = []
 
     # Loop through keys in the raw dict.
     for key, value in data.items():
@@ -89,20 +152,31 @@ def _collect_orphan_keys(data: dict, cls: type, path: str = '') -> list[str]:
         # Check if the key is in the attrs class. If not, add to orphans.
         if key not in field_names:
             orphans.append(full_path)
+            continue
+
+        # Nothing further to check unless the schema puts a nested class here.
+        nested_cls = _extract_attrs_class(hints.get(key))
+        if nested_cls is None:
+            continue
 
         # If this is a dict, then we need to go deeper.
-        elif isinstance(value, dict):
-            nested_cls = _extract_attrs_class(hints.get(key))
+        if isinstance(value, dict):
+            sub_orphans, sub_mistyped = _collect_key_problems(value, nested_cls, full_path)
+            orphans.extend(sub_orphans)
+            mistyped.extend(sub_mistyped)
+        elif _expects_single_table(hints.get(key)):
+            # The schema puts one table here and the file supplies something
+            # else, most often because the section was written as [[name]]
+            # rather than [name]. The name still matches a field, so nothing
+            # above notices, while structuring discards the section whole and
+            # every parameter inside it falls back to its default.
+            mistyped.append(full_path)
 
-            # Recursion on this function
-            if nested_cls is not None:
-                orphans.extend(_collect_orphan_keys(value, nested_cls, full_path))
-
-    return orphans
+    return orphans, mistyped
 
 
-def check_config_orphan_free(raw_dict: dict) -> bool:
-    """Detect if *raw_dict* contains keys that Config schema doesn't define.
+def find_orphan_keys(raw_dict: dict) -> list[str]:
+    """List the keys in *raw_dict* that the Config schema does not define.
 
     Parameters
     ----------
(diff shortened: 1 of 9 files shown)


src/proteus/config/orphans.py
```diff
@@ -45,6 +57,33 @@ def _extract_attrs_class(hint: type) -> type | None:
...
+    """
+    origin = typing.get_origin(hint)
+    if origin is None:
+        return isinstance(hint, type) and attrs.has(hint)
+    if origin in (types.UnionType, typing.Union):
+        return any(isinstance(arg, type) and attrs.has(arg) for arg in typing.get_args(hint))
+    return False
+
+
 def _collect_orphan_keys(data: dict, cls: type, path: str = '') -> list[str]:
```
egpbos: Are this function and `find_orphan_keys` still used anywhere? If not, maybe just remove them, they seem superfluous here at least.

src/proteus/config/orphans.py
```diff
@@ -64,9 +103,32 @@ def _collect_orphan_keys(data: dict, cls: type, path: str = '') -> list[str]:
         *data* but are not declared fields of *cls* or any nested attrs class.
     """
 
+    return _collect_key_problems(data, cls, path)[0]
+
+
+def _collect_key_problems(data: dict, cls: type, path: str = '') -> tuple[list[str], list[str]]:
```
egpbos: This could be simplified by leaving out the unused cls and path arguments, just hardcoding this to the `Config` class that we actually use and also dropping the private `_` prefix and just making this one `find_key_problems` directly. Perhaps I missed the reason for this more generic approach, though.

tests/config/test_config.py
```diff
@@ -65,6 +66,156 @@ def test_read_config_returns_dict():
...
+    dropped = read_config_object(array, strict=False)
+    assert dropped.planet.mass_tot == pytest.approx(1.0)
+    assert dropped.planet.mass_tot != pytest.approx(3.7)
+
+
+@pytest.mark.unit
+def test_read_config_object_without_strict_applies_the_default(tmp_path):
+    """With strict off the unknown key is dropped and the default takes over.
+
+    This is exactly the outcome strict loading exists to prevent, and it is the
```
egpbos: Why would we want non-strict loading to be kept as an option?

egpbos (dismissed):
Very nice addition! Good test coverage.

I seem to remember that we did discuss this somewhere, @nichollsh I think flagged this issue at some point too, but maybe the discussion was offline.

A few suggestions below to possibly make this slightly more compact.

timlichtenberg:
> Very nice addition! Good test coverage.
> 
> I seem to remember that we did discuss this somewhere, @nichollsh I think flagged this issue at some point too, but maybe the discussion was offline.
> 
> A few suggestions below to possibly make this slightly more compact.

Agree, we discussed this multiple times already. I couldn't find an issue to link this PR to though.

nichollsh:
Could you clarify how this is different to the config validator functions I added a few weeks ago? These were meant to check if the config included untracked keys.

Maybe the main difference here is just expanding it to the grid script?

timlichtenberg:
All three taken, thanks.

`find_orphan_keys` and `_collect_orphan_keys` were indeed dead once both faults were collected in one pass, so they are gone and the walk is just `find_key_problems` now.

I did keep `cls` and `path` on it, with defaults, so a caller passes only the raw dict. They are not unused: they carry the recursion down into the nested schema classes, and hardcoding to `Config` would stop the walk at the top level and miss every nested typo. If you would rather have the recursion hidden behind a private helper again, say so, but that is the shape I just removed and two defaulted arguments seemed the lesser evil.

The `strict` parameter is gone. `read_config_object` always refuses now, and the structuring step it used to skip is `structure_config`, which the runner composes itself. The runner is the only caller that needs a configuration built from a file it is about to reject, because the output directory it records the refusal in is named inside that file. It also means the file gets read once instead of twice.

While in there I covered the union branch of the shape check, which is the path optional sections like `star.mors` take.

timlichtenberg:
> Could you clarify how this is different to the config validator functions I added a few weeks ago? These were meant to check if the config included untracked keys.
>
> Maybe the main difference here is just expanding it to the grid script?

Good point, and it is yours: the walk from #705 is still what does the work, I only built on it.

The difference is where it runs. Yours was called from `Proteus.__init__` only, so `proteus grid`, the download commands and install/update all bypassed it. It now sits in `read_config_object`.

The grid is not just one more caller though. It writes each case config out from the parsed object, so an unknown key in the base file is gone before any case file exists, and the per-case check then sees clean files. Misspell `maximum` in `tests/grid/base.toml` and every case in the ensemble quietly runs on the 1e7 default instead of the 3e7 you asked for.

Two smaller ones: the key names never actually reached the user, since the `log.error` calls fire before a handler is attached, and `[[planet]]` instead of `[planet]` is spelled correctly but drops the whole section, which a name check cannot see.

I have retitled the PR and rewritten the description to make the #705 heritage explicit, since it read as if this were new tooling rather than a continuation of yours.

egpbos:
> Could you clarify how this is different to the config validator functions I added a few weeks ago?

I knew I had seen it before! 😄

> I did keep `cls` and `path` on it, with defaults, so a caller passes only the raw dict. They are not unused: they carry the recursion down into the nested schema classes, and hardcoding to `Config` would stop the walk at the top level and miss every nested typo.

Ah, of course, my bad!

egpbos:
Thanks for addressing the comments @timlichtenberg, I dismissed my blocking review.

One additional general comment that maybe I mentioned somewhere before as well: this whole configuration machinery might be a good candidate for making a separate generic package out of; could be useful beyond PROTEUS. Or perhaps just fwl-config to start with ;) Beyond the scope of this PR, of course.

src/proteus/config/orphans.py
```diff
@@ -89,48 +139,85 @@ def _collect_orphan_keys(data: dict, cls: type, path: str = '') -> list[str]:
...
+        subject = 'This key is' if single else 'These keys are'
+        setting = 'Setting it' if single else 'Setting them'
+        blocks.append(
+            f'{heading} in {path}:\n'
+            f'  {keys}\n'
+            f'  {subject} not part of the configuration schema. {setting} '
+            f'has no effect, so the file is refused rather than run on defaults '
+            f'that were not asked for. Check for a typo or an outdated option '
+            f'name.'
+        )
```
nichollsh: This phrasing was confusing. Suggested update...

```suggestion
            f'  {subject} unrecognised, so {setting} will have no effect. '
            f'Check for typos or outdated option names.'
        )
```

src/proteus/config/orphans.py
```diff
@@ -89,48 +139,85 @@ def _collect_orphan_keys(data: dict, cls: type, path: str = '') -> list[str]:
...
+            else 'Misdeclared configuration sections'
+        )
+        subject = 'This section is' if single else 'These sections are'
+        blocks.append(
+            f'{heading} in {path}:\n'
+            f'  {sections}\n'
+            f'  {subject} declared as a table by the schema, but the file gives '
+            f'another kind of value. Writing a section as [[name]] rather than '
+            f'[name] is the usual cause. As written the whole section is '
+            f'discarded and every parameter inside it falls back to its default.'
```
nichollsh:
```suggestion
            f'  {sections}\n'
            f'  {subject} declared as a table by the schema, but the file gives '
            f'another kind of value. Check for typos or double-brackets.'
```

src/proteus/config/orphans.py
```diff
@@ -89,48 +139,85 @@ def _collect_orphan_keys(data: dict, cls: type, path: str = '') -> list[str]:
...
+        )
+
+    blocks.append('See input/all_options.toml for the full parameter reference.')
```
nichollsh:
```suggestion
    blocks.append('See input/all_options.toml for reference, or read the docs: ')
    blocks.append('https://proteus-framework.org/PROTEUS/Reference/config/params.html')
```

nichollsh (changes requested):
Great! Thanks for expanding the orphan-keys functionality in this way. It's easier to maintain and covers the other CLI commands.

I've tested this on my laptop by adding/removing/modifying various keys in the `all_options.toml` config. This prints the appropriate message and exits. I've also tested it through the grid utility (one of the main aims of this PR) and this works as described. The status file is updated appropriately.

Some phrasing/expositional suggestions below.

timlichtenberg:
> this whole configuration machinery might be a good candidate for making a separate generic package out of; could be useful beyond PROTEUS. Or perhaps just fwl-config to start with ;)

Good idea in the abstract, so I went and looked for who would consume it.

MORS, CALLIOPE and ZEPHYRUS read no configuration file at all. JANUS and Zalmoxis both `toml.load` into a plain dict with no schema behind it, so there is nothing for the walk to compare against; they would each need to define a schema first, next to which adopting a library is the small part of the job.

That leaves aragog, which does have attrs configuration classes. Even there it is half the problem: `from_dict` builds each section as `_SolverParameters(**data['solver'])`, so a bad key inside a known section already raises. What it misses is a misspelled section name, where `initial_condition` in particular falls back to every default via `.get`. That is a twenty-line check in aragog, not a dependency.

So I would keep it here for now. The shared surface is the 220-line walk and nothing else, and configuration loading is the one path where a version-pin mismatch stops everything from starting. fwl-io earns its keep because every module downloads reference data; this does not have the same shape yet.

timlichtenberg:
All three taken, thanks, and glad the grid path checked out on your side.

One adjustment inside the first suggestion. `setting` is built as `'Setting it'` / `'Setting them'` because it began a sentence in the old wording, so applied as written it came out as "so Setting it will have no effect". I lowercased both variants.

Two tests pinned the old phrasing, one on the singular and plural agreeing with the key count and one on the bracket advice appearing only when a section is actually misdeclared. Both now key on the new wording rather than the old. The reference link resolves.

nichollsh (approved):
Thanks! Good with me to merge.

PR #800, opened 2026-07-31
Write the JANUS snapshot from the column it solved

timlichtenberg:
## Description

Three integration-tier tests are failing on `main` right now. The last nightly ran on `eadad47e`, before both #775 and #634 merged, so nothing has reported them yet; tonight's run would.

The JANUS one is a real bug, not a stale test. `write_atmosphere_snapshot` wrote the atmosphere object the caller holds between iterations, but JANUS copies that object before it integrates and resamples the copy onto the radiative grid. The object the caller keeps therefore has no profile of its own: its arrays stay at the integration length with only the surface cell written, while the fluxes written back onto it are on the radiative grid. `write_ncdf` takes its level count from the profile and then assigns a flux array of the other length into it, so the write raised `shape mismatch ... (10001,) and (50,)` and no JANUS run could produce a final snapshot at all.

The solve already builds the column whose profiles and fluxes share a grid, and already writes its own per-iteration file from it, but never returned it. It now comes back alongside the outputs and the snapshot is written from it. The seed is left alone deliberately: the resample does not resize the integration index arrays, so reusing the solved column as the next seed would write past the end of them.

The other two are the whole-planet element checks, which compared the element list against a literal of the nine elements it used to hold. #634 added Al, Ti, Ca and K, so both failed although nothing they protect had changed. They now require the elements the mass balance depends on to be present, which is the property at stake, and separately require the list to reach every element its sources declare and hold each once. Growing the list is no longer a failure; dropping an element from it still is.

## Validation of changes

macOS 15 (Darwin 25.3.0), Python 3.12, conda env with the current module pins.

All three previously-failing tests pass, and the full local tiers are green on the branch tip: unit 2823 passed / 11 skipped, integration 197 passed / 11 skipped, smoke 24 passed / 1 skipped. `ruff check` and `ruff format --check` clean, `tools/validate_test_structure.sh` complete, and the test-quality baseline moves from 11 to 10.

Each fix was checked against the bug it is supposed to catch rather than only against the current tree. Reverting `RunJANUS` to hand back the seed fails the JANUS unit test; pointing the snapshot dispatch back at the old attribute fails the wrapper test; removing `Fe` from `vap_element_list` fails both element checks, which the previous wording of those checks did not.

The written snapshot was also inspected directly rather than only checked for the absence of a crash: a JANUS run through `Proteus.start()` produces `2_atm.nc` with `nlev_c=50`, `nlev_l=51`, `ngases=11`, no NaN or Inf, and `net_flux[0]` matching the coupled loop's `F_atm` to float32 precision, so the file holds the converged state the interior was given.

One thing this does not fix, worth a separate issue: the `gravity` array in the JANUS snapshot is not consistent with the `z` array beside it. `write_ncdf` reintegrates the heights just before writing but leaves `grav_z` at whatever the adiabat solve last set, so the two disagree by tens of percent at altitude. The per-iteration snapshots already carried this, so it is not introduced here, but `observe/petitRADTRANS.py` reads that column and it should be looked at on the JANUS side.

## Checklist

### DIFF
--- a/src/proteus/atmos_clim/common.py
+++ b/src/proteus/atmos_clim/common.py
@@ -22,6 +22,16 @@ class Atmos_t:
         # Atmosphere object internal to JANUS or AGNI
         self._atm = None
 
+        # The most recently solved atmosphere, for the modules whose solver
+        # returns a column that is not the one it was given. JANUS copies
+        # `_atm` before it integrates and resamples the copy onto the radiative
+        # grid, so only the copy carries a profile at all: `_atm` keeps the
+        # surface boundary condition between iterations and nothing else, its
+        # profile arrays sitting at their allocated length with only the
+        # boundary cell written. Read this one, never `_atm`, for anything that
+        # wants the atmosphere itself. AGNI solves in place and leaves it None.
+        self._atm_solved = None
+
         # Whether the most recent atmosphere call converged. For AGNI this
         # is True iff the Newton/LM solver converged on at least one attempt;
         # JANUS, dummy, and transparent solvers always set it True. The main
--- a/src/proteus/atmos_clim/common.py
+++ b/src/proteus/atmos_clim/common.py
@@ -22,15 +22,15 @@ class Atmos_t:
         # Atmosphere object internal to JANUS or AGNI
         self._atm = None
 
-        # The most recently solved atmosphere, for the modules whose solver
-        # returns a column that is not the one it was given. JANUS copies
-        # `_atm` before it integrates and resamples the copy onto the radiative
-        # grid, so only the copy carries a profile at all: `_atm` keeps the
-        # surface boundary condition between iterations and nothing else, its
-        # profile arrays sitting at their allocated length with only the
-        # boundary cell written. Read this one, never `_atm`, for anything that
-        # wants the atmosphere itself. AGNI solves in place and leaves it None.
-        self._atm_solved = None
+        # The column JANUS last solved, and JANUS only: None under every other
+        # atmosphere module. JANUS copies `_atm` before it integrates and
+        # resamples the copy onto the radiative grid, so only the copy carries
+        # a profile at all. `_atm` keeps the surface boundary condition between
+        # iterations and nothing else, its profile arrays sitting at their
+        # allocated length with only the boundary cell written, so anything
+        # that wants the atmosphere itself has to read this instead. AGNI
+        # solves its column in place and needs no second reference.
+        self._atm_janus_last = None
 
         # Whether the most recent atmosphere call converged. For AGNI this
         # is True iff the Newton/LM solver converged on at least one attempt;
--- a/src/proteus/atmos_clim/wrapper.py
+++ b/src/proteus/atmos_clim/wrapper.py
@@ -130,7 +130,7 @@ def run_atmosphere(
             InitStellarSpectrum(dirs, wl_un, fl_un, spectral_file_nostar)
             atmos_o._atm = InitAtm(dirs, config)
 
-        atmos_o._atm_solved, atm_output = RunJANUS(
+        atmos_o._atm_janus_last, atm_output = RunJANUS(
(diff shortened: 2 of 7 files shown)


src/proteus/atmos_clim/common.py
```diff
@@ -22,6 +22,16 @@ def __init__(self):
...
 
+        # The most recently solved atmosphere, for the modules whose solver
+        # returns a column that is not the one it was given. JANUS copies
+        # `_atm` before it integrates and resamples the copy onto the radiative
+        # grid, so only the copy carries a profile at all: `_atm` keeps the
+        # surface boundary condition between iterations and nothing else, its
+        # profile arrays sitting at their allocated length with only the
+        # boundary cell written. Read this one, never `_atm`, for anything that
+        # wants the atmosphere itself. AGNI solves in place and leaves it None.
+        self._atm_solved = None
```
nichollsh: Is it not possible to avoid adding *another* atmosphere variable into the atmosphere object? (See question below)

If not, perhaps consider giving this a clearer name, given that it's not used by AGNI. Maybe something JANUS-specific like `self._atm_janus_last`.

src/proteus/atmos_clim/wrapper.py
```diff
@@ -246,11 +248,17 @@ def write_atmosphere_snapshot(atmos_o: Atmos_t, config: Config, dirs: dict, hf_r
     elif config.atmos_clim.module == 'janus':
         from proteus.atmos_clim.janus import write_atmos_ncdf
 
-        if atmos_o._atm is None:
-            log.warning('Cannot write atmosphere; JANUS object unallocated')
+        # The solved column, not `_atm`: JANUS copies `_atm` before it
+        # integrates and resamples the copy, so `_atm` carries no profile of
+        # its own, only the surface boundary condition and the fluxes written
```
nichollsh, line 253: Fine with this approach, but I wonder if a better solution might be to change how JANUS handles this internally. Your call.

nichollsh (changes requested):
Thanks, @timlichtenberg, for catching this failure on the nightly integration. I believe that this is my fault for not testing the coupling with JANUS in my recent PR with @leojola.

Overall I think this is a good solution since JANUS is now more of a legacy module. However, I have one optional suggestion and one non-optional request. See below.

src/proteus/atmos_clim/common.py
timlichtenberg: I did try to avoid it. The only way to keep a single reference is to rebind `_atm` to the column JANUS returns, and that breaks the next iteration: `interpolate_atm` resamples the level arrays onto `nlev_save` but leaves `nlev` and the integration index arrays alone, so the following adiabat integration writes past the end of them. The two objects genuinely have different lifetimes, so they need different names.

Renamed to `_atm_janus_last` as you suggest, and the comment now leads with the fact that it is JANUS-only and None everywhere else. I also added a unit test on the dispatch that stores it, since that assignment is the one line where this could silently go back to writing the seed.

src/proteus/atmos_clim/wrapper.py
timlichtenberg: Agreed, and I think you are right that the cleaner home for it is inside JANUS, since which object it hands back and which it keeps are its own concepts rather than something the coupler should be reasoning about. I kept it on this side because it fixes the nightly now without needing a coordinated JANUS release, but I would not object to it moving later.

One thing for whenever that rework happens. While checking that the snapshot now holds the right state I noticed the `gravity` array in the file disagrees with the `z` array beside it: `write_ncdf` reintegrates the heights just before writing but leaves `grav_z` at whatever the adiabat solve last set, so the two differ by tens of percent at altitude. `observe/petitRADTRANS.py` reads that column for the reference gravity. The per-iteration snapshots already carried this so it is not new here, but it would fold naturally into the same internal change.

nichollsh (approved):
Understood. We can re-visit this in JANUS, then. Maybe worth making an issue in that repo to track it?

Happy for this to be merged.

Issue #710, opened 2026-06-16
Revisit timestepping for robustness and callbacks


nichollsh:
## Feature description

The PROTEUS modules are loosely coupled, which means that robust and physically-informed timestepping becomes important for handing off between modules. Better time-stepping can also provide better performance from the model.

Particularly important when handling impact of miscibility on planet structure.

## Preferred solution

Discuss how to do time-stepping better. Perhaps comparison against physics timescales would help to determine when to hand-off between modules? Perform callbacks/events during integration?

## Relevant people
@egpbos @timlichtenberg

nichollsh:
One additional problem relating to this issue: the `_estimate_*()` functions in `timestep.py` are not accessed when using time-stepping schemes other than `'adaptive'`. This means that setting `method='maximum'` will cause the model to overshoot termination events.

timlichtenberg:
A concrete case for this issue, measured on current main after #809, where the timestep and the escape module disagree about how much mass exists.

`calc_new_elements` in `src/proteus/escape/wrapper.py` computes `esc_mass = esc_rate_total * dt`, subtracts it from each reservoir whole, and clamps the result with `max(0.0, new_total)`. Nothing compares the requested loss against what the reservoir holds, so a step that asks for more than the whole atmosphere is absorbed silently instead of being rejected or shortened.

Two runs from a 30-case reducing-composition batch (fO2 = -5), read off their helpfiles. The first holds `M_atm` = 2.532e21 kg at an escape rate of 3.018e10 kg/s, then takes a step of dt = 3.156e11 s, so the step removes 9.53e21 kg, 3.8 times the entire atmosphere, and the run ends "volatiles escaped". The second does it over three steps, predicted and observed drops agreeing to about 7 % (2.418e22 to 1.526e22 to 1.57e21), the last step asking for 9.76e21 kg out of 1.57e21, a 6.2x overshoot, after which the inventory floors at 3.3e16 kg.

The rate itself is a separate problem, and the Hill clip in #809 bounds the radius it is computed from. What this issue can cover is the step. 26 of those 30 cases end at the same rate to four significant figures across an eleven-fold range in initial atmospheric mass, and in the two I dumped row by row the requested loss exceeds the reservoir, which is a signal the integrator could act on. Limiting the loss to the reservoir would hide it; shortening the step when the requested loss exceeds some fraction of what is available would not, and it fits the physics-timescale comparison already proposed here.

[...]

PR #802, opened 2026-07-31
Option for time-dependent scaling of stellar fluxes

### LINKED ISSUE #741: Parametrisation of temporary stellar dimming
## Feature description

For the purposes of a project with Tuhin Gosh and @shorttle (and myself), PROTEUS needs a simple user-configured parametrisation of temporary stellar dimming. This would be modulate the incoming stellar flux according to some user-defined configuration. 

## Preferred solution

Modulate `F_bol` using an inverted top-hat function, potentially with some amount of smoothing. It would take user-defined parameters which set its depth (0 to 1), width (years), and location (years).

## Additional information
Should happen after #726.

## Relevant people
@shorttle, but also tagging @timlichtenberg and @maraattia

nichollsh:
## Description

Support for time-dependent bolometric stellar luminosity scaling, allowing users to specify when and for how long a scaling factor is applied. This is part of a project with Tuhin Gosh and @shorttle. Closes #741 

Items below describe the changes that were necessary to implement this.

* Added new configuration options `bol_scale_start` and `bol_scale_duration` to control the timing and duration of bolometric luminosity scaling. Added default values into `all_options.toml`.
* Updated the stellar flux calculation in `update_instellation` to apply `bol_scale` only within the specified age range, and to record the active scaling factor in the output.
* Updated helpfile key to include the current `bol_scale` value.
* Implemented `_estimate_bolscale` to prevent the timestep from skipping over the start or end of the scaling period
* Avoid divide-by-zero in the calculation of Bond albedo and eclipse depths

## Validation of changes

* Ran `all_options.toml` with new keys at their default values. Works as before.
* Tested with `dummy.toml` with various start/duration times for dimming. See screenshots below.
* Added new tests for this new behaviour.
* Tests pass on my laptop (Fedora 44, Python 3.12) and on the GitHub CI.
* Performed self-review with Claude

**Demo with modified dummy config**

[example.zip](https://github.com/user-attachments/files/30585979/example.zip)

Using 
```
    bol_scale                 = 0.0          # bolometric luminosity scaling factor
    bol_scale_start           = 0.2          # stellar age from which bol_scale is applied [Gyr]
    bol_scale_duration        = 0.05          # duration for which bol_scale is applied [Gyr]
```

Note that the absorbed stellar flux (ASF) drops to zero and then recovers after the dimming period. This corresponds to a decrease in the surface temperature, which then also recovers to its previous value. The helpfile CSV contains rows in which `age_star` exactly equals 2e8 and 2.5e8, which confirms that the event trigger in `timestep.py` is operational.

image

## Checklist

### DIFF
--- a/docs/Reference/config/star_orbit.md
+++ b/docs/Reference/config/star_orbit.md
@@ -17,6 +17,10 @@ See also [Model description](../../Explanations/model.md#stellar-evolution-mors)
 | `mass` | float | `1.0` | Stellar mass \[M$_\odot$] |
 | `age_ini` | float | `0.1` | Model start age \[Gyr] |
 | `bol_scale` | float | `1.0` | Bolometric luminosity scaling factor |
+| `bol_scale_start` | float or none | `none` | Stellar age from which `bol_scale` is applied \[Gyr]; `none` disables scaling entirely |
+| `bol_scale_duration` | float | `0.0` | Duration for which `bol_scale` is applied \[Gyr], starting at `bol_scale_start` |
+
+`bol_scale` only takes effect while the model's stellar age lies within the start-plus-duration interval. So, all three options should be set together. The scaling factor is applied uniformly to all stellar flux terms.
 
 ### MORS stellar tracks `[star.mors]`
 
--- a/docs/Reference/config/star_orbit.md
+++ b/docs/Reference/config/star_orbit.md
@@ -17,7 +17,7 @@ See also [Model description](../../Explanations/model.md#stellar-evolution-mors)
 | `mass` | float | `1.0` | Stellar mass \[M$_\odot$] |
 | `age_ini` | float | `0.1` | Model start age \[Gyr] |
 | `bol_scale` | float | `1.0` | Bolometric luminosity scaling factor |
-| `bol_scale_start` | float or none | `none` | Stellar age from which `bol_scale` is applied \[Gyr]; `none` disables scaling entirely |
+| `bol_scale_start` | float or none | `none` | Stellar age from which `bol_scale` is applied \[Gyr]; `none` to disable, in which case `bol_scale` should be set to unity |
 | `bol_scale_duration` | float | `0.0` | Duration for which `bol_scale` is applied \[Gyr], starting at `bol_scale_start` |
 
 `bol_scale` only takes effect while the model's stellar age lies within the start-plus-duration interval. So, all three options should be set together. The scaling factor is applied uniformly to all stellar flux terms.
--- a/input/all_options.toml
+++ b/input/all_options.toml
@@ -173,6 +173,8 @@ config_version = "3.0"
     age_ini                   = 0.100        # model start age [Gyr]
     module                    = "mors"       # mors | dummy
     bol_scale                 = 1.0          # bolometric luminosity scaling factor
+    bol_scale_start           = "none"       # stellar age from which bol_scale is applied [Gyr]; None to disable
+    bol_scale_duration        = 0.0          # duration for which bol_scale is applied [Gyr]
 
     [star.mors]
         # Rotation and evolution
--- a/input/dummy.toml
+++ b/input/dummy.toml
@@ -10,13 +10,15 @@ config_version = "3.0"
 
     [params.dt]
         initial  = 1e2           # initial step [yr]
-        minimum  = 1e2           # min step [yr]
-        maximum  = 3e7           # max step [yr]
+        minimum  = 1e3           # min step [yr]
+        maximum  = 2e8           # max step [yr]
+        method = 'maximum'
 
         [params.stop.time]
+            enabled = true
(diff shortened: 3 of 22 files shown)


nichollsh:
This also motivates looking into #710 sometime soon.

nichollsh:
@timlichtenberg apologies for the last-minute changes after requesting review, but I noticed some test failures on the CI that didn't get picked up locally at first. I have now corrected these, and this is ready. No more planned changes.

timlichtenberg (changes requested):
Nice feature, and the timestep event is the right mechanism for it. Two things want fixing before this goes in.

The window closes one timestep late. `update_instellation` uses a closed upper bound, so the row whose age equals `bol_scale_start + bol_scale_duration` still carries the scaled flux, and `_estimate_bolscale` returns infinity from that age onwards, leaving the following step unconstrained. The scaled state then drives the interior and atmosphere for one extra step of arbitrary length. Running the dummy config with a 100 yr window gave 704 yr of dimming; with the stock `dt.maximum` of 3e7 yr the overshoot can dwarf the window. **Please make the upper edge exclusive**, `bol_scale_ini <= age < bol_scale_end`. That also makes `bol_scale_duration = 0.0` a real no-op instead of one full step of dimming. `test_update_instellation_window_boundaries_are_inclusive` pins the current behaviour and would need to flip.

`bol_scale` on its own is now silently ignored. Any existing 3.0 config that sets it without knowing about the new keys picks up `bol_scale_start = None` and loses the scaling with no message at all. A dummy config at `bol_scale = 0.5` gives F_ins identical to the unscaled run. **Please warn loudly when `bol_scale != 1.0` and `bol_scale_start is None`, or reject the combination in the validator.** The migration override does not cover this, since those configs are already 3.0.

Smaller point: `bol_scale_start` lost its type annotation, presumably because cattrs cannot structure `"none"` into `float | None`. The idiom already used a few lines above works, `float | str | None` as on `rot_pcntle` and `rot_period`, and the config, tools and star suites all pass with it. Worth restoring so the field stays visible to `typing.get_type_hints`.

nichollsh:
Thanks, @timlichtenberg. I've addressed all three points in my latest commit. 

1) Made upper edge exclusive on age
2) Reject invalid config combinations (`bol_scale != 1` and `bol_scale_start=None`)
3) Restored type hint using the idiom you suggested. I wasn't sure how to resolve it before, but adding the `| str` part worked.

Ready for another review pass :)

timlichtenberg (changes requested):
Thanks, the half-open edge and the validator both do what I asked. One thing survives, and it is the same symptom as before.

Short windows are still stretched. `update_stellar_quantities` runs only when `abs(Time - sinst_prev) > params.dt.starinst`, which defaults to 100 yr, while `_estimate_bolscale` sets dt to exactly the remaining window. When the window is comparable to `starinst`, the step that lands on the edge is not long enough to trigger a stellar update, so `F_ins` and `bol_scale` keep their stale values until the next update happens to fire.

On the dummy config a 100 yr window still gives 704 yr of dimming; at 500 yr it comes out exact, so the interval convention is right now and the residue is purely the update cadence. **Please force a stellar update at the window edges**, either by resetting `sinst_prev` when `_estimate_bolscale` clamps the step, or by folding the window state into the cadence condition. Otherwise any episode shorter than `dt.starinst` is quietly wrong, which is the regime the timestep event exists to serve.

Second, `bol_scale_duration` still defaults to 0.0, and `test_update_instellation_zero_duration_is_no_op` now pins that as a no-op. So `bol_scale = 0.5` with `bol_scale_start = 0.2` and the duration left at the shipped default passes the new validator and scales nothing, anywhere. **Please extend `valid_bol_scale_start` to also require `bol_scale_duration > 0` when `bol_scale != 1.0`**, so the guard covers both ways of writing a window that never opens.

Minor: the reference table still describes `none` as disabling scaling entirely, but it is now a hard error unless `bol_scale` is 1.0.

nichollsh:
Ah, thank you! I didn't notice this previously because all my configs have used very small values for `params.dt.starinst`. But you're right, this can easily be skipped even if a reasonably large value is used. Added a boolean into `interior_o` which records scenarios that have clamped the time-step. This approach can also be extended in the future by #710.

I have also updated the config validator guard to handle both ways in which the window can be deactivated/inactive. Also clarified the phrasing in the docs table.

This addresses all three comments from this most review. Ready for another look @timlichtenberg, please!

nichollsh:
Bonus plot with application to a stellar dimming event to the case of young Venus, below. This is with AGNI+MORS+CALLIOPE and the dummy interior, so provides a more physical validation than the example above.

image

timlichtenberg (approved):
This is in good shape now: the window edges land where they should, and the config guard covers both ways of writing a window that never opens. Thanks for turning these around so quickly. Good to merge on your side.

PR #808, opened 2026-08-01
Fail the resume parity check when a column drops out of it

timlichtenberg:
## Description

The resume parity check walks the helpfile column by column and compares the uninterrupted run against the restarted one. It skips any column that is not numeric on both sides, or not finite on the reference side, and it skips them without saying so. The only thing standing behind those two screens was a count of how many columns did get compared, with a floor of 400.

That count cannot do the job, because most of what this configuration writes holds still for the whole run. Of the 761 columns compared, 675 are constant across every row and 633 are identically zero; only 86 actually vary. A constant column agrees between any two runs, so it lifts the count whatever happens to the physics. An entire block could stop writing finite values in both legs, the escape rates (24 columns), the energy-conservation columns (about 20) or the per-gas ocean columns (37), and the count would stay far above 400 while the test reported full parity.

The comparison now records what each screen dropped and fails if either dropped anything, naming the columns. Nothing in this configuration is non-numeric or non-finite, so the expected set is empty and any future column that starts being screened out has to be looked at rather than absorbed. The count floor stays as a backstop for the one case the screens cannot see: a helpfile that collapsed outright, where both sides lose the same columns and still agree over what is left.

Changes:

- `tests/integration/test_integration_resume.py`: the two screens collect what they reject and the test asserts the collection is empty, reporting the count in each category and naming the first columns. The old inline floor becomes a named `MIN_COLUMNS_COMPARED = 400` carrying what it is now for.
- Same file: the header quoted 39 s for the file and 14 s for its slowest test, and said the tier ceiling applies to the file rather than to each test. Both now match what the file does.

No issue is tracked for this; it came out of the work on #790 and is small enough to fix directly.

## Validation of changes

Measured rather than assumed:

- The dummy run writes 762 columns over 115 rows and the check compares 761, with `runtime` the only skip. Of those 761, 675 have zero variance across the run and 633 are identically zero, leaving 86 that carry any signal. That is the measurement the change rests on.
- Blanking five varying physics columns (`T_magma`, `F_atm`, `Phi_global`, `M_atm`, `P_surf`) on the reference side now fails with `0 non-numeric and 5 non-finite columns were dropped before comparison`, naming all five. The old floor of 400 passed that case, and so would any count floor: 756 columns were still compared.
- The whole file passes. Wall time varies with machine load across repeated runs, 76 s to 106 s for the file and 26 s to 36 s for its slowest test, which is where the header range comes from.

Note that the integration tier does not run on pull requests, so the checks below do not exercise this test. I dispatched the nightly workflow on this branch to run it in CI.

Test configuration: macOS 15 (arm64), Python 3.12, rebased onto `main` at e60f5799.

## Checklist

### DIFF
--- a/tests/integration/test_integration_resume.py
+++ b/tests/integration/test_integration_resume.py
@@ -63,8 +63,9 @@ from helpers import PROTEUS_ROOT
 from proteus import Proteus
 
 # Integration tier. Each test runs two or three real dummy-module
-# simulations, so the file costs about 39 s locally and the slowest single
-# test about 14 s. The 300 s tier ceiling leaves room for slower CI runners.
+# simulations, so the file costs roughly 80 to 105 s locally and its slowest
+# single test 26 to 36 s, depending on machine load. The 300 s tier ceiling is
+# per test and leaves room for slower CI runners.
 pytestmark = [pytest.mark.integration, pytest.mark.timeout(300)]
 
 CONFIG = PROTEUS_ROOT / 'input' / 'dummy.toml'
@@ -135,6 +136,12 @@ PARITY_RTOL = 1.0e-6
 # Columns excluded from that comparison because they are not part of the
 # simulated state.
 PARITY_SKIP = ('runtime',)
+
+# Backstop on the comparison's coverage, for a helpfile that collapsed
+# outright. The screens below are asserted to drop nothing, so this fires only
+# if the schema itself lost most of its columns, which both sides would lose
+# together while still agreeing over the few that remained.
+MIN_COLUMNS_COMPARED = 400
 
 
 # Scratch directories created by Proteus construction, cleared after each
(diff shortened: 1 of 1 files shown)


nichollsh:
I don't quite understand what this is tackling. Does the helpfile now contain non-numeric values?

timlichtenberg:
No, nothing in the helpfile became non-numeric. That was my old comment's fault: it named "the helpfile schema changed to non-numeric columns" as an example of what the guard defends against, and it reads as though that had happened. I cut the phrase in this PR.

What this tackles is that the guard had gone stale. The assertion was `compared > 400`, written when the dummy helpfile carried about 460 numeric columns, so it left a margin of roughly 13%. That configuration now writes 758 columns and the comparison covers 757 of them, `runtime` being the only skip, and the floor never moved with it. It had come to permit almost half the helpfile dropping out of the parity check without anything failing.

That matters because both sides of the comparison come from the same run. If the column filter ever starts rejecting most columns, whatever survives still agrees with itself and the test passes while checking nearly nothing. The count of columns compared is the only assertion that catches it. 650 puts the margin back roughly where it started, about 86% of what the configuration writes today, leaving 107 columns of headroom for normal growth.

Issue #727, opened 2026-07-08
Visualise and record failed iterations


nichollsh:
## Feature description
Sometimes, one of the modules within PROTEUS will 'fail' to obtain a solution during a PROTEUS iteration. When this happens, it's usually due to Atmodeller, Aragog, SPIDER, or AGNI. The information is recorded in the simulation's log file.

However, people don't reliably look at log files, especially when running PROTEUS simulations as part of a grid.

## Preferred solution

* Record failed iterations in the helpfile (perhaps as +1 for success and -1 for failure)
* Show these on the plots. Specifically, I am imagining that these could be indicated on `plot_global` with vertical red lines using `axvline`.

## Additional information
Maybe worth re-visiting the plots at some point anyway, as they now feel a bit scattered and it's unclear which shows what.

## Relevant people
@timlichtenberg @egpbos @EmmaPostolec @IKisvardai

PR #809, opened 2026-08-02
Stop escape using radii from an atmosphere solve that did not converge

### LINKED ISSUE #807: Escape uses radii from an AGNI solve that did not converge
## Problem description

Follow-up to https://github.com/nichollsh/AGNI/issues/215, which covers the AGNI side of this. In that thread @nichollsh traced the failure to the atmosphere being unstable to a hydrostatic solution (the `-Ub` solver flag): hot low-mmw columns inflate until each layer is thousands of km thick, which is outside what a plane-parallel scheme can represent. I also noted there that the high-H cases "run on non-converged states and then exit via the escape criterion at ~10 kyr, which is spurious". This issue is the PROTEUS half of that sentence, which is what PROTEUS does with the radii when AGNI reports that it did not converge.

When AGNI exhausts all of its solver attempts, `run_atmos_agni` still reads `R_obs` and `R_xuv` off the unconverged atmosphere struct and writes them into `output`, and escape then consumes them. The inflated structure has an XUV radius of 6 to 22 Hill radii. Escape is called as `EL_escape(..., scaling=3)`, so the energy-limited rate grows as the cube of that radius: it comes out at 4.5e13 to 9.5e13 kg/s against 1.26e10 kg/s when the same configuration converges. The envelope lifetime falls below the timestep floor, the whole atmosphere is removed inside one step, and the case exits as "volatiles escaped" at t = 1.006e4 yr.

Steps to reproduce: run GJ 9827 d with AGNI plus Aragog, `fO2_shift_IW = -3`, `H_budget = 5000` ppmw, `tidal = false`, several times over. Compare the count of `Maximum attempts when executing AGNI` in the first few iterations against `R_xuv` and the escape rate.

## Expected behavior

Escape should not consume a radius from a solve that the solver itself rejected. `run_atmos_agni` already returns `agni_converged`, but the radii are written to `output` either way and only the deadlock check ever looks at the flag. Holding the last converged `R_xuv` and `R_obs`, or skipping the escape update for that iteration, both seem preferable to feeding a rejected structure into a rate that goes as its cube. Whatever we choose should make the outcome visible, since the present behaviour turns a convergence failure into a plausible-looking science result.

## Evidence

Eight runs of one identical GJ 9827 d configuration, same code, same inputs, no deliberate difference. In four of them AGNI exhausted every attempt on all seven initialisation iterations. In the other four it converged on five of the seven. Which one happens varies from run to run, and it decides the outcome completely.

| AGNI over the first 7 iterations | runs | R_xuv / R_Hill at t = 62 yr | escape rate [kg/s] | outcome |
|---|---|---|---|---|
| converged 5 of 7 | 4 | 0.449 | 1.259e10 to 1.260e10 | run on to t = 2.0e6 to 1.1e7 yr |
| never converged | 4 | 6.31 to 9.62 | 4.51e13 to 9.54e13 | stop at t = 1.006e4 yr, envelope gone |

Fisher exact one-tailed p = 0.014.

In the wider grid this is not a rare corner. 58 of 465 GJ 9827 d cases exit this way, all of them with status 15, and 57 of the 58 had exceeded the Hill radius at some point during the run, against 1 of the 384 that never did. Both AGNI failure reasons lead here, 38 of the 58 reporting mostly `max_iterations` and 20 mostly `singular_jacobian`, so the operative condition is that every attempt was exhausted rather than which reason was printed.

As a check that the radius is the operative variable, bounding `R_xuv` and `R_obs` to the Hill radius before escape reads them lets all 57 of the 58 stalled cases that produce a trajectory pass the point where they previously stopped. 47 of those then lose the envelope over 4 to 48 resolved steps instead of one, and 10 keep it.

## My computer

AlmaLinux 8.10 on the Habrok cluster, Python 3.12.13. PROTEUS 099a773d, AGNI 1.11.4 with SOCRATES 2603.8 and Julia 1.12.5, Aragog 26.7.4, CALLIOPE 26.7.3, ZEPHYRUS 26.7.10, MORS 26.01.02.

## Additional notes

Three places already come close to catching this and none does.

`orbit/wrapper.py` logs `Atmosphere extends beyond the Hill radius` whenever `max(R_obs, R_xuv) > hill_radius`. `run_orbit` runs before `run_escape` in the same iteration, so the condition is detected, written to the log, and then ignored.

ZEPHYRUS `EL_escape` raises `ValueError` for `R_Hill/R_xuv <= 1`, documented as the point where the energy-limited approximation no longer applies. That check is only evaluated when `tidal_contribution` is true, so with `tidal = false` it is unreachable.

`_check_atmosphere_deadlock` does read the convergence flag, but it only counts an iteration in which the interior is frozen as well. Here the interior kept evolving, so the counter was reset on every one of the seven consecutive failures. This is why these cases exit quietly through the escape criterion instead of aborting: the guard meant for this regime, and whose error text names it, never fires.

This is worth doing regardless of what happens in AGNI#215. Even once the inflated-atmosphere regime is handled there, a solve can still fail, and PROTEUS should not derive an escape rate from a structure the solver rejected.

## Relevant people

@nichollsh
@maraattia since it will be relevant for the sub-Neptune escape cases

timlichtenberg:
## Description

Closes #807.

When AGNI rejects its solution it still returns an atmospheric structure, and PROTEUS read the photospheric and XUV levels off that structure. Escape reads `R_xuv` from the previous iteration and the energy-limited rate goes as its cube, so a convergence failure turned into a large mass-loss rate that looked like a physical result. In the runs where I found this, escape was computing a rate from a radius of six to twenty-two Hill radii, and the affected cases lost their envelopes in one or two steps.

Two mechanisms address this, one for rejected solves and one for the level itself.

Each converged solve records `R_obs`, `p_obs`, `T_obs`, `g_obs`, `R_xuv`, `p_xuv`, `T_xuv`, `g_xuv` and the volume mixing ratios at the XUV level, and a rejected solve has those replaced by the record. They are carried as a group, so a held radius is never combined with the pressure, temperature, gravity or composition of a rejected structure; BOREAS reads the composition alongside the radius, so carrying only the radii would have left it mixing two structures. The quantities derived from the radius, `rho_obs` and the transit depth, follow the carried value. Fluxes and surface state are never carried, since the coupling advances on them and the deadlock detector needs to see them stop moving. The record lives in memory only, so a resumed run whose first solve is rejected falls back on the last committed row, written before the run began; later solves do not reach back, since once a run has written rows of its own an empty record means those rows carry rejected levels too.

Independently of convergence, the XUV level is bounded: gas beyond the Hill radius is not bound to the planet, so with `escape.hill_clamp` (on by default) each atmosphere module limits `R_xuv` to `escape.hill_clamp_frac` of the Hill radius, floored at the solid-body radius, and re-reads `p_xuv`, `T_xuv`, `g_xuv` and the XUV mixing ratios at the clipped radius, keeping the level self-consistent. This is also the bound for a run that has never converged a solve, where the carry has nothing to hold. The v2-to-v3 config migration pins the clip off so migrated configs reproduce their previous escape rates. The observed level is not clipped; the orbit module's existing beyond-Hill warning still covers it.

`T_xuv` and `g_xuv` are new outputs in all three atmosphere modules. Two helpfile columns persist the solve outcome: `atm_converged` (+1 converged, -1 rejected, 0 before the first solve) and `atm_levels_stale` (consecutive iterations without a converged solve of the run), which is the recording half of #727. Every substitution is logged with that count and reported at error level past ten in a row, since the deadlock detector fires only when the interior has stopped moving as well. Modules without a nonlinear solve (JANUS, dummy, AGNI's transparent and prescribed-temperature branches) always report convergence.

Resuming a run whose helpfile predates a schema addition remains unchanged by this PR and is filed as #811.

## Validation of changes

macOS 15 (Darwin 25.3.0, arm64), Python 3.12, AGNI 1.12.0, SOCRATES 2603.8.

- `pytest -m "unit and not skip and not slow and not integration" --ignore=tests/examples`: 2932 passed, 11 skipped.
- `ruff check` and `ruff format --check` clean; `bash tools/validate_test_structure.sh` passes.
- 21 new tests across `tests/atmos_clim/` and `tests/config/`. I checked each behaviour by breaking its code path and confirming only the covering test fails: dropping the substitution, dropping the XUV composition from the carried set, reading the wrong committed row, counting only failed solves towards the first-solve rule, replacing the record instead of merging, removing the non-finite filter, hardcoding the logged provenance, removing the streak escalation, dropping the R_int floor from the clip, clipping the radius without re-reading the pressure, and ignoring the config switch.
- The wiring is covered as well as the helpers: mocked-AGNI tests drive `run_atmosphere` for a rejected solve mid-run and for the first solve after a resume, asserting on `hf_row`, which is what escape reads; module-level clip tests pin the re-read `p/T/g` at the clipped radius against profile values chosen so reading the wrong end of the column fails by a factor of 3.2.
- A full dummy-module run confirms the new helpfile columns populate with no NaN or unknown-key warnings.

## Checklist

### DIFF
--- a/docs/Explanations/coupling_loop.md
+++ b/docs/Explanations/coupling_loop.md
@@ -217,6 +217,21 @@ converged, -1 rejected, 0 before the first atmosphere call), and
 `atm_levels_stale` records the number of consecutive iterations without a
 converged solve of the run, zero on every converged row.
 
+## The XUV level is limited to the Hill radius
+
+Independently of convergence, the XUV level itself is bounded: gas beyond the
+Hill radius is not bound to the planet, so an XUV radius outside it would size
+the escape cross-section with material the planet does not hold, and the
+energy-limited rate grows as the cube of the excess. With `escape.hill_clamp`
+enabled (the default), each atmosphere module limits $R_\mathrm{xuv}$ to
+`escape.hill_clamp_frac` of the Hill radius, floored at $R_\mathrm{int}$ since
+the solid body is always bound. The level moves as a whole: $p_\mathrm{xuv}$,
+$T_\mathrm{xuv}$, $g_\mathrm{xuv}$ and the XUV-level mixing ratios are read at
+the clipped radius, so escape never sees a radius from one level combined with
+a composition from another. Each engagement is logged as a warning. The
+observed level is not clipped; a transit radius beyond the Hill radius is
+reported as computed and flagged by the orbit module's existing warning.
+
 ## Energy conservation diagnostics
 
 When using the Aragog interior module, PROTEUS tracks cumulative energy
--- a/docs/Explanations/coupling_loop.md
+++ b/docs/Explanations/coupling_loop.md
@@ -181,17 +181,18 @@ planet. Escape reads $R_\mathrm{xuv}$ from the previous iteration, and the
 energy-limited rate goes as $R_\mathrm{xuv}^3$, so an unusable structure would
 otherwise become a large mass-loss rate that looks like a physical result.
 
-When the atmosphere solve does not converge, PROTEUS therefore substitutes
-$R_\mathrm{obs}$, $p_\mathrm{obs}$, $T_\mathrm{obs}$, $g_\mathrm{obs}$,
-$R_\mathrm{xuv}$, $p_\mathrm{xuv}$ and the volume mixing ratios at the XUV level
-with the values from the most recent converged solve, and logs the substituted
-value of each of the six scalars as a warning. They are carried as a group, so a
-held radius is never
-combined with the pressure, temperature, gravity or composition of a rejected
-structure; the quantities derived from the radius, such as $\rho_\mathrm{obs}$
-and the transit depth, are computed from the carried value. Fluxes and surface
-state are never carried: the coupling advances on them, and the deadlock
-detector above needs to see them stop moving.
+When the atmosphere solve does not converge, PROTEUS therefore substitutes the
+radius, pressure, temperature and gravity of both levels ($R_\mathrm{obs}$,
+$p_\mathrm{obs}$, $T_\mathrm{obs}$, $g_\mathrm{obs}$, $R_\mathrm{xuv}$,
+$p_\mathrm{xuv}$, $T_\mathrm{xuv}$, $g_\mathrm{xuv}$) together with the volume
+mixing ratios at the XUV level, using the values from the most recent converged
+solve. A warning reports each such iteration, and the substituted values are
+listed at debug level. The levels are carried as a group, so a held radius is
+never combined with the pressure, temperature, gravity or composition of a
+rejected structure; the quantities derived from the radius, such as
+$\rho_\mathrm{obs}$ and the transit depth, are computed from the carried value.
+Fluxes and surface state are never carried: the coupling advances on them, and
+the deadlock detector above needs to see them stop moving.
 
 The record of converged levels is not written to the output files, so a resumed
 run starts without one. If the first solve of such a run is rejected, it falls
@@ -210,8 +211,11 @@ row the run reports it at error level. A streak that long means the interior is
 evolving while the levels stand still, which the deadlock detector above cannot
 catch: it fires only when the interior has stopped moving too.
 
-The substitution appears in the log, not in the helpfile: no output column marks
-a row whose levels were carried.
+Two helpfile columns persist the outcome, so a carried row is identifiable from
+the output alone: `atm_converged` records the solve outcome of each row (+1
+converged, -1 rejected, 0 before the first atmosphere call), and
+`atm_levels_stale` records the number of consecutive iterations without a
+converged solve of the run, zero on every converged row.
 
 ## Energy conservation diagnostics
 
--- a/docs/Explanations/coupling_loop.md
+++ b/docs/Explanations/coupling_loop.md
@@ -173,6 +173,46 @@ by tracking consecutive iterations where:
 After three consecutive deadlocked iterations, PROTEUS aborts with a
 diagnostic message identifying the stuck state.
 
+## Levels from a rejected atmosphere solve
+
+A solver that rejects its solution still returns an atmospheric structure, and
+the photospheric and XUV levels read off that structure can sit far outside the
+planet. Escape reads $R_\mathrm{xuv}$ from the previous iteration, and the
+energy-limited rate goes as $R_\mathrm{xuv}^3$, so an unusable structure would
+otherwise become a large mass-loss rate that looks like a physical result.
+
+When the atmosphere solve does not converge, PROTEUS therefore substitutes
+$R_\mathrm{obs}$, $p_\mathrm{obs}$, $T_\mathrm{obs}$, $g_\mathrm{obs}$,
+$R_\mathrm{xuv}$, $p_\mathrm{xuv}$ and the volume mixing ratios at the XUV level
+with the values from the most recent converged solve, and logs the substituted
+value of each of the six scalars as a warning. They are carried as a group, so a
+held radius is never
+combined with the pressure, temperature, gravity or composition of a rejected
+structure; the quantities derived from the radius, such as $\rho_\mathrm{obs}$
+and the transit depth, are computed from the carried value. Fluxes and surface
+state are never carried: the coupling advances on them, and the deadlock
+detector above needs to see them stop moving.
+
+The record of converged levels is not written to the output files, so a resumed
+run starts without one. If the first solve of such a run is rejected, it falls
+back on the last committed row instead, which was written before the run began
+and is the state escape would have used in any case; the warning names which of
+the two sources it used. Later solves never reach back to the committed rows,
+because once a run has begun writing rows, an empty record means those rows carry
+rejected levels themselves. A run with nothing to fall back on keeps the levels of
+the rejected structure and warns that it is doing so. Modules without a nonlinear
+solve (JANUS, the dummy module, and AGNI's transparent and prescribed-temperature
+branches) always report convergence and are unaffected.
+
+Each warning counts the consecutive iterations whose levels the run did not
+resolve itself, the ones with nothing to fall back on included, and past ten in a
+row the run reports it at error level. A streak that long means the interior is
+evolving while the levels stand still, which the deadlock detector above cannot
+catch: it fires only when the interior has stopped moving too.
+
+The substitution appears in the log, not in the helpfile: no output column marks
+a row whose levels were carried.
+
 ## Energy conservation diagnostics
 
 When using the Aragog interior module, PROTEUS tracks cumulative energy
--- a/src/proteus/atmos_clim/common.py
+++ b/src/proteus/atmos_clim/common.py
@@ -39,6 +39,25 @@ class Atmos_t:
         # failures with no interior state change). Transient, not persisted.
         self.converged: bool = True
 
+        # Photospheric and XUV level properties from the most recent solve
+        # that converged, keyed as in the helpfile row. Empty until this run
+        # converges a solve of its own; a run that has to substitute before
+        # then falls back on the last committed row. Transient, not persisted.
+        self.levels_converged: dict[str, float] = {}
+
+        # Where the levels above came from, and how many consecutive iterations
+        # have run on levels this run did not resolve itself. The source is
+        # tracked here rather than re-derived per call, since a run that has
+        # only ever fallen back on the committed row must not report those
+        # levels as converged. Transient, not persisted.
+        self.levels_source: str = 'last converged solve'
+        self.levels_carried: int = 0
+
+        # Atmosphere solves this run has made. Only the first one may fall back
+        # on the rows committed before the run started; after that, an empty
+        # record means this run's own rows carry rejected levels too.
+        self.solves_seen: int = 0
+
 
 def ncdf_flag_to_bool(var) -> bool:
     """Convert NetCDF flag (y/n) to Python bool (true/false)"""
--- a/src/proteus/atmos_clim/common.py
+++ b/src/proteus/atmos_clim/common.py
@@ -342,6 +342,52 @@ def get_spfile_path(fwl_dir: str, config: Config):
     return os.path.join(fwl_dir, 'spectral_files', group, bands, group) + '.sf'
 
 
+def clip_radius_to_hill(config: Config, hf_row: dict, radius: float) -> float:
+    """Limit a level radius to the Hill radius, never below the solid body.
+
+    Gas beyond the Hill radius is not bound to the planet, so an XUV radius
+    outside it sizes the escape cross-section with material the planet does
+    not hold, and the energy-limited rate grows as the cube of the excess.
+    The limit is ``escape.hill_clamp_frac`` of the Hill radius, floored at
+    ``R_int`` since the solid body is always bound.
+
+    Parameters
+    ----------
+        config : Config
+            Configuration options for PROTEUS.
+        hf_row : dict
+            Current helpfile row; provides ``hill_radius`` and ``R_int``.
+        radius : float
+            Level radius to limit [m].
+
+    Returns
+    ----------
+        float
+            The radius, limited when the clip is enabled and applicable.
+    """
+    if not getattr(config.escape, 'hill_clamp', False):
+        return radius
+
+    # Zero before the first orbit update; nothing to clip against yet.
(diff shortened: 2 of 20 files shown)


timlichtenberg:
One limitation worth recording before this is merged, measured on the batch of cases that motivated the issue.

The fallback only helps a run that converges and then stumbles. In a batch of 46 cases selected for the artefact in #807, 38 have never converged a single AGNI solve: the first atmosphere call of the run exhausts its attempts and so does every call after it. For those there is no converged structure to hold, the record stays empty, and escape still runs on the rejected radius exactly as before. What this fix adds there is the warning and the streak escalation, not a bounded radius.

In the wider grid the picture is the other way round. Sampling 32 live cases across four quarters, 22 never fail a solve at all, 8 fail intermittently, which is the case this fix covers, and 2 never converge.

So the open question is what a run with no converged solve should do. Bounding the radius, by the Hill radius or otherwise, is a different physical statement from holding a resolved one, so I have not put it in here. Happy to add it if you think the bound is the right floor case.

nichollsh:
> what a run with no converged solve should do.

Good question. I suppose there's no correct answer, since a non-converged run won't have a "correct" radius.

We could first calculate a fully-adiabatic temperature profile before the AGNI solver is called, and obtain a "fallback" set of R_obs and R_xuv from that. It's fairly reasonable to assume the atmospheres are initially convective, and adiabatic structures tend to be more safely within the hydrostatic regime.

Bounding by the hill radius seems like a reasonable limit to apply, as you suggest. This limit could be applied on every time-step.

We could certainly do both of these as part of this PR - I would support this.

src/proteus/atmos_clim/common.py
```diff
@@ -39,6 +39,25 @@ def __init__(self):
...
+        # converges a solve of its own; a run that has to substitute before
+        # then falls back on the last committed row. Transient, not persisted.
+        self.levels_converged: dict[str, float] = {}
+
+        # Where the levels above came from, and how many consecutive iterations
+        # have run on levels this run did not resolve itself. The source is
+        # tracked here rather than re-derived per call, since a run that has
+        # only ever fallen back on the committed row must not report those
+        # levels as converged. Transient, not persisted.
+        self.levels_source: str = 'last converged solve'
```
nichollsh: Rather than a string, it might be more appropriate to use a simple integer flag here, or even an [enum](https://docs.python.org/3/library/enum.html).

src/proteus/atmos_clim/common.py
```diff
@@ -39,6 +39,25 @@ def __init__(self):
...
+        self.levels_carried: int = 0
```
nichollsh: This variable name is extremely ambiguous. Please clarify and add a comment describing it. It could even go in the helpfile, to be honest.

src/proteus/atmos_clim/common.py
```diff
@@ -39,6 +39,25 @@ def __init__(self):
...
+
+        # Atmosphere solves this run has made. Only the first one may fall back
+        # on the rows committed before the run started; after that, an empty
+        # record means this run's own rows carry rejected levels too.
+        self.solves_seen: int = 0
```
nichollsh, line 78: This could also go in the helpfile.

src/proteus/atmos_clim/wrapper.py
```diff
@@ -13,11 +13,154 @@
...
 from proteus.utils.helper import UpdateStatusfile, safe_rm
 
 log = logging.getLogger('fwl.' + __name__)
 
+# Scalar properties of the photospheric and XUV levels, all read off the
+# atmospheric structure the solver returns. They are carried together so that
+# each level keeps a radius, pressure, temperature and gravity that describe the
+# same structure. Fluxes and surface state are deliberately absent: the coupling
+# and the deadlock detector both need to see those move when a solve fails.
+LEVEL_KEYS = ('R_obs', 'p_obs', 'T_obs', 'g_obs', 'R_xuv', 'p_xuv')
```
nichollsh: Should also include `g_xuv` and `T_xuv` for completeness.

src/proteus/atmos_clim/wrapper.py
```diff
@@ -13,11 +13,154 @@
...
+# Composition at the XUV level, read off the same structure and used by BOREAS
+# to set the mean molecular weight of the outflow. Carried with the levels, so
+# a held radius is never combined with the composition of a rejected structure.
+XUV_VMR_KEYS = tuple(f'{gas}_vmr_xuv' for gas in gas_list)
+
+# Number of consecutive iterations on carried levels after which the run says so
+# at error level. Long enough that a solver which stumbles for an iteration or
+# two passes unremarked, short enough to appear while a run is still worth
+# steering.
+CARRIED_LEVELS_ALERT = 10
```
nichollsh, line 48: I think the solve should print an error or warning for every iteration that uses a carried-over set of values for `*xuv` and `*obs`

src/proteus/atmos_clim/wrapper.py
```diff
@@ -13,11 +13,154 @@
...
+    to the output files. If its very first solve is rejected it falls back on the
+    last committed row, which was written before this run began and is the state
+    escape would have used in any case. That row is the last one written, not
+    necessarily one that converged, so the substitution says which of the two
+    sources it used.
+
+    Later solves never reach back to the committed rows. Once this run has begun
+    writing rows, an empty record means every solve so far was rejected, so those
+    rows carry rejected levels themselves and are worth no more than the levels
+    in hand.
```
nichollsh: Please re-phrase this. It's hard to parse and understand.

src/proteus/atmos_clim/wrapper.py
```diff
@@ -13,11 +13,154 @@
...
+        atmos_o.levels_carried,
+    )
+    _alert_on_long_streak(atmos_o)
+
+    for key in LEVEL_KEYS:
+        if key in atmos_o.levels_converged and key in hf_row:
+            log.warning(
+                '    %-5s %.5e  ->  %.5e'
+                % (key, float(hf_row[key]), atmos_o.levels_converged[key])
+            )
```
nichollsh: This creates a lot of log-spam. Maybe leave it as `log.debug`? The user is already warned every iteration if you set `CARRIED_LEVELS_ALERT=1`.

nichollsh (changes requested):
Thanks for tackling this, @timlichtenberg. It's important we cover this behaviour since it can arise naturally from the physics in a wide range of phase-space.

In addition to my comment/suggestion above, I have some more requests below.

src/proteus/atmos_clim/common.py
timlichtenberg: It's an enum now (LevelsSource, two members), and the log line maps it to a phrase.

src/proteus/atmos_clim/common.py
timlichtenberg: Renamed to levels_stale_iters with a comment saying exactly what it counts, and it now goes to the helpfile as atm_levels_stale on every row.

src/proteus/atmos_clim/common.py
timlichtenberg: Persisting the raw call counter felt less useful than persisting the outcome, so the helpfile now gets atm_converged (+1 converged, -1 rejected, 0 before the first solve) per row instead, which is the recording half of #727. Loading a helpfile also reconciles its columns against the current schema now, so runs written before these columns existed still resume.

src/proteus/atmos_clim/wrapper.py
timlichtenberg: Added both. AGNI reads them off its cell-centre profiles at p_xuv (surface values in the transparent branch, where the profiles keep the pre-solve guess); JANUS takes T from its profile and g by the inverse-square law like its g_obs, with a surface fallback when the hydrostatic integration failed; dummy's XUV level coincides with the observed one.

src/proteus/atmos_clim/wrapper.py
timlichtenberg: It does warn on every affected iteration; the message with the running count a few lines down is per iteration, and the CARRIED_LEVELS_ALERT error at ten in a row is on top of that, not instead of it. The outcome is also in the helpfile now (atm_converged, atm_levels_stale), so it's visible without reading the log at all.

src/proteus/atmos_clim/wrapper.py
timlichtenberg: Rewritten in plainer terms.

src/proteus/atmos_clim/wrapper.py
timlichtenberg: Agreed, the per-value lines are debug now; the one-line warning per iteration stays.

src/proteus/atmos_clim/common.py
nichollsh: Thanks for adding this to the helpfile.

docs/Explanations/coupling_loop.md
```diff
@@ -210,8 +211,11 @@ row the run reports it at error level. A streak that long means the interior is
 evolving while the levels stand still, which the deadlock detector above cannot
 catch: it fires only when the interior has stopped moving too.
 
-The substitution appears in the log, not in the helpfile: no output column marks
-a row whose levels were carried.
+Two helpfile columns persist the outcome, so a carried row is identifiable from
+the output alone: `atm_converged` records the solve outcome of each row (+1
+converged, -1 rejected, 0 before the first atmosphere call), and
+`atm_levels_stale` records the number of consecutive iterations without a
+converged solve of the run, zero on every converged row.
```
nichollsh, line 218: I am happy with this now. Thanks for recording these in the HF and adding this explanation.

src/proteus/utils/coupler.py
```diff
@@ -1283,12 +1287,26 @@ def WriteHelpfileToCSV(output_dir: str, current_hf: pd.DataFrame):
...
-    return pd.read_csv(fpath, sep=r'\s+')
+    df = pd.read_csv(fpath, sep=r'\s+')
+    missing = [key for key in GetHelpfileKeys() if key not in df.columns]
+    if missing:
+        log.warning(
+            'Helpfile lacks %d column(s) added since it was written; filling with zero: %s',
+            len(missing),
+            missing,
+        )
+    return df.reindex(columns=GetHelpfileKeys(), fill_value=0.0)
```
nichollsh: This is an important change in behaviour and not something to be buried here so lightly. It means resuming from old simulations comes with substantial physical implications. I think that taking this blanket approach of zeroing-out columns that were previously missing is presumptive. We might instead wish to have the model hard-exit in such a case, or maybe add a flag like `--force` that complements `--resume`.

Please do not include this in the current PR. Instead, make an issue so we can consider this further.

nichollsh (changes requested):
Thanks for addressing most of my comments. However, I have two more issues with this PR (see also my inline comment below).

1. The new functionality to resume simulations forcefully by zeroing-out HF columns seems like an overreach. We should leave this for a separate PR.

2. You didn't address my endorsement of the Hill Radius clipping, which you originally suggested. I encourage including that in this PR since it's easy to physically justify and would improve the robustness of your new functionality substantially. Why not include it here?

src/proteus/utils/coupler.py
timlichtenberg: Agreed, it's out: the loader is back to the plain read and the design question is #811, with hard-exit and a --force-style flag as the candidate behaviours to weigh there.

timlichtenberg:
Both points are in.

The Hill clip is included and on by default. Each atmosphere module limits R_xuv to escape.hill_clamp_frac of the Hill radius (floored at R_int) and re-reads p_xuv, T_xuv, g_xuv and the XUV mixing ratios at the clipped radius, so the level stays self-consistent rather than mixing a bounded radius with an unbounded structure. BOREAS is bounded where its rate is computed: the solved radius is clipped on storage and the bulk rate scales with the area ratio, so rate and recorded radius describe the same cross-section. Carried radii from older helpfile rows pass through the same bound. This also gives a run that has never converged a solve, where the carry has nothing to hold, a physical ceiling. The v2-to-v3 migration pins the clip off so migrated configs keep their previous escape rates.

The resume reconciliation is reverted to the plain read and filed as #811 with the hard-exit and --force designs you suggested as the starting options.

nichollsh (approved):
Fantastic - thank you! 

I think this PR is now in good shape. Happy with the documentation updates and new functionalities. I think keeping the HF-restore stuff in a separate issue is the right move. Hill Radius clipping seems like a good scope for this PR too, so thanks for adding this in.

Tested on my laptop and works as described. Happy for this to be merged!

PR #812, opened 2026-08-03
Record the trajectory again for the four columns #809 added

timlichtenberg:
## Description

The nightly golden-run check went red on `main`. #809 added four columns to the helpfile, `T_xuv`, `g_xuv`, `atm_converged` and `atm_levels_stale`, without recording the trajectory again in the same commit, so the reference does not carry them and the run reports them as produced but absent. The guard is behaving correctly; the reference is stale.

I checked what the four hold before recording them rather than after, since re-recording is how a wrong value gets blessed permanently. On this configuration the dummy atmosphere puts the XUV level at the observed radius, so `T_xuv` is the surface temperature and `g_xuv` is the surface gravity scaled by the inverse square of `R_int` over `R_xuv`. Every solve converges, so `atm_converged` holds +1 and `atm_levels_stale` holds 0 on all 56 rows. The Hill limit never engages here, the observed radius staying at a hundredth of the Hill radius.

The recording adds four lines and the column count in the header, and leaves the other 757 columns and the configuration digest untouched, so #809 did not shift the trajectory on the dummy backends.

The second commit stops this recurring: the comparison runs in the nightly tier, so a helpfile key added without a recording is only reported once it has already merged. Reading the schema and the reference needs no run, so it is now compared on every pull request instead, and the same four columns reach the output reference page.

## Validation of changes

Reproduced the failure on `main` first, so the local run sees what the nightly sees, then verified each column against a closed form rather than against the run:

- `g_xuv` equals `gravity * (R_int / R_xuv)**2` with zero relative deviation on every row, and equals `g_obs`
- `T_xuv` equals `T_surf` and `T_magma` exactly, spanning 4000 K down to 1572 K
- `R_obs` equals `R_int + 3H` exactly, with `H` from the recorded `T_surf`, mean molecular weight and gravity
- `atm_converged` is +1 and `atm_levels_stale` is 0 on all 56 rows, which is what the dummy path can produce
- maximum `R_obs / R_hill` is 1.0e-2, so the Hill clip is inactive

The same contracts are pinned independently in the unit tier, in `tests/atmos_clim/test_atmos_clim.py` for the dummy levels and `tests/atmos_clim/test_wrapper.py` for the two flags, so these are the intended values rather than an artefact of one run. Perturbing each of the four new columns is reported on that column alone, and the all-zero column carries no tolerance, so any nonzero value in it fails.

The new guard was checked against the stale reference: it names the exact four missing columns in under a second.

Test configuration: macOS 15 (Apple silicon), Python 3.12, all modules on their dummy backends.

- Integration tier: 198 passed, 11 skipped
- Unit tier: 2943 passed, 7 skipped
- `ruff check src/ tests/` and `ruff format --check src/ tests/` clean
- `bash tools/validate_test_structure.sh` clean

## Checklist


tests/integration/test_integration_golden_run.py
nichollsh, line 1: Do we really need to write the numbers and fractions of tested columns here? They're just a code comment and it seems like a burden to maintain them without any added value.

nichollsh (approved):
Thanks for catching this on the nightly run. Happy with the changes although I have one optional comment.

Good idea to add the fast test to handle this before it propagates to the nightly, in the future.

Good to merge, unless you also want to adopt my suggestion.

tests/integration/test_integration_golden_run.py
timlichtenberg: Agreed, and the second commit here makes them redundant: the new check derives the covered set from GetHelpfileKeys rather than from anything written down, so the count in the comment is no longer the thing holding the floor honest. I am stripping the numbers in a follow-up rather than here, because a push dismisses your approval and this one is fixing the red nightly.

The floor itself stays at 650. What I will keep is why it exists, without the arithmetic that has to be hand-edited every time the helpfile grows.

PR #813, opened 2026-08-03
Stop an installed module changing how imports are sorted

timlichtenberg:
## Description

ruff decides whether an import is first party by searching for the name, and by default it searches the repository root as well as the source tree. The ecosystem modules are installed as clones at that root, so on a machine that has one, the name is found there and the module is read as part of this project: its imports move into the group that holds `proteus`. The lint hook runs `ruff check --fix`, so it rewrites the groups of any file it touches, and the rewrite it commits is what the checks then reject, because the machine running them carries no clone. On macOS the collision is wider than it looks, since the filesystem is case-insensitive and `BOREAS` therefore answers to `boreas`.

Setting the search to the source tree alone covers every module at once rather than one name at a time. With it, `boreas`, `calliope`, `janus`, `mors`, `vulcan` and `zephyrus` all sort as dependencies, which is what they are; `aragog`, `atmodeller` and `zalmoxis` already did, through the pins that are unchanged here.

Narrowing the search leaves the `tests` package outside it, but every file that imports the package by name sits inside it, and ruff reads a file's own package as first party without being told, so nothing sorts differently.

## Validation of changes

ruff caches per file content and settings, and creating or removing a sibling directory changes neither, so every check below was run with `--no-cache`; a cached run returns a stale verdict here.

- Before the change, with the clones present, `ruff check --no-cache src/ tests/` reports 3 errors in `tests/escape/test_escape_boreas.py`; after it, all checks pass
- The round trip no longer inverts: `--fix` on a machine carrying the clones leaves that file byte-unchanged
- Probe files confirm all nine ecosystem modules sort as third party afterwards, while `proteus` stays first party
- Adding the sibling directories to the search instead is not an alternative: pointing it at `tests` as well finds the bare-name helpers and moves 32 other imports
- The `tests` package sorts identically with and without an explicit pin, checked from inside a real package, so no pin was added

The accompanying test asserts the root is not searched, and separately that the modules which can appear there are imported, so it cannot pass by guarding nothing. Its at-risk set is derived from the clone pins and the `fwl-*` distributions rather than from a fixed list, covering eight modules, and it drops standard-library names so `fwl-io` does not become `io`. Removing the setting, or putting the root back on the search path, each fails the test and brings the same 3 errors back.

Test configuration: macOS 15 (Apple silicon), Python 3.12, with the AGNI, aragog, BOREAS, CALLIOPE, JANUS, MORS, SPIDER, socrates, VULCAN, Zalmoxis and ZEPHYRUS clones present at the repository root.

- Unit tier: 2944 passed, 7 skipped
- `ruff check --no-cache src/ tests/ tools/ conftest.py` and `ruff format --check src/ tests/` clean
- `bash tools/validate_test_structure.sh` clean

## Checklist

### DIFF
--- a/tests/test_import_sorting_config.py
+++ b/tests/test_import_sorting_config.py
@@ -4,7 +4,8 @@ The ecosystem modules are installed as clones beside ``src/`` at the repository
 root. Whatever ruff treats as an import search path is also where it looks for
 first-party names, so with the root searched, a clone directory carrying its
 own module name makes that module read as part of this project and its imports
-sort into the group that holds ``proteus``.
+sort into the group that holds ``proteus``. Naming a module in the first-party
+list of the isort settings arrives at the same place by a shorter route.
 
 The damage is not symmetric between a contributor's machine and the checks,
 because only the machine with the module installed carries the clone. The lint
@@ -18,6 +19,14 @@ machine that installed no modules at all, which is the case the checks
 themselves run in. Each derivation carries a canary as well as an assertion,
 so a check that stops reaching its subject fails rather than passing on
 nothing.
+
+``docs/Explanations/test_framework.md`` asks every check for an edge case and a
+path through the error contract, in wording written for the physics modules.
+Read here, the edge cases are the limits a configuration and a scan can reach:
+a spelling of the root nobody wrote down, an import that does not begin its
+line, a candidate set with nothing in it. The contract a failure reports is the
+assertion message, which says what the setting defends rather than only that it
+changed.
 
 References:
   - docs/How-to/testing.md
@@ -40,10 +49,6 @@ pytestmark = [pytest.mark.unit, pytest.mark.timeout(30)]
 REPO_ROOT = Path(__file__).resolve().parent.parent
 SCANNED_ROOTS = (REPO_ROOT / 'src', REPO_ROOT / 'tests')
 
-# Spellings of the repository root as a search path entry. Anything resolving
-# to the root puts the module clones back in front of ruff.
-ROOT_SPELLINGS = frozenset({'.', './', '', str(REPO_ROOT)})
-
 # Directories that belong to this repository rather than to a module beside it.
 REPO_OWN_DIRECTORIES = frozenset({'src', 'tests', 'tools', 'docs', 'examples', 'input'})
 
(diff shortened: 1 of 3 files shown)


nichollsh:
@timlichtenberg, you've requested a review from the PROTEUS maintainer group. Unfortunately, I likely will not have time to thoroughly test and review this PR today. I would suggest requesting a review from @egpbos  or @maraattia .

maraattia:
I can have a look today or tomorrow @timlichtenberg.

tests/test_import_sorting_config.py
```diff
@@ -0,0 +1,209 @@
...
+    collidable = sorted(declared & imported)
+    assert len(collidable) >= 5, (
+        f'only {collidable} of the ecosystem modules are imported by this repository, '
+        'so the search path guard is protecting far less than it appears to and the '
+        'reasoning behind it should be revisited rather than left in place'
+    )
+
+    # Both routes to the set should agree: a clone on this machine that the
+    # configuration does not name would mean one of them has gone stale.
+    unaccounted = sorted(_cloned_module_directories() - declared)
```
maraattia: **warning:** The `unaccounted` check treats every directory at the repository root carrying a `.git` as a module clone, so anything else at the root that happens to be a git repository fails it, whether or not it could change how an import sorts. Two cases I hit while testing this branch:

- A second checkout or a fork beside the pinned ones: creating `AGNI-fork/` gives `AssertionError: ['agni-fork'] stand at the repository root but are named nowhere in the configuration, so the set read from it no longer covers what is installed`.
- A git worktree placed inside the root: `wt-experiment/` gives the same failure. A worktree's `.git` is a file rather than a directory, and `Path.exists()` accepts it.

Both turn the unit tier red on a contributor's machine for a directory that has nothing to do with their change, and the message sends them to the configuration when the cause is on their filesystem. In CI the check cannot fail: the unit jobs do clone, through `setup-proteus`, but they clone only the pinned set, which is by construction what the configuration names. That is worth keeping, since it means the check does real work in CI against a module that appears at the root without being named in the configuration. The narrowing below keeps the part of that which can affect sorting and drops the rest.

The hazard the check is after is narrower than "any repository at the root": it is a directory whose name this repository imports and the configuration does not name. Intersecting with the scan the test already has keeps that case and drops the rest:

```python
unaccounted = sorted((_cloned_module_directories() & imported) - declared)
```

`imported` holds names as written and the clone set is lowercased, so the intersection stays conservative on a case-insensitive filesystem, which is the case that matters. With this, a new ecosystem module cloned at the root and imported but not yet pinned still fails, which is the staleness the docstring describes, while `AGNI-fork` and a worktree pass because neither can shadow a name.

tests/test_import_sorting_config.py
```diff
@@ -0,0 +1,209 @@
...
+REPO_OWN_DIRECTORIES = frozenset({'src', 'tests', 'tools', 'docs', 'examples', 'input'})
+
+
+def _read_config() -> dict:
+    """Return the parsed ``pyproject.toml``."""
+    return tomllib.loads((REPO_ROOT / 'pyproject.toml').read_text(encoding='utf-8'))
+
+
+@cache
+def _imported_root_names() -> frozenset[str]:
```
maraattia: **warning:** `_imported_root_names` walks `src/` and `tests/` with `rglob('*.py')` and AST-parses every file. Measured with `--durations` on this branch, `test_the_modules_beside_the_source_are_the_ones_that_would_be_misread` takes 0.98 s, against the 100 ms budget `docs/How-to/testing.md` sets for the unit marker. The `timeout(30)` at line 36 is the right ceiling for the tier, but the ceiling and the budget are different numbers, and the docstring at line 61 records the cost without reckoning it against the budget.

Three ways out, in increasing order of effort: parse the tree once at module scope rather than inside a test, so the cost lands in collection rather than in a timed test; narrow what gets parsed, since the only names that matter are the ecosystem ones and a file containing none of them as a substring cannot contribute; or keep the cost and say in the docstring why the tier budget is being exceeded, so the next reader does not take it for an oversight. The middle option is the one that scales, given the tree only grows.

While on the file: `docs/Explanations/test_framework.md` also asks every new test function for an edge case and a path exercising the error contract, which two configuration assertions do not have. Those rules read as written for physics code, so I would not hold the PR to them here, but it is worth a sentence in the module docstring saying that is the reading you took.

tests/test_import_sorting_config.py
```diff
@@ -0,0 +1,209 @@
...
+
+The ecosystem modules are installed as clones beside ``src/`` at the repository
+root. Whatever ruff treats as an import search path is also where it looks for
+first-party names, so with the root searched, a clone directory carrying its
+own module name makes that module read as part of this project and its imports
+sort into the group that holds ``proteus``.
+
+The damage is not symmetric between a contributor's machine and the checks,
+because only the machine with the module installed carries the clone. The
+formatter runs with ``--fix``, so on that machine it rewrites the import groups
```
maraattia: **nit:** The module docstring here, and the test docstring at line 125, attribute the rewrite to the formatter running with `--fix`. The hook that rewrites import groups is the `ruff` check hook, configured with `args: [--fix]` in `.pre-commit-config.yaml`; the `ruff-format` hook is commented out there, and `ruff format` does not sort imports in any case. A reader chasing "the formatter" lands on the disabled hook and cannot reproduce the story. Naming `ruff check --fix`, or "the lint hook", would keep the explanation reproducible. The same wording is in the pull-request description.

tests/test_import_sorting_config.py
```diff
@@ -0,0 +1,209 @@
...
+        'root, where the ecosystem module clones are installed'
+    )
+
+    configured = [str(entry) for entry in ruff['src']]
+    on_the_root = sorted(entry for entry in configured if entry.strip() in ROOT_SPELLINGS)
+    assert not on_the_root, (
+        f'{on_the_root} put the repository root on the import search path, where the '
+        f'module clones are installed. Clones present here: {sorted(_cloned_module_directories())}'
+    )
+    assert configured == ['src'], (
```
maraattia: **nit:** The two assertions above already cover the stated contract, that nothing on the search path resolves to the root. This one additionally forbids any other entry, so if `tools/` ever becomes an importable package and wants to be on the path, the natural one-line configuration change fails a test whose stated concern is the root. That may well be the intent, in which case it would help to say so in the message: as written it reads as though any other entry were a defect, rather than a decision worth revisiting. The same applies to the `boreas in imported` canary at line 182, which will report that nested imports are escaping the scan if BOREAS support is ever dropped or the import is renamed.

maraattia (dismissed):
Thanks for chasing this one down. Pointing the search path at the source tree closes it for every module at once instead of one name at a time.

I reproduced it before reading your validation. With a lowercase `boreas` directory at the root and the setting removed, `ruff check --no-cache --select I src tests tools conftest.py` gives your 3 errors in `tests/escape/test_escape_boreas.py`, and restoring the setting clears them with the directory still there. A throwaway probe holding adjacent `import aragog` and `import proteus` statements isolates the classification: with the root searched and the pins removed, `aragog` reads as first party purely because a directory of that name stands there, and with the pins present it stays third party either way, so your comment about pinned names is right. On a clone-free tree both configurations are clean. Your `tests` argument holds from the other side too: `tests/__init__.py` is present, every import of the package by name sits inside it, the root `conftest.py` imports only the standard library, and `tools/` has neither an `__init__.py` nor a self-import. The new test passes here bare and with the nine clones at the root, fails when I delete the setting or point it back at the root, and its at-risk set comes out as your eight modules. The structure validator is clean.

Two things to tighten, neither blocking, both on the new test. The cross-check on root-level repositories fires on directories that cannot affect sorting: a fork or a second checkout fails it, and so does a git worktree placed there, since a worktree's `.git` is a file and the existence test accepts it. CI is safe, because the unit jobs clone only the pinned set, so the failure lands on contributors, and there the message blames the configuration. Intersecting the clone set with the names the scan already collected keeps the case you want and drops the rest; the one-line form is on that line. Separately, the scan parses every `.py` file under `src/` and `tests/`, which measures 0.98 s here against the 100 ms unit budget in the testing guide. Your 30 s timeout is the right ceiling for the tier, but the budget is a different number and this is about ten times it.

Two nits: the docstrings credit the formatter running with `--fix`, but the hook that rewrites import groups is the `ruff` check hook, and `ruff-format` is commented out and would not sort imports in any case; and the equality assertion locks the value rather than the property the docstring states, which is worth a word in the message if that is deliberate.

Approving. Not for this pull request: with the root off the search path the `known-third-party` pins no longer do any work, which I checked with lowercase directories of all three at the root. Keeping them as a safety net is fine, but worth a comment saying that is what they are for. Happy to look again if you rework the cross-check.

timlichtenberg:
Thanks for the two warnings, both are closed. The check no longer treats any directory at the root as a module clone, so a fork checkout or a git worktree at the root passes; I confirmed the worktree case by running the suite from inside one, where `.git` is a file. The scan is down from about a second to a fifth of that, by asking only for the names the assertions turn on and reading the tree once.

On the two smaller points: the rewrite is attributed to the hook running `ruff --fix`, and the assertion you flagged is gone, replaced by one that measures the parse the guard prevents rather than a result that comes back empty either way.

A second file now runs ruff itself, since the first only reasons about what ruff would do.

PR #833, opened 2026-08-19
Bound the apt-get retry in CI setup with a timeout

timlichtenberg:
## Description
`apt-get update`/`apt-get install` in the Linux setup action can stall mid-transfer against the Ubuntu mirrors, and an unbounded call once consumed an entire nightly job. This wraps both calls in a timeout with one retry and backoff, and repairs any dpkg state a killed attempt leaves behind before retrying.

## Validation of changes
I traced the retry loop by hand and confirmed its bash syntax with `bash -n` on the extracted script block. I also bounded the `dpkg --configure -a` repair call with its own timeout, since it originally had none and could hang the job indefinitely on a wedged dpkg lock, defeating the point of the surrounding retry wrapper; and I corrected a comment that claimed the worst case stayed well inside a 30-minute job budget, which does not hold for every job that calls this action.

## Checklist

### DIFF
--- a/.github/actions/setup-proteus/action.yml
+++ b/.github/actions/setup-proteus/action.yml
@@ -108,8 +108,40 @@ runs:
       shell: bash
       run: |
         set -euo pipefail
-        sudo apt-get update -qq
-        sudo apt-get install -y --no-install-recommends \
+        # apt-get talks to the Ubuntu archive mirrors, which occasionally stop
+        # responding mid-transfer instead of failing outright; an unbounded
+        # call once consumed the whole job waiting on a stalled connection.
+        # Bound each call with a timeout and retry with backoff, in the same
+        # spirit as the AGNI spectral data fetch below (a separate step, so
+        # the loop is duplicated here rather than shared). One retry keeps
+        # the worst case for this whole step to a few minutes, which matters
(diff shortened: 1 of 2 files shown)


nichollsh (commented):
Overall looks good, although I have a couple of optional suggestions which you might consider:

1. Check if the Ubuntu package mirrors are accessible before performing the apt-get call. If the mirrors are down, it would be worth updating them or exiting the run, since further attempts to access them won't succeed either.

2. Wrap the MacOS brew calls with a similar loop. If this can happen for the Ubuntu packages, we can expect it to happen for MacOS packages too. Perhaps a "get packages" script called by the GH action step would be a cleaner approach for handling this, instead of having separate (and long) steps in this action.

What do you think?

timlichtenberg:
@nichollsh Both done (tip 3b8a08c1). The Linux step now probes the mirrors actually configured on the runner before the apt-get call and fails fast with a clear message if none answer. The macOS brew calls are wrapped with the same bounded retry, and both steps now source one shared `retry-with-timeout.sh` instead of duplicating the loop.

Ready for another look.

nichollsh (approved):
Changes look good! Thank you for implementing these. Happy for this to be merged.

PR #841, opened 2026-08-31
Bump the fwl-io floor to 26.8.31

timlichtenberg:
## Description

Bump the `fwl-io` floor from `>=26.7.25` to `>=26.8.31` so PROTEUS guarantees the fixes released in fwl-io 26.8.31: the Dataverse mirror 2xx-bad-body fix (https://github.com/FormingWorlds/fwl-io/pull/37), the `fwl.`-prefix logger fix (https://github.com/FormingWorlds/fwl-io/issues/34), and the `_request` docstring and success-path test (https://github.com/FormingWorlds/fwl-io/pull/38). Release: https://github.com/FormingWorlds/fwl-io/releases/tag/26.08.31 (published on PyPI as 26.8.31, the normalized CalVer form).

The adjacent comment is updated to describe the current pin.

## Validation of changes

Confirmed fwl-io 26.8.31 is live on PyPI, and `pip install --dry-run "fwl-io>=26.8.31"` resolves cleanly in a PROTEUS conda env ("Would install fwl-io-26.8.31", all sub-dependencies already satisfied). macOS with Python 3.12.

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -82,11 +82,11 @@ dependencies = [
     "sympy",
     "astropy",
     "toml",
-    # 26.7.25 bounds the time a fetch can take: every request carries an
+    # This floor bounds the time a fetch can take: every request carries an
     # explicit connect and read timeout, and a transient failure is retried
     # with backoff instead of ending the fetch. Below that floor a stalled
     # mirror hangs the start of a run with no diagnostic.
-    "fwl-io>=26.7.25",
+    "fwl-io>=26.8.31",
     "zenodo-get",
     "zenodo-client>=0.4.1",
     "osfclient",


egpbos (dismissed):

egpbos:
The tests that are failing trip over a hardcoded specific version string in `src/proteus/data/__init__.py`, `FWL_IO_FLOOR`, which is set to the version of fwl-io. The comment notes that this string refers to a schema version. Is that really the case? Is the data schema intended to change with every fwl-io version? If not, a solution would be to separate the schema versioning from the fwl-io versioning. We already have a configuration file version, something similar would make sense for the output as well. I would go for just a single version integer.

timlichtenberg:
You're right, the constant marks the manifest schema floor. 26.8.31 changed nothing in that schema (request timeouts, retries, a logger fix), so the constant stays at 26.7.25 and the test now checks that the pyproject floor is at least the schema floor instead of equal to it. That keeps the upgrade message accurate without touching two files on every fwl-io release.

On a separate schema version: fwl-io 26.08.31 added an optional manifest_schema key, so the single integer you suggested now exists on the fwl-io side. I'll open an issue for declaring it in the PROTEUS manifest rather than fold it into this pin bump.

Issue #744, opened 2026-07-14
Status file not consistently updated before exiting or throwing exceptions


nichollsh:
## Problem description

Sometimes PROTEUS crashes without updating the status file. This is misleading, especially when running grids.

## Expected behavior

The file should be updated appropriately, in all cases. 

## Relevant people

@rdc49

egpbos:
Quick note: exit handling?

nichollsh:
Can build upon Karen's work with implementing a Failure State python class in per recent PR: #876

Issue #241, opened 2024-10-29
An multi-column mode (atmosphere and interior)


nichollsh:
All the modelling in PROTEUS is done under the assumption that the planet can be globally-averaged into a single 1D column. However, we could extend this by simulating multiple columns around the planet's equator, with a description of heat transport between them. This has been done before with models of exoplanet atmospheres, but never for a magma ocean model.

This would allow us to capture day-night differences on tidally locked planets, including the potential for nightside clouds, and also could allow comparison with phase curve observations.

nichollsh:
https://arxiv.org/pdf/2505.13066

PR #853, opened 2026-09-07
Set AGNI parameters (hydrostatic integration, spin rate, geometry) and ensure optical depth is recorded

### LINKED ISSUE #444: Effect of planetary spin on its structure
Planetary rotation rate and oblateness impact the structure of its interior and atmosphere. For young planets, such as those modelled by PROTEUS, this could potentially have a systematic impact when comparing modelled radii to observations.
e.g. https://arxiv.org/pdf/2507.02112

Handling this can be done across a wide range of complexities. However, it will require modifications in atmosphere modules and interior/structure modules.

Potentially related to #390 and #63

### LINKED ISSUE #679: Pass planet spin rate information to AGNI
## Feature description
AGNI version 1.10 accounts for the atmospheric inflation effect induced by outward centrifugal acceleration offsetting the inward gravitational acceleration. Calculating this offset depends on the planet axial rotation period (day length) and the latitude of the modelled column. PROTEUS should pass these parameters to AGNI, since they are assumed to be 'disabled' otherwise. More important for fast rotating planets with low MMW envelopes.

## Preferred solution
Set the `axial_period` (seconds) and `latitude` (degrees) parameters when calling `AGNI.atmosphere.setup!` in the PROTEUS `agni.py` wrapper.

## Additional information
Related to ongoing issues #444 and #567 and #484

nichollsh:
## Description

This pull request reworks some of the ways that PROTEUS interfaces with AGNI. It also comes with a substantial performance improvement by making  the global variables in AGNI constants, or making them function keyword arguments.

* Added `longitude` and `latitude` as tracked and configurable fields. These are passed through the configuration, included in the helpfile, and handled in both initialization and update routines. Also passes spin rate information from the `axial_period` column, which closes #679 and closes #444. This is a necessary step before moving towards #241.

* Introduced new configuration options in the config workflow for hydrostatic integration (`hydrograv_steps`, `hydrograv_maxdr`, `hydrograv_mindr`, `hydrograv_ming`, `hydrograv_constg`, `hydrograv_selfg`) and ensured these are passed to AGNI during initialization. Updating these will certainly help with the performance and convergence issues encountered by @timlichtenberg and @rdc49  for reducing conditions; e.g. by setting `hydrograv_constg=True`.

* Refactored schema validation: replaced the custom `AgniSchemaMismatch` exception with a standard `RuntimeError` and generally improved error reporting in the atmosphere climate wrapper. Ensured that all `raise` calls first update the status file (related to #744).

* Updated calls to AGNI solver routines to reflect changes in keyword arguments and improved the interface for energy and flux calculations. 

* Fix bug where PROTEUS was not requesting that AGNI return optical depths, so they were being stored as zero. 

## Performance improvement

I am pleased to say that this massively improves the performance of AGNI in my tests. The default AGNI config runtime reduced from 36 seconds to 24 seconds. The options should also enable more stable calculations under reducing conditions.

## Validation of changes

* AGNI tests pass on my laptop - Fedora 44, Python 3.12, Julia 1.12
* AGNI tests pass on GitHub CI - Ubuntu and MacOS
* PROTEUS tests pass on my laptop
* PROTEUS unit tests pass on GitHub CI

## Checklist


timlichtenberg (changes requested):
Thanks Harrison. The AGNI interface rework, the hydrostatic-integration options, and the schema-exception cleanup all look coherent, and the test additions are strong (discrimination guards, the status-code-22 assertion, the new retry-ladder test). A few points before merge, one of which touches a physics path.

## Before merge

**1. `read_ncdf_profile` gravity fallback silently feeds a physics path.** In `common.py`, a missing `gravity` variable now logs and falls back to `g = np.zeros_like(p)` instead of raising `KeyError`. This profile dict is not plotting-only: `read_atmosphere_data` feeds `observe/petitRADTRANS.py`, where `_get_reference_prt_values` reads `atm['g']` as `reference_gravity`. A zero-filled gravity would pass through as `reference_gravity = 0.0` with no error, which is unphysical and hard to trace back. (The other non-plotting consumer, `atmos_chem/vulcan.py`, reads the same dict but takes gravity from `hf_row['gravity']`, so it is not affected.) Suggest one of: keep the hard failure for the observe path; restrict the zeros fallback to the plotting callers; or have `petitRADTRANS` detect the sentinel and skip. At minimum, please state in the PR description why the fallback is safe for the observe consumer.

## Non-blocking

**2. `ncdf_flag_to_bool` now returns `None` and logs instead of raising `ValueError`.** The caller coerces `None` to `False`, so an unparseable flag byte becomes a silent `False` rather than a hard stop. That is defensible for a batch read of old archives, but it is a behaviour change not mentioned in the PR description. Please note it there so the softening is a deliberate decision rather than an incidental one.

**3. `longitude` / `latitude` are tracked but not configurable in this diff.** The description says they are "tracked and configurable" and "passed through the configuration", but the only config additions here are the `hydrograv_*` fields; `run_orbit` sets `longitude = 0.0` and `latitude = 0.0`. That is fine as a first step toward #241, but the description overstates the current state. One further point: the assignment sits in the non-satellite branch, so with `orbit.satellite = true` `run_orbit` skips it and both columns fall back to the `ZeroHelpfileRow` zero-fill (still 0.0, so no crash). Is that asymmetry intended, or should the two lines move above the `if config.orbit.satellite:` split so the value is set the same way on both paths?

**4. The rationale comment on the bare `except Exception` around `solve_energy_b` was removed.** The `try/except` itself is retained, but the comment that explained why it exists (AGNI's failure-path `plot_step` raising `InexactError` on a NaN flux, which would otherwise kill the Python process and bypass the deadlock detector) is gone. That "why" is not obvious from the code. The new `test_solve_energy_exhausts_attempts_on_repeated_exception` documents the behaviour, which helps, but a one-line comment naming the reason would keep it discoverable at the call site.

## Questions

- The description says the optical-depth fix comes from requesting AGNI to return the values. In the diff the only related change is `calc_cf=True` on `calc_fluxes_b`. Can you confirm `calc_cf` is what populates the optical-depth field, and that the stored-as-zero symptom is resolved in a run (not only in the unit stubs)? I could not verify the `calc_cf` to optical-depth link from the PROTEUS side alone.
- `update_agni_atmos` now pushes `grav_surf`, `rp`, `interior_mass`, `axial_period`, `col_lon`, and `col_lat` onto the AGNI struct every step. None of these were set on the struct from the PROTEUS side before (AGNI kept its init-time geometry through the run). For `grav_surf` and `rp` in particular, which evolve as the interior contracts, does pushing the per-step value change any coupled result relative to holding the init-time value?

Codecov reports full coverage on the modified lines, and the pins are bumped consistently in both `pyproject.toml` and `docs/Reference/module_versions.md`.

## What I checked and did not

I read the full diff, traced the `read_ncdf_profile` and `read_atmosphere_data` consumers, and confirmed the `try/except` retention. I did not run the unit suite or a live AGNI run, so the `calc_cf` optical-depth fix and the `golden_run.tsv` numeric deltas are unverified by me, as is the 36 s to 24 s performance number in the description. The `esc_*` changes in the golden run are below the 1e-6 relative level, consistent with the SOCRATES and AGNI pin bumps rather than a logic change, but I did not reproduce them.

timlichtenberg:
Too late I guess. :>

Issue #843, opened 2025-11-03
Self-consistent solid- and liquid-tides model within PROTEUS


timlichtenberg:
Implementation roadmap for a self-consistent tidal model in PROTEUS that handles various tidal regimes and orbital evolution of the Earth-Moon system and a Star-Exoplanet system.

Issue #670, opened 2026-04-21
Add Obliqua to `proteus doctor` command and PROTEUS version validation scheme


nichollsh:
## Feature description
PROTEUS checks that submodules are sufficiently up to date. When the first release of  Obliqua is made, it should also be incorporated into this version checking scheme within PROTEUS.

## Preferred solution
Should be straight forward to follow the same approach taken for AGNI, since both are Julia-based codes.

## Additional information
Suggested in #669 

## Relevant people
@MarijnJ0 @egpbos @timlichtenberg

Issue #484, opened 2025-08-08
Axial evolution of planet, interactions with tides


nichollsh:
With orbit-raised tides generated within the planet, it should soon be possible to self-consistently model the rotational (axial) state of the planet (its day length). Meaning that tidal locking should come about naturally as a result of the model, rather than having a user prescribe it. 

Currently in PROTEUS the axial period of the planet doesn't matter, because the model is 1D and the user specifies a fixed zenith angle.Ideally, the zenith angle would be calculated from the axial period (issue #63 covers this).

I propose that we include some calculation in PROTEUS which evolves the `axial_period` based on the tidal calculation. The user would specify an initial value, and whether it should evolve or stay constant in time.

Related to #63, #444, #450, #449, #241. I think Mara has also done some work related to this?

@timlichtenberg @MarijnJ0

nichollsh:
This issue is following on from my conversation with Marijn this morning. The physics here can easily get very complicated, so we should be careful in working out what to do first.

nichollsh:
> Does this directly affect the implementations in issues https://github.com/FormingWorlds/PROTEUS/issues/449 and https://github.com/FormingWorlds/PROTEUS/issues/450?

This problem came up because the integration of the satellite orbit following Korenaga+23 includes evolution of the planet's axial period.

However, it's not clear to me how this would interact with orbital tides which also modulate the planet's day length. Particularly when star-planet tides are acting to main tide-locked synchronous rotation, but then you also might have a satellite acting on the planet. This is the evection resonance state. So it doesn't matter so much for #449 but more for #450.

That being said, it would also matter if orbital evolution pulled the planet closer to the star, and allowed star-planet tides to induce tide-locked synchronous rotation.

timlichtenberg:
If I understand correctly you want to enable both stellar- and satellite-induced tides at the same time? Indeed I hadn't planned this to be the case initially, instead focusing on either the one or the other case. However, if you think you can find a solution then please go ahead ofc.

PR #859, opened 2026-09-08
Md/obliqua

### LINKED ISSUE #532: Tidal heating in liquid/melt phase of mantle
Potentially using model of Farhat+24, implemented alongside existing LovePy calculation.

Making an issue for this so we can specifically track it. @MarijnJ0 @timlichtenberg

### LINKED ISSUE #578: Record tidal heat fluxes as a function of longitude/latitude
## Feature description
The spatial distribution of tidal heating on Io is numerically modelled by [Segatz+88](https://www.sciencedirect.com/science/article/abs/pii/0019103588900012). The heat fluxes are highly anisotropic - in some cases being maximised in the polar regions, and other times nearer the equator, depending on mantle structure. This may be seen as consistent with widely distributed hot spots observed on Io.

It would be worthwhile storing/plotting our own spatial distributions of tidal heating when using our [own models ](https://github.com/FormingWorlds/Love.jl), which already calculate the spatial distribution of tidal heating. These data would probe the validity of our assumptions, given that we mostly consider these bodies in 1D, while also providing pretty plots for proposals/websites/etc.

## Preferred solution
Store the lat-long heating rates from Love.jl in an appropriate file format, at each time step. Add a function in PROTEUS for visualising these.

## Additional information
Example from Segatz et al:
Image

## Relevant people
@timlichtenberg @MarijnJ0

### LINKED ISSUE #743: Self-consistent orbital evolution of Earth-Moon system
https://github.com/FormingWorlds/PROTEUS/tree/md/Obliqua

### LINKED ISSUE #801: Self-consistent orbital evolution of Star-Exoplanet system

MarijnJ0:
## Description
The PR introduces the Obliqua tidal module into PROTEUS, along with three compatible orbital dynamics models. The implementation adds:
- Obliqua tides option, allowing for multiphase tidal calculation in 1D for the primary (planet).
- Obliqua tides lookup tables, allowing the user to provide pregenerated lookup tables/interior data files, and extract/obtain associated full Love number spectra.
- Star-Planet orbital model with spin-orbit dynamics and eccentricity evolution (planet-only).
- Planet-Satellite orbital model with SO+e for both primary and perturber.
- Planet-Satellite orbital model with SO+e for both primary and perturber + evection resonance.
- For these orbit models, a hansen.py file is added, which computes a complete lookup table. This removes any overhead away from the solver, and speeds up the integration.
- The orbit models (bare ODEs) are all evolved through a shared time stepping function, which ensures proper convergence and can be tweaked in the config.
- Time stepping is handled using three clocks: PROTEUS clock, set overall coarse time stepping; solver clock, handles the model integration with event detection for ER; storage clock, controls what time steps/data is saved to disk.
- Angular momentum conservation holds, and additional changes in the planet's gyration parameter are smoothly cancelled by AM conserving changes in the spin.
- New plotting scripts are added to track the Love number evolution, the orbit (primary+perturber), and the ER.

I added Obliquato the docs and also changed some of the phrasing in the docs, specifically things like:
- 'Obliqua evolves the orbit' --> 'Obliqua models tides, PROTEUS internally evolves the orbit.' 
- 'Lovepy is multi-phase' --> 'Lovepy is solid-phase'

I tried keeping to the 500 line file limit, but this was not possible for the satellite.py file.

I added basic unit and smoke tests, and one particular case that recreates figure 4 in Canup & Rufu (2020). This takes a little while to run, but is a strong model validation test.

## Issues

Obliqua should still be added to PROTEUS doctor (issue #670)

Closes #578, #484, #743, #801

Directly related #843

## Validation of changes
The unit tests pass on my machine: Ubuntu 24.04 LTS (noble) with Python 3.12.11

I also have some example output from the ER capture in PROTEUS. Basic orbital dynamics has been tested, and AM remains conserved even at the 1e9 yr mark.

## Checklist

@timlichtenberg @nichollsh


MarijnJ0:
@timlichtenberg and @nichollsh, 

This PR is ready for review. I added the relevant functions/code for integrating Obliqua into PROTEUS along with three compatible orbital dynamics models. The PR also introduces the satellite object with its own orbital parameters. Some small changes where made to the existing docs, and the docs are further extended to include the new functionality. Obliqua has been added to the install/test/doctor scripts, and is treated similar to AGNI, but as an optional module. Unit tests where added to achieve complete coverage over the newly code. 

The final test still seems to fail since there have been a handfull of new merges into main, I will update the PR to include this when it is ready to be merged.

nichollsh:
Thanks @MarijnJ0, I'll try to take a look at this tomorrow.

docs/Explanations/model.md
```diff
@@ -45,7 +45,8 @@ atmosphere), enabling hierarchical model intercomparison.
 | Star | [MORS](https://proteus-framework.org/MORS/), dummy | Stellar evolution and spectrum |
 | Escape | [ZEPHYRUS](https://github.com/FormingWorlds/ZEPHYRUS), dummy | Atmospheric escape |
 | Outgassing | [CALLIOPE](https://proteus-framework.org/CALLIOPE/), [atmodeller](https://github.com/djbower/atmodeller), dummy | Volatile exchange between interior and atmosphere |
-| Orbit | [Obliqua](https://github.com/FormingWorlds/Obliqua), dummy | Orbital evolution and tidal heating |
+| Tides | [Obliqua](https://github.com/FormingWorlds/Obliqua), [Lovepy](https://github.com/nichollsh/LovePy), dummy | Tidal response of the planet (Love numbers, heating) |
```
nichollsh, line 48: This should point to the Obliqua docs page

input/all_options.toml
```diff
@@ -61,6 +61,7 @@ config_version = "3.0"
         max_growth_factor     = 0.0          # cap on dt growth ratio between consecutive steps; 0 = disabled
         mushy_maximum         = 1.0e6        # max dt [yr] while stop.solid.phi_crit < Phi < mushy_upper (0 = disabled); caps steps through the mushy band so the integrator resolves the solidification front rather than overstepping it
         mushy_upper           = 0.99         # mushy regime upper bound (melt fraction)
+        evection_maximum      = 0.0          # max dt [yr] while hf_row['in_evection_band'] (0 = disabled); caps steps while the planet-satellite system is inside the evection resonance
```
nichollsh, line 64: This should not be zero by default, surely? Since this would enforce `dt=0`.

input/all_options.toml
```diff
@@ -99,6 +100,10 @@ config_version = "3.0"
             spin_enabled      = true         # check rotational breakup
             offset_spin       = 0            # correction to breakup period [s]
 
+        [params.stop.satellite]              # satellite escape criterion
```
nichollsh, line 103: Again, I feel like `sma_max` should also not be zero, since enabling this would immediately cause the simulation to terminate.

input/all_options.toml
```diff
@@ -212,25 +217,62 @@ config_version = "3.0"
...
+
+    # Satellite
+    [orbit.satellite]
+        include_satellite     = false        # include satellite
+        mass_sat              = 0.012        # satellite mass [M_earth]
+        radius_sat            = 0.273        # satellite radius [R_earth]
+        axial_period_sat      = "none"       # day length [hours]; none = tidally locked
+        semimajoraxis_sat     = 1.33e-4      # satellite orbit semi-major axis [AU]
+        eccentricity_sat      = 0.0          # satellite initial orbital eccentricity [dimensionless]
+        evection_angle        = 0.0          # evectionangle [degrees]
```
nichollsh, line 248: Initial evection angle?

input/all_options.toml
```diff
@@ -212,25 +217,62 @@ config_version = "3.0"
...
+        c_factor_sat          = 0.4          # satellite normalized moment of inertia factor [dimensionless] (Note: 0.4 = uniform density sphere, < 0.4 = differentiated body; see https://en.wikipedia.org/wiki/Moment_of_inertia_factor)
```
nichollsh, line 249: Point to somewhere in the PROTEUS docs here, rather than to wikipedia. The docs page can point to wikipedia if necessary, and ideally suggest how this parameter would be chosen.

input/all_options.toml
```diff
@@ -212,25 +217,62 @@ config_version = "3.0"
...
+
+        max_rel_da             = 0.01        # max tolerated relative change in semi-major axis per substep
+        max_rel_de             = 0.01        # max tolerated relative change in eccentricity per substep
+        max_rel_dOmega         = 0.02        # max tolerated relative change in a spin rate per substep
+        de_floor               = 0.05        # floor on the eccentricity-change-ratio denominator
+        max_substeps           = 10_000_000  # maximum number of substeps attempted per call
+
+        resonance_margin_enter = 0.10        # evection-band entry margin (ps1d_evec only)
+        resonance_margin_exit  = 0.50        # evection-band exit margin (ps1d_evec only)
+        fine_csv_target_rel_dt = 0.01        # target storage-clock spacing for out-of-band fine samples (ps1d_evec only)
```
nichollsh, line 273: These need to be better described here and in the documentation

input/all_options.toml
```diff
@@ -240,6 +282,87 @@ config_version = "3.0"
         visc_thresh           = 1e9          # minimum viscosity for tidal heating [Pa s]
         ncalc                 = 1000         # grid points for tidal calculation
 
+    [orbit.obliqua]
+        store_3D              = false        # store 3D tidal heating profile
+        enforce_ec            = true         # enforce eccentricity in tidal heating calculation
```
nichollsh, line 287: What happens if this is set to false?

input/all_options.toml
```diff
@@ -240,6 +282,87 @@ config_version = "3.0"
...
+        optimize_scales       = false        # optimize tidal heating scales for speed
+        solid_shell           = true         # use solid shell model for tidal heating
+
+        min_frac              = 0.02         # minimum segment radius fraction before smoothing [dimensionless]
+        visc_l                = 1e2          # pure liquid viscosity [Pa s]
+        visc_lus              = 5e5          # liquidus viscosity to use [Pa s]
+        visc_s                = 1e22         # pure solid visosity [Pa s]
+        visc_sus              = 5e5          # solidus viscosity to use [Pa s]
```
nichollsh, line 295: Don't we already have parameters for these in the interior_energetics section of the config?

input/all_options.toml
```diff
@@ -240,6 +282,87 @@ config_version = "3.0"
...
+        n                     = [2]          # power of the radial factor (goes with (r/a)^{n}, since r<<a only n=2 contributes significantly)
+        m                     = [0, 2]       # harmonic of the true anomaly. m=2 corresponds to the semidiurnal tide, m=1 diurnal tide
```
nichollsh, line 297: Why are these arrays?

input/all_options.toml
```diff
@@ -240,6 +282,87 @@ config_version = "3.0"
...
+        k_min                 = "none"       # minimum tidal mode range (k is the Fourier index in mean anomaly), set <nothing> to obtain tidal mode range from eccentricity relation
+        k_max                 = "none"       # maximum tidal mode range (k is the Fourier index in mean anomaly), set <nothing> to obtain tidal mode range from eccentricity relation
```
nichollsh, line 299:
```suggestion
        k_min                 = "none"       # minimum tidal mode range (k is the Fourier index in mean anomaly), set "none" to obtain tidal mode range from eccentricity relation
        k_max                 = "none"       # maximum tidal mode range (k is the Fourier index in mean anomaly), set "none" to obtain tidal mode range from eccentricity relation
```

input/all_options.toml
```diff
@@ -240,6 +282,87 @@ config_version = "3.0"
...
+        # - "solid1d-relax"       : Same as solid1d, but uses the relaxation method to obtain a solution
+        #                           (instead of the shooting method) which is generally more stable.
+        #                           However, this method is generally slower.
+        # - "solid1d-mush"        : Same as solid1d but includes a partially molten/porous ("mushy") region
+        #                           with evolving rheology between solid and liquid behavior.
+        # - "solid1d-mush-relax"  : Same as solid1d_relax but includes a partially molten/porous ("mushy") region
+        #                           with evolving rheology between solid and liquid behavior.
+        # - "solid1d-equil-relax" : Same as solid1d_relax but limited to the fully fluid interior case, specifically
+        #                           designed for the equilibrium tide. Automatically selected when using the other
+        #                           solid1d* models when the forcing frequency is less than the 1/Hubble time.
```
nichollsh, line 325: This needs to go into the documentation; point the user there instead.

input/all_options.toml
```diff
@@ -443,6 +564,8 @@ config_version = "3.0"
         atm_heat_capacity_const   = true         # use the constant fallback heat capacity above
         silicate_density          = 4103.0       # silicate density [kg/m^3]
         core_density              = 10738        # core density if not defined by structure module [kg/m^3]
+        core_shear                = 1.0e-1       # core shear modulus [Pa]
+        core_bulk                 = 5.0e11       # core bulk modulus [Pa]
```
nichollsh, line 568: Where do these values come from?

src/proteus/config/_orbit.py
```diff
@@ -57,6 +59,194 @@ class Lovepy:
...
+    Attributes
+    ----------
+    b_width: float
+        Scale width of the bottom heating decay profile [dimensionless].
+    t_width: float
+        Scale width of the top heating decay profile [dimensionless].
+    """
+
+    b_width: float = field(default=5e-1)
+    t_width: float = field(default=3e-2)
```
nichollsh, line 114: Please add validators on these

src/proteus/utils/coupler.py
```diff
@@ -1744,7 +1759,10 @@ def UpdatePlots(hf_all: pd.DataFrame, dirs: dict, config: Config, end=False, num
     from proteus.plot.cpl_global import plot_global
     from proteus.plot.cpl_interior import plot_interior
     from proteus.plot.cpl_interior_cmesh import plot_interior_cmesh
-    from proteus.plot.cpl_orbit import plot_orbit
+    from proteus.plot.cpl_orbit import (
+        plot_Lovenumber,
```
nichollsh, line 1763: Function names should be lowercase.

src/proteus/orbit/common.py
```diff
@@ -0,0 +1,486 @@
...
+
+
+@dataclass
+class TidalInteraction:
+    primary: Any
+    perturber: Any
+
+    nmk: Optional[NDArray[np.floating]] = None
+    sigma: Optional[NDArray[np.floating]] = None
+    LNk: Optional[NDArray[np.floating]] = None
```
nichollsh, line 25: This needs comments/documentation. What are these variables for and is it necessary to give them a dataclass?

src/proteus/orbit/common.py
```diff
@@ -0,0 +1,486 @@
...
+            self.interactions.append(interaction)
+            return interaction
+
+    def get(self, primary, perturber):
+        for interaction in self.interactions:
+            if interaction.primary == primary and interaction.perturber == perturber:
+                return interaction
+        raise KeyError(f'No tidal interaction: {primary} <- {perturber}')
+
+    def add_from_file(self, primary, perturber, file_path: str):
```
nichollsh, line 46: Again, needs docs and comments.

src/proteus/orbit/common.py
```diff
@@ -0,0 +1,486 @@
...
+
+        return interaction
+
+
+def get_C_planet(hf_row: dict, config: Config, interior_o: Interior_t):
+    """Compute the planet's principal moment of inertia (C_planet) based on the interior structure.
+
+    Note: This function should live in the interior_energetics module, but is currently here for
+    convenience. It may be moved in the future. When moving this, ensure that the smoothing of the
+    structural C_planet update in run_adaptive_orbit_substeps() is preserved.
```
nichollsh, line 66: Please move it into `interior_energetics` now, rather than later.

src/proteus/orbit/common.py
```diff
@@ -0,0 +1,486 @@
...
+    tests). The two effects compose without either overriding the other:
+    this ramp only ever conserves `C_planet * Omega_p` across the
+    non-tidal, non-resonance structural change, exactly as the
+    unsmoothed single-jump version did; it does not touch, and does not
+    constrain, whatever AM change `step_fn` itself produces afterward.
+    """
+    solver = config.orbit.solver
+
+    # Specify the initial timestep size
+    dt_yr = hf_row.pop('_orbit_dt_yr', solver.dt0_yr)
```
nichollsh, line 242: Doesn't this remove a column/key from the `hf_row` variable? This should not be allowed since `hf_row` has the same keys as the whole helpfile.

src/proteus/orbit/common.py
```diff
@@ -0,0 +1,486 @@
...
+    t_total_yr = interior_o.dt
+    dt_max = solver.dt_max_yr if solver.dt_max_yr is not None else t_total_yr
+    dt_yr = min(dt_yr, t_total_yr) if t_total_yr > 0 else dt_yr
+
+    # Accumulators
+    t_elapsed = 0.0
+    n_steps = 0
+    n_rejected = 0
+
+    log.info(
```
nichollsh, line 254: Should probably be a debug message?

src/proteus/orbit/common.py
```diff
@@ -0,0 +1,486 @@
...
+        except Exception:
+            log.error(
+                '%s: C_planet update RAISED at Time=%.6e yr (C_p_old=%r, model=%s); re-raising',
+                log_label,
+                float(hf_row['Time']),
+                C_p_call_start,
+                model,
+                exc_info=True,
+            )
+            raise
```
nichollsh, line 334: Don't use bare `raise`. Also, this should update the status file before throwing an error.

src/proteus/orbit/common.py
```diff
@@ -0,0 +1,486 @@
...
+        t_elapsed,
+        t_total_yr,
+        n_steps,
+        n_rejected,
+        dt_yr,
+        t_elapsed >= t_total_yr - 1e-9,
+    )
+
+    # Persist the controller's own step-size state for the next call.
+    hf_row['_orbit_dt_yr'] = dt_yr
```
nichollsh, line 486: Does this really need to be put into the `hf_row` variable, or could it be held within `Tidal_t`?

src/proteus/orbit/hansen.py
```diff
@@ -0,0 +1,379 @@
...
+`init_k_range_table`/`init_hansen_table` explicitly once, up front, so
+the one-time FFT sweep (order a minute of wall time at the default grid)
+happens before any timing-sensitive substep loop runs. `get_all_m_hansen`
+-- the hot-path entry point, called once per ODE right-hand-side
+evaluation inside `orbit.py`'s `sp1d` and `satellite.py`'s
+`ps1d`/`ps1d_evec` -- ALSO lazily builds `_hansen_table` on first use if
+that explicit setup step was skipped (e.g. in a direct or test call), so
+nothing strictly depends on the init call having happened; it is purely a
+warm-up that keeps the one-time cost out of the timed substep loop.
+"""
```
nichollsh, line 69: This docstring is huge and barely comprehensible. Could you make it shorter, clearer, and more concise?

src/proteus/orbit/hansen.py
```diff
@@ -0,0 +1,379 @@
...
+    """
+    M = np.array(M, dtype=float)
+    E = np.copy(M)
+
+    # Danby-style improved initial guess
+    if e > 0:
+        E = M + (e * np.sin(M)) / (1 - np.sin(M + e) + np.sin(M))
+
+    # Newton iterations
+    for _ in range(10):
```
nichollsh, line 124: What happens if >10 iterations are required? Why not use a scipy `root_scalar` call?

src/proteus/orbit/hansen.py
```diff
@@ -0,0 +1,379 @@
...
+    """Tabulated eccentricity-appropriate [kmin, kmax] window. Built once by
+    init_k_range_table(); never rebuilt except via force=True."""
+
+    e_grid: NDArray[np.floating]
+    kmin: NDArray[np.integer]
+    kmax: NDArray[np.integer]
+
+
+_hansen_table: Optional[_HansenTable] = None
+_k_range_table: Optional[_KRangeTable] = None
```
nichollsh, line 222: Please avoid globals where possible. It's fine if they are constants which are never changed (e.g. physical constants). Are these expected to be updated?

src/proteus/orbit/hansen.py
```diff
@@ -0,0 +1,379 @@
...
+    kmins = np.empty(len(e_grid), dtype=int)
+    kmaxs = np.empty(len(e_grid), dtype=int)
+    for i, e in enumerate(e_grid):
+        kmins[i], kmaxs[i] = _select_k_range(e)
+    _k_range_table = _KRangeTable(e_grid=e_grid, kmin=kmins, kmax=kmaxs)
+    log.info(
+        f'k-range table built: {len(e_grid)} grid points'
+        f'(e in [{e_grid.min():.3f}, {e_grid.max():.3f}], '
+        f'n_modes in [{(kmaxs - kmins + 1).min()}, {(kmaxs - kmins + 1).max()}])'
+    )
```
nichollsh, line 278: ... seems like this is updating the globals above. Instead, perhaps store them in `Tides_t`?

src/proteus/orbit/hansen.py
```diff
@@ -0,0 +1,379 @@
...
+        Array of k values from kmin to kmax.
+    results : dict
+        m -> ndarray of Hansen coefficients X_k^{n_deg,m}(e), same shape as k_range.
+    """
+    if _hansen_table is None:
+        init_hansen_table(n_deg=n_deg)
+    table = _hansen_table
+
+    if kmin < table.kmin or kmax > table.kmax:
+        raise ValueError(
```
nichollsh, line 359: Update status file before raising errors

src/proteus/orbit/lovepy.py
```diff
@@ -33,7 +34,9 @@ def _jlsca(sca: float):
     return juliacall.convert(jl.LovePy.prec, sca)
 
 
-def run_lovepy(hf_row: dict, dirs: dict, interior_o: Interior_t, config: Config) -> float:
+def run_lovepy(
```
nichollsh, line 37: Have you run LovePy to check that it still works with these code changes? Not sure it's well covered by integration tests.

src/proteus/orbit/obliqua.py
```diff
@@ -0,0 +1,639 @@
...
+
+import json
+import logging
+import os
+from typing import TYPE_CHECKING
+
+import juliacall
+import netCDF4 as nc
+import numpy as np
+from juliacall import Main as jl
```
nichollsh, line 12: Careful here. Often, `juliacall` needs to be imported in a certain order. See the other `import juliacall` statements in the PROTEUS codebase.

src/proteus/orbit/obliqua.py
```diff
@@ -0,0 +1,639 @@
...
+
+
+def _jlsca_float(sca: float):
+    # Make a copy of a scalar, and convert to Julia type
+    return juliacall.convert(jl.Obliqua.Float64, sca)
+
+
+def _jlsca_prec(sca: float):
+    # Make a copy of a scalar, and convert to Julia type
+    return juliacall.convert(jl.Obliqua.prec, sca)
```
nichollsh, line 59: Are these duplicates of the functions in `agni.jl`?

src/proteus/orbit/obliqua.py
```diff
@@ -0,0 +1,639 @@
...
+
+Obliqua_LOGFILE_NAME = 'obliqua_recent.log'
+
+
+def import_obliqua():
+    log.debug('Import Obliqua...')
+    jl.seval('using Obliqua')
+
+
+def to_julia_dict(obj):
```
nichollsh, line 33: Perhaps a file shared by the AGNI/Obliqua/LovePy wrappers for having the Julia-related utilities would be useful, and help break up this file.

src/proteus/orbit/obliqua.py
```diff
@@ -0,0 +1,639 @@
...
+                'solid': {
+                    'ncalc': config.orbit.obliqua.solid.ncalc,
+                    'dr_min': config.orbit.obliqua.solid.dr_min,
+                    'dr_max': config.orbit.obliqua.solid.dr_max,
+                    'core': config.orbit.obliqua.solid.core,
+                    'core_props': config.orbit.obliqua.solid.core_props,
+                    'inertial_terms': config.orbit.obliqua.solid.inertial_terms,
+                    'bulk_l': config.orbit.obliqua.solid.bulk_l,
+                    'porosity_thresh': config.orbit.obliqua.solid.porosity_thresh,
+                    'dbulk_power': config.orbit.obliqua.solid.dbulk_power,
```
nichollsh, line 194: You can make this code a lot more concise by writing it dynamically. It would also make it easier to maintain. Rather than manually setting every dictionary entry from the config object attributes, you can recursively loop through them to construct a nested dictionary. I am sure Claude can help do this, but you'd expect a recursive call to a function which looks roughly like:

```
v = config.obliqua.orbit.solid
'solid' : {getattr(v,k) for k in dir(v)}
```

src/proteus/orbit/obliqua.py
```diff
@@ -0,0 +1,639 @@
...
+            PROTEUS config object
+    """
+
+    log.info('Running Obliqua for lookup table generation...')
+
+    # Read interior arrays from json file
+    file_path = config.orbit.satellite.love_number_sat
+
+    if not file_path:
+        raise ValueError(
```
nichollsh, line 305: Update status file

src/proteus/orbit/obliqua.py
```diff
@@ -0,0 +1,639 @@
...
+    # Create configuration dictionary for Obliqua
+    cfg = {
+        'title': 'Lookup_table',
+        'params': {
+            'out': {
+                'path': dirs['output/data'],
+                'time': round(0.0),  # store the Obliqua output file at t=0 to avoid
+            },  # overwriting the main PROTEUS+Obliqua output file
+        },
+        'orbit': {
```
nichollsh, line 340: See comment above about generalising this type of code.

src/proteus/orbit/obliqua.py
```diff
@@ -0,0 +1,639 @@
...
+        core_density = density_full[0]
+        rho = _jlarr(density_full[1:])
+
+        radius = _jlarr(np.array(d['radius'], copy=True, dtype=float))
+        visc = _jlarr(np.array(d['visc'], copy=True, dtype=float))
+        shear = _jlarr(np.array(d['shear'], copy=True, dtype=float))
+        bulk = _jlarr(np.array(d['bulk'], copy=True, dtype=float))
+        phi = _jlarr(np.array(d['phi'], copy=True, dtype=float))
+
+    # Create configuration dictionary for Obliqua
```
nichollsh, line 331: Why are we doing this again? Why two configs?

src/proteus/orbit/obliqua.py
```diff
@@ -0,0 +1,639 @@
...
+
+        # check the file extension to determine how to read the lookup data or generate it if it doesn't exist
+        if file_path.endswith('.nc'):
+            # Read the netcdf file
+            pass
+        elif file_path.endswith('.json'):
+            # Generate the lookup data
+            lookup_from_interior(dirs, config)
+            # Update the file path to point to the generated netcdf file
+            file_path = os.path.join(dirs['output/data'], 'moon_tides.nc')
```
nichollsh, line 517: What is this file `moon_tides.nc` and what generates it?

src/proteus/orbit/obliqua.py
```diff
@@ -0,0 +1,639 @@
...
+
+    # Interpolate Love numbers degree-by-degree
+    LNk_s = np.zeros(len(nmk_p), dtype=complex)
+
+    for n in np.unique(nmk_p[:, 0]):
+        # Lookup entries for this degree
+        lookup_mask = nmk_lookup[:, 0] == n
+
+        if not np.any(lookup_mask):
+            raise ValueError(f'Lookup table does not contain degree n = {int(n)}.')
```
nichollsh, line 536: Update status file

src/proteus/orbit/obliqua.py
```diff
@@ -0,0 +1,639 @@
...
+
+    for key in ds.variables.keys():
+        out[key] = ds.variables[key][:]
+
+    ds.close()
+    return out
+
+
+def read_ncdfs(output_dir: str, times: list):
+    return [read_ncdf(os.path.join(output_dir, 'data', '%d_obliqua.nc' % t)) for t in times]
```
nichollsh, line 596: Output data files should use a '%.0f' string formatter rather than '%d' throughout the code.

src/proteus/orbit/obliqua.py
```diff
@@ -0,0 +1,639 @@
...
+            if i == 0 and '[' in line:
+                line = '[' + line.split('[', 1)[1]
+            # copy the line
+            outfile.write(line)
+
+    # Remove logfile content
+    with open(obliqua_logpath, 'w') as hdl:
+        hdl.write('')
+
+    return inlines
```
nichollsh, line 639: Again, this seems to largely duplicate a similar function in `agni.py`. Rather than duplicate it, please create a common file which both wrappers can make use of.

src/proteus/orbit/orbit.py
```diff
@@ -4,66 +4,200 @@
...
-    """
-    ODE describing evolution of semimajor axis based on Eq. 15 of
-    Driscoll and Barnes (2015), Astrobiology 15, 739.
-    """
-    return 2 * a * e * de_dt(a, e, params)
+def evolve_orbit_star(hf_row: dict, config: Config, tides_o: Tides_t, interior_o: Interior_t):
+    """Evolve the planet's orbital parameters by interior_o.dt of physical time.
 
+    Dispatches to the requested star-planet model (sp0d, sp1d).
+    ``sp1d`` goes through the shared adaptive-substep controller in
```
nichollsh, line 48: So what is `sp0d`? It's still quite unclear to me what these models/integrators are, and how/why they differ.

src/proteus/orbit/orbit.py
```diff
@@ -4,66 +4,200 @@
...
+    elif model == 'sp1d':
+
+        def step_fn(hf_row, dt_yr, t_elapsed_yr):
+            sp1d(hf_row, tides_o, dt_yr, config)
+            return None
+
+        needs_c_planet = True
+
+    else:
+        raise ValueError(f'unrecognised star_planet_model: {model!r}')
```
nichollsh, line 83: Update status file

src/proteus/orbit/satellite.py
```diff
@@ -1,128 +1,562 @@
...
-    Implements Korenaga (2023) Icarus 400, 115564, Eq. 60:
+    Checks the satellite hasn't spiralled below/through the planet's
+    surface (a <= 1.05 R_earth), that eccentricity is in the physically
+    sane, sub-parabolic range [0, 0.999), and that both spin periods are
+    finite. Used by `evolve_orbit_satellite`'s accept/reject controller;
+    a False return triggers a state rollback and a smaller retry `dt_yr`.
+    """
+    a = hf_row.get('semimajorax_sat', np.nan)
+    e = hf_row.get('eccentricity_sat', 0.0)
+    if not np.isfinite(a) or a <= 1.05 * R_earth:
```
nichollsh, line 33: Do not hardcode this to `1.05 * R_earth`. What happens if the planet has a radius of `3 * R_earth`, for example? Instead, compare against planet radius in the `hf_row`.

src/proteus/orbit/satellite.py
```diff
@@ -1,128 +1,562 @@
...
+            fine_entry['phi'][keep],
+            fine_entry['da_planet_tide_cum'][keep],
+            fine_entry['da_sat_tide_cum'][keep],
+            fine_entry['de_planet_tide_cum'][keep],
+            fine_entry['de_sat_tide_cum'][keep],
+            fine_entry['filter'][keep],
+        )
+    )
+
+    fine_file_path = os.path.join(data_dir, 'fine_evection_data.csv')
```
nichollsh, line 195: What is this file for? Do we really need to save a file here, or could it be passed in memory? If it's for debugging, maybe only have it write when `verbosity==2`.

src/proteus/orbit/satellite.py
```diff
@@ -1,128 +1,562 @@
...
+
+    CAVEAT (not yet verified): this correction assumes the ΔC_planet
+    between calls comes purely from mass-conserving contraction /
+    internal differentiation. If some of it instead comes from a genuine
+    mass change (e.g. hf_row['M_planet'] including an atmosphere that is
+    being lost to escape between calls), that portion legitimately
+    carries angular momentum away from the system and should NOT be
+    absorbed into this rescale -- it needs to be treated as an explicit
+    sink instead. This has not been checked against how M_planet is
+    actually populated elsewhere in PROTEUS.
```
nichollsh, line 300: An example of 'chatty' LLMs here. Please make this more concise and understandable.

src/proteus/orbit/satellite.py
```diff
@@ -1,128 +1,562 @@
...
+                hf_row,
+                resonance_state,
+                margin_enter=solver.resonance_margin_enter,
+                margin_exit=solver.resonance_margin_exit,
+            )
+
+            # Log only actual band transitions (not every substep), gives
+            # a clean timeline of resonance capture/escape.
+            if last_in_band[0] is not None and in_band != last_in_band[0]:
+                log.info(
```
nichollsh, line 362: Should be debug call?

src/proteus/orbit/satellite.py
```diff
@@ -1,128 +1,562 @@
...
+                _flush_fine_evection_csv(
+                    hf_row,
+                    dirs['output/data'],
+                    fine_sink[0],
+                    in_band=bool(in_band),
+                    storage_target_interval_yr=storage_target_interval_yr,
+                )
+
+    else:
+        raise ValueError(f'unrecognised planet_satellite_model: {model!r}')
```
nichollsh, line 423: Update status file

src/proteus/orbit/satellite.py
```diff
@@ -1,128 +1,562 @@
...
+    hf_row['_orbit_resonance_state'] = resonance_state
+
+    # Diagnostic/coupling column: whether ps1d_evec judged itself inside
+    # the evection band at the end of this call. Stays 0.0 (its
+    # fillna(0.0) default in the helpfile schema) for ps0d/ps1d, whose
+    # resonance_state is never touched by _in_evection_band. Read by
+    # proteus.interior_energetics.timestep.next_step to cap dt while in
+    # the band (params.dt.evection_maximum), one PROTEUS iteration
+    # later -- see that function's docstring entry.
+    hf_row['in_evection_band'] = 1.0 if resonance_state.get('active', False) else 0.0
```
nichollsh, line 497: I don't like storing these 1/0 values in the `hf_row`. Would it be possible to keep them in `tides_t` instead?

src/proteus/orbit/satellite.py
```diff
@@ -132,69 +566,816 @@ def update_satellite(hf_row: dict, config: Config, dt: float):
...
+        K_p0 = -LNk_p_m0.imag
+        K_p2 = -LNk_p_m2.imag
+        K_s0 = -LNk_s_m0.imag
+        K_s2 = -LNk_s_m2.imag
+
+        dE_orb_p = I_p * n_mm * np.sum(s_arr * (K_p0 * X0_sq + 3.0 * K_p2 * X2_sq)) / 4
+        dE_orb_s = I_s * n_mm * np.sum(s_arr * (K_s0 * X0_sq + 3.0 * K_s2 * X2_sq)) / 4
+
+        dE_rot_p = -I_p * 3 * Omega_p * np.sum(K_p2 * X2_sq) / 2
+        dE_rot_s = -I_s * 3 * Omega_s * np.sum(K_s2 * X2_sq) / 2
```
nichollsh, line 854: Where do these equations come from and what model is it implementing? The function docstring is really unclear.

src/proteus/orbit/satellite.py
```diff
@@ -132,69 +566,816 @@ def update_satellite(hf_row: dict, config: Config, dt: float):
...
+        )
+        de_s = np.sum(
+            K_s0 * X0_sq * s_arr * sqrt_e - 3.0 * K_s2 * X2_sq * (2.0 - s_arr * sqrt_e)
+        )
+
+        sqrt_term = np.sqrt(1.0 - e_safe**2)
+        da_dt_p = a * (E_p / 2.0) * da_p
+        da_dt_s = a * (E_s / 2.0) * da_s
+        de_dt_p = (E_p * sqrt_term / (4.0 * e_safe)) * de_p
+        de_dt_s = (E_s * sqrt_term / (4.0 * e_safe)) * de_s
```
nichollsh, line 903: Need sources for these equations, too.

src/proteus/orbit/wrapper.py
```diff
@@ -223,6 +251,113 @@ def run_orbit(hf_row: dict, config: Config, dirs: dict, interior_o: Interior_t):
...
+
+            init_k_range_table()
+
+            e_grid_wide = np.concatenate(
+                [
+                    np.arange(0.0, 0.1, 0.002),
+                    np.arange(0.1, 0.9, 0.003),
+                ]
+            )
+            init_hansen_table(e_grid_wide)
```
nichollsh, line 292: I believe that these initialisation functions are being called every time this `run_orbit` function is called. Does this mean they are re-initialised at every PROTEUS step? This is especially problematic if they update global variables.

src/proteus/orbit/wrapper.py
```diff
@@ -246,27 +381,69 @@ def run_orbit(hf_row: dict, config: Config, dirs: dict, interior_o: Interior_t):
...
         from proteus.orbit.lovepy import run_lovepy
 
-        hf_row['Imk2'] = run_lovepy(hf_row, dirs, interior_o, config)
+        hf_row['Imk2'] = run_lovepy(hf_row, dirs, interior_o, tides_o, config)
+
+    elif config.orbit.module == 'obliqua':
+        from proteus.orbit.obliqua import run_obliqua, setup_logging
+
+        # setup logging for Obliqua
+        setup_logging(dirs, config.orbit.obliqua.verbosity)
```
nichollsh, line 403: This also seems to re-setup the Obliqua logging at every PROTEUS iteration?

src/proteus/orbit/wrapper.py
```diff
@@ -246,27 +381,69 @@ def run_orbit(hf_row: dict, config: Config, dirs: dict, interior_o: Interior_t):
...
+    if len(times) == 0:
+        return []
+
+    if model == 'obliqua':
+        from proteus.orbit.obliqua import read_ncdfs
+
+        return read_ncdfs(output_dir, times)
+
+    else:
+        return []
```
nichollsh, line 449: This should print a warning

timlichtenberg (changes requested):
Thanks, this is a large and careful addition and the physics reads well. A few points to resolve before merge; I have left style nits out.

The most important is a likely crash. `run_lovepy` returns early at `lovepy.py:79` and `:99` (viscosity below threshold, and fully liquid), both before the mode store at `:158`, `tides_o.add(primary='planet', perturber='star')`. So a fully molten mantle never creates that entry, and the later `tides_o.get('planet','star')` in the sp1d path raises `KeyError`, reachable in the early magma-ocean phase. **Please store a zero-heating mode set on the early-return branches, or guard the consumer.**

At `lovepy.py:150-152`, `LNk` is built with nested brackets, so its shape is `(3,1)` where a flat `(3,)` looks intended; this can break the `_dense_love` broadcast. **Please confirm the shape.**

In `common.py`, `get_C_planet` integrates density times radius^4 from the CMB to the surface, so C is the mantle only, with no core term, and it is then divided by the total planet mass. Since C feeds the spin angular momentum, this biases the spin. **Please add the core term, or document C as mantle-only and confirm the downstream use.**

The satellite semi-major axis is set in AU (`wrapper.py`) but the escape threshold in Earth radii (`terminate.py`). The comparison is metres on both sides, but the mixed units invite errors, the 0.133 AU default is beyond the Earth Hill radius, and `stop.satellite.sma_max` defaults to 0, so enabling that stop with the default ends the run on iteration 1. **Please use one unit, a physical default, and warn near the Hill radius.**

Two physics questions. `wrapper.py` zeroes `hf_row['Imk2']` when `obliqua.n != [2]`, yet `orbit.py` reads it for the sp0d `de_dt`, so that pairing would freeze eccentricity evolution; is it meant to be blocked? And `obliqua.py` passes `sigma_R_inf * sigma_R`: is `sigma_R_inf` a relative scale? The toml (1e-7) and attrs (0.5) defaults differ by orders of magnitude.

src/proteus/plot/cpl_orbit.py
```diff
@@ -181,12 +244,323 @@ def _plot_planet(i):
...
+    # Save figure
+    os.makedirs(os.path.join(output_dir, 'plots'), exist_ok=True)
+    fpath = os.path.join(output_dir, 'plots', 'plot_evection.%s' % plot_format)
+    fig.savefig(fpath, dpi=200, bbox_inches='tight')
+
+    plt.close(fig)
+    plt.ioff()
+
+
+def plot_Lovenumber(
```
nichollsh, line 408: Function names should be lowercase

src/proteus/plot/cpl_orbit.py
```diff
@@ -181,12 +244,323 @@ def _plot_planet(i):
...
+
+    vmin_real, vmax_real = np.min(all_real_log), np.max(all_real_log)
+    vmin_imag, vmax_imag = np.min(all_imag_log), np.max(all_imag_log)
+
+    # Setup Figure
+    scale = 1.0
+    fig, axs = plt.subplots(1, 2, figsize=(14 * scale, 6 * scale), sharey=True)
+
+    cmap_real = plt.get_cmap('plasma')
+    cmap_imag = plt.get_cmap('viridis')
```
nichollsh, line 473: Please use the scientific colour maps

src/proteus/utils/coupler.py
```diff
@@ -775,27 +775,41 @@ def GetHelpfileKeys():
...
+        'eccentricity_sat', # orbital eccentricity of satellite [1]
+        'ecc_dot_sat',      # eccentricity derivative [1 s-1]
         'M_sat',            # mass of satellite [kg]
-        'plan_sat_am',      # angular momentum of sat+pla [kg m2 s-1],
+        'R_sat',            # radius of satellite [m]
+        'C_sat',            # principal moment of inertia of satellite [kg m2]
+        'plan_sat_am',      # angular momentum of satellite+planet [kg m2 s-1],
+        'axial_period_sat', # day length of satellite around its axis [s]
+        'evection_angle',   # evection angle [rad]
+        'in_evection_band', # whether inside the evection resonance band [bool, 0/1]
```
nichollsh, line 804: Ideally we should not store booleans in the helpfile, even when casting them to integers (which are then rounded to floats...)

docs/Explanations/model.md
nichollsh, line 1: The explanation of the orbital/tides modelling needs substantial expansion in the documentation. There are several large docstrings (and sections of code) which are unclear in this PR. A clear explanation in the documentation is absolutely necessary.

nichollsh (changes requested):
Thanks for this PR, @MarijnJ0 - it's really exciting to see PROTEUS and Obliqua integrated together. Since these changes are wide reaching and large, I have not yet tested the coupling, and just performed a first pass (by inspection) of the code and your changes.

I have a number of concerns which need to be addressed before this is ready for another pass. There are also some good suggestions from copilot. Please also see my in-line comments below.

1. The new files are extremely log and, to some extent, duplicate the AGNI wrapper. I believe they could be made substantially shorter by (a) generalising some of the code rather than hardcoding parts of it, and (b) sharing functions with the AGNI wrapper.

2. The docstrings are often very long, or very short. In either case it makes the new functions (particularly around the orbital evolution) hard to understand. As a result, I am left with a very poor understanding of what the model is actually doing.

3. There are a few global variables. These should be avoided in general, and should certainly not be updated at runtime, if they are unavoidable.

4. The documentation is lacking. You can probably move some of the docstrings and config comments there.

5. Please can you provide an example config appropriate for testing this integration?

6. Have you confirmed that this works with Aragog, SPIDER, the BLT  interior module, and the dummy one?

7. There are some cases which are hard-coded to an Earth scenario.

Thanks again for this, @MarijnJ0. It's a huge task to implement this so it will naturally require a few rounds of PR review. Happy to look at this again when it's ready, and then test it out!

MarijnJ0:
Thank you both for the feedback and comments.

I am in the process of implementing the fixes. To answer some of the questions:

> Two physics questions. wrapper.py zeroes hf_row['Imk2'] when obliqua.n != [2], yet orbit.py reads it for the sp0d de_dt, so that pairing would freeze eccentricity evolution; is it meant to be blocked? 

Yes, since Obliqua can output for arbitrary tidal degree `n` I added this block. The `Imk2` value passed to the orbital model is per definition degree 2, so mixing this with other degrees would be wrong. Since Obliqua generally only outputs n=2, we can take the mean (already a stretch) and use it here, but if the user requested n > 2, than that completely breaks energy conservation between the tides and orbit, hence it is not permitted here. I will add a check that catches this at the config level. For this exact reason there is now the `sp1d`, which is the only correct star-planet orbital model to be used with Obliqua at the moment. 

We can add the option to Obliqua to only return the [2, 0, 1], [2, 2, 1], [2, 2, 3] nmk modes, which under the small eccentricity approximation and spin-orbit synchronization reduce to the terms in Lovepy. This is the only correct way to end up with a single dominant Love number.

> And obliqua.py passes sigma_R_inf * sigma_R: is sigma_R_inf a relative scale? The toml (1e-7) and attrs (0.5) defaults differ by orders of magnitude.

This is a typo, since Obliqua uses sigma_R for the friction channel and sigma_R_inf for the drag + shear/bulk/darcy channels I had to introduce the factor in PROTEUS. Using the factor instead allows us to run grids over sigma_R, without the need to update two params simultaneously, which cannot be done in grids I think.

input/all_options.toml
MarijnJ0: I updated the docs and description. Store_3d enable/disables the internal angular integration, so that the output files only contain 1D data, this reduces file size. Enforce_ec is a hardcoded energy conservation check, in principle it doesn't do anything, since Obliqua self-consistently computes Love numbers and dissipation. In limiting cases where the model is unstable (e.g. very small forcing frequencies < 1e-14 Hz) the heating is negligible anyways, and this just cleans up any noise in those cases. In the high frequency regime (> 1e-3 Hz) the same applies, but here heating rates are more substantial. Regardless, the Lovenumbers remain unaffected, only the heating is scaled to match it.

input/all_options.toml
MarijnJ0: The tidal response comprises of an infinite sum over tidal modes. Here we select the degree(s) `n` and the order(s) `m` of that sum that we want to include. Obliqua can handle all of them except for obliquity tides associated with odd `m` at the moment. We could make it implicit that if n = 3, the code uses n = 2 and 3, but this way it is more explicit. Regardless, we need to keep `m` as an array given `m = 1` is missing in Obliqua.

input/all_options.toml
MarijnJ0: At the moment they are arbitrary. They are used by the `inertial`, `inertial-liquid`, and `solid` core solution vectors. At the moment PROTEUS assumes an incompressible fluid iron core (I think) which is only compatible with the `liquid` solution. But given these config parameters, we can also use the other solution vectors. I can change the defaults to the frictionless incompressible limit, what do you think?

src/proteus/orbit/common.py
MarijnJ0: The Tides_t class stores all the tides information between multiple primary and perturber combinations, so you can store planet-satellite, satellite-planet, planet-star, etc, in a convenient way that is scalable. Additionally, this allows us to store lookup tables for the objects e.g. through satellite_dict-planet, keeping that fixed, and updating satellite-planet at each timestep.

src/proteus/orbit/hansen.py
MarijnJ0: This would be too slow, we need this to be vectorized. We could use `scipy.optimize.newton`, however the current implementation is adequate for e~0.9, so that should never be a problem. Moreover, the Obliqua implementation uses the same structure, so for consistency I would prefer this.

src/proteus/orbit/hansen.py
MarijnJ0: I understand, however these tables only contain constants, they never change, so this should not be an issue.

src/proteus/orbit/hansen.py
MarijnJ0: The tables are build once at the start of the simulation, or when they are lost during a --resume. Additionally, if the system ends up outside of the precomputed range, then the table is automatically expanded to cover the additional eccentricities. No entries are ever changed.

src/proteus/orbit/obliqua.py
MarijnJ0: Moved juliacall imports to top.

src/proteus/orbit/obliqua.py
MarijnJ0: They where copied/adapted from lovepy.py. I have now added the shared functions under `utils/julia_common.py`, while still allowing each module to use its own `prec`. I checked, but do not see how this can easily be shared with AGNI too, eitherway, AGNI can access these functions if needed.

src/proteus/orbit/obliqua.py
MarijnJ0: I have now added the shared functions under `utils/julia_common.py`

src/proteus/orbit/obliqua.py
MarijnJ0: `Lookup_from_interior` creates a simplified tidal setup that generates a basic lookup table for the satellite object. If the user passes the interior data, then PROTEUS can keep the Rayleigh drag consistent between the planet and satellite, by generating the lookup tables at the start of the run. If the user passes the lookup table directly, and still varies Rayleigh drag for the planet, then the Rayleigh drag for the planet and for the satellite may not match. Hence, for grid runs it is more convenient to pass the interior data, and let PROTEUS generate the lookup data for each similation independently based on the config associated with that particular run. Thus we need to configs, one for the planet and one for the satellite.

src/proteus/orbit/obliqua.py
MarijnJ0: It contains the tidal response of the satellite (renamed it to `sat_tides.nc`) and can be directly imported into `Tides_t` using `add_from_file`. It is generated from the provided interior json file by `lookup_from_interior`. In case the user provides the `*.nc` file than the generation step is skipped, and the data is directly imported.

src/proteus/orbit/satellite.py
MarijnJ0: It is purely for storing the data with high time resolution so that we can make plots that display evection resonance capture. Potentially we can make the writting periodic, not on every PROTEUS step. I will check this

input/all_options.toml
nichollsh: I would suggest looking around in the literature for reasonable values - or at least order of magnitude estimates. A quick search on Google Scholar turned this up, which might be useful? https://journals.aps.org/prb/abstract/10.1103/PhysRevB.53.14063

src/proteus/orbit/hansen.py
nichollsh: Fine with me. Thanks.

src/proteus/orbit/obliqua.py
nichollsh: Makes sense now, thanks. Should this happen only at the start of the simulation, then?

src/proteus/orbit/satellite.py
MarijnJ0: I am working on this atm, I am adding a timestep.py file in the orbit folder which contains the relevant functions to control dt. Then passing only dt to the helper file (units yr). This way we dont need to pass tides_t through the interior functions, and allows us to store the booleans in tides_t, which is then only requested in the functions in the orbit folder.

src/proteus/orbit/obliqua.py
MarijnJ0: Yes exactly, currently the code tries to access the lookup entry in tides_o, and if the key doesn't exist (i.e. at startup or after resume) then it tries to build the lookup table.

src/proteus/orbit/wrapper.py
MarijnJ0: As long as PROTEUS is on t<=1yr these get called, after that the `else` branch becomes active, which is where the orbital integrators are located.

Note that these functions check for existing tables, they dont recompute the table. So if PROTEUS takes multiple cycles to initiate it has no effect on these globals. 
E.g.
```
global _k_range_table
if _k_range_table is not None and not force:
    return
```
and
```
global _hansen_table
if _hansen_table is not None and not force:
    return
```

MarijnJ0: If I recall correctly, these config-to-hf_row initiations where originially located in each of the orbit model functions. With this change, everything gets initiated in the wrapper, and all the orbit models simply read the current state from the help file. I think this is much cleaner than scattering this across the different integrators.

nichollsh:
Thanks for the update, @MarijnJ0. Please let us know when this is ready for another review. I note that there are some conflicts with the main branch which need resolving too.

MarijnJ0:
I expect to have it ready for another round of review tomorrow 👍

nichollsh:
I have now had some time to have another look at this branch. I noticed a few more things which need to be addressed, that were not picked up during my first review pass. Please see the comments below, @MarijnJ0.

### Obtaining Obliqua

The `get_obliqua.sh` script works and the tests pass, but it is a bit fragile. It should additionally perform the following actions.

* If the Obliqua folder already exists, remove the `Manifest.toml` file inside of it.
* Call `Pkg.resolve()` before calling `Pkg.instantiate()`.
* Throw an error if the `if` statement on line 36 returns false

When running with PROTEUS, I then also get:
```
juliacall.JuliaError: ArgumentError: Package Obliqua not found in current path.
- Run `import Pkg; Pkg.add("Obliqua")` to install the Obliqua package.
```


### Tutorial and docs

The new documentation page is fantastic and really useful for understanding which module to use. Thanks for adding this.

This PR would strongly benefit from a tutorial or example config which runs PROTEUS and Obliqua together. I could not find one, so constructed one from the `all_options.toml` keys which you added. Please can you try the config I have made below and include it in this branch before we merge this PR, please? Maybe you want to suggest some changes on it too.

[tutorial_earth_moon.toml.txt](https://github.com/user-attachments/files/32159131/tutorial_earth_moon.toml.txt)

However, I get the following error:
```
  File "/home/harrison/Projects/PROTEUS/src/proteus/orbit/wrapper.py", line 509, in run_orbit
    log.info('    Pla Im(k)  = %.1e ' % Imk)
             ~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~
TypeError: only 0-dimensional arrays can be converted to Python scalars
```

tests/config/test_config_schema_invariants.py
```diff
@@ -706,108 +708,274 @@ def test_boundary_requires_fixed_surface_state_passes_with_fixed():
...
+    orbit.obliqua.n at all -- must not raise regardless of its value,
+    confirming the check is gated on module=='obliqua' first."""
+    instance = _make_config_instance(
+        **{
+            'orbit.module': 'lovepy',
+            'orbit.star_planet_model': 'sp0d',
+        }
+    )
+    instance.orbit.obliqua = SimpleNamespace(n=[2, 3])  # would fail if module were obliqua
+    result = sp0d_obliqua_degree_mismatch(instance, None, None)
```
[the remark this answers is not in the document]
nichollsh: @MarijnJ0 please incorporate the suggestion by copilot above.

MarijnJ0:
Thanks Harrison! I am currently working on these final tweaks, after that I will do the full integration test across all orbit/tides models. I will update you here when everything passes.

nichollsh:
Okay, sure - thanks!

MarijnJ0:
@nichollsh @timlichtenberg 

I have added the final tweaks/bug fixes and verified that the tidal and orbital models work as expected. I have tested them with SPIDER, (found fix, currently testing Aragog), but not with the boundary module. <del> I have an issue with my jax installation when running Aragog locally and have not had time to fix this yet. If possible, could one of you try running the tutorial with Aragog? I suspect it will work fine, since nothing substantial changed between the interior-to-tides implementation w.r.t. the original Lovepy implementation. </del> When running the boundary model I encountered a `division by zero` error, internal to the boundary module before tides or the orbit are called, so this is likely an issue that is unrelated to the orbit/tides code.

>   File "/home/marijn/PROTEUS/src/proteus/interior_energetics/boundary.py", line 544, in drs_dTp
    dr_s_dT_p = -(volume_diff) / (3 * T_range * r_s**2)
                ~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~
ZeroDivisionError: float division by zero

That said, when testing the orbit models a few things motivated the following major changes:
- `ps0d` requires an additional check that limits the relative growth da/a ~0.01, anything beyond that breaks the convergence, so there is now a check specifically for this.
- `ps0d` requires `imk2`, to allow for backwards comaptibility I added a new `legacy` mode to Obliqua, which makes it behave like Lovepy and output a single `imk2` value. 
- When encountering a seismic resonance (`inertial_terms = true`) the love numbers would rapidly inflate on timescales not resolved by the PROTEUS step. To prevent this from inflating the orbit and interior heating, there is now a `cap_LN` option in obliqua that limits the Love numbers. Alternatively, you can shrink `mushy_maximum` as is described in the docs.

A note on the Lovepy implementation: currently the orbit models assume a certain symmetry holds when expanding the modes in `tides_o`, this means that the heating profile returned by `Lovepy` and `Obliqua` needs to already account for this. Obliqua does this, but Lovepy does not. The solution would be to add a single additional mode in Lovepy, it likely won't change much. In addition to this, there are several bugs in Lovepy that would cause a notably larger drift in the output. Potentially, it is worth fixing all of this at some point, all corrections are already present in Obliqua, so we would just copy over a few lines of code.

Finally, there is now also a tutorial with example output, it uses the sketch config file from Harrison. I did make some tweaks to it, like enabling tidal heating in the interior ;) . 

Given this, we should consider merging the Obliqua PR as well, I kept it small so it should not take to long to get through. (https://github.com/FormingWorlds/Obliqua/pull/32)

With that, I think this is ready for another round of review!

**Verified working with Aragog as well now.

nichollsh (changes requested):
Thanks for addressing some of my previous comments/suggestions, @MarijnJ0. This PR is now beginning to take good shape, however, some more changes are required.

Importantly, I am not able to check of a lot of your changes because you did not provide the PROTEUS config file described in the new tutorial. Will be happy to re-review this and check the integration once this file is provided.

Do you want to merge the Obliqua `md/docs` branch first?

## Install script and version check

You are still missing the `Pkg.add()` call in the `get_obliqua.sh` script.

On a related note, Obliqua should be added to the `validate_module_versions()` fully - its installed version printed when PROTEUS starts.


## Tutorial

The new tutorial page is useful but incomplete and contains errors. 

Why does it describe making a JSON config file for Obliqua? It says to "Create a new file called 0d_test_moon.json in the examples/satellite/" but this should not be necessary since PROTEUS does step during the simulation.

The actual suggested PROTEUS config file does not exist. Perhaps you forgot to add it? `input/tutorials/tutorial_earth_moon.toml`. Without this file I am unable to test the simulations.

## Error about Imk2

See comment above in this thread. I don't see where in the code you have corrected this error where `Imk2` is treated as a float, when it can be an iterable.

## Config options

Thanks for addressing many of my concerns about the config. However, you've still got `evection_maximum = 0.0` as the default timestep ceiling. I'd suggest making this `'none'` to disable and instead adding a validator that it must be `>0` or `none`. Similarly for `evection_growth_factor` and `evection_cooldown_iters`.

MarijnJ0:
Thanks @nichollsh, for pointing this out!

Ideally we merge the Obliqua PR first so I can pin to that version in PROTEUS. 

---

I added the following to `get_obliqua.sh:

>echo "Registering Obliqua into the default Julia environment..."
LD_LIBRARY_PATH="" julia -e 'using Pkg; Pkg.add(path=".")'

It should now be added to the main julia environment.

---

I forgot to add the tutorial config, it should now be there.

The interior properties of the satellite are user inputs, we do not compute these at the moment, hence the user does have to provide the `JSON` file or the lookup tables directly as `NetCDF`. The file you are refering to is the planet interior, but this only concerns the satellite, not the planet.

---

The Love number was being computed from `omega` which was passed through `_jlsca_prec`. I added the conversion float(omega) before np.sign() to resolve this issue, and used `storage.LNk` instead of the raw `LNk` for the same reason.

input/tutorials/tutorial_earth_moon.toml
```diff
@@ -0,0 +1,229 @@
...
+    [orbit.satellite]
+        include_satellite     = true         # include satellite
+        mass_sat              = 0.012        # satellite mass [M_earth]
+        radius_sat            = 0.273        # satellite radius [R_earth]
+        axial_period_sat      = "none"       # day length [hours]; none = tidally locked
+        semimajoraxis_sat     = 3.5          # satellite orbit semi-major axis [R_earth]
+        eccentricity_sat      = 0.05         # satellite initial orbital eccentricity [dimensionless]
+        evection_angle        = 0.0          # initial evection angle [degrees]
+        c_factor_sat          = 0.4          # satellite normalized moment of inertia factor [dimensionless]
+        love_number_sat       = "examples/satellite/0d_test_moon.json"
```
nichollsh: This file is not provided in the PROTEUS repo. I understand that it's described in the tutorial docs page, but I would suggest packaging it here directly as `input/tutorials/tutorial_earth_moon.json`.

docs/Tutorials/orbit_tides.md
```diff
@@ -0,0 +1,173 @@
...
+
+Reference data is also fetched automatically when `proteus start` runs
+without the `--offline` flag, so the download commands above are only
+required for offline use.
+
+## Physical setup
+
+The physical setup for the orbit and tides tutorial involves defining the 
+properties of the planet and its satellite. Create a new file called 
+`0d_test_moon.json` in the `examples/satellite/` directory relative to the
```
nichollsh: See comment about moving this to `tutorials/` subdir

docs/Tutorials/orbit_tides.md
```diff
@@ -0,0 +1,173 @@
...
+parameters for this test case, we still need to provide them in the data file. The 
+values provided here are arbitrary and do not affect the tidal response calculations.
+
+## Running the simulation
+
+```bash
+conda activate proteus
+mkdir -p output/tutorial_earth_moon
+nohup proteus start -c input/tutorials/tutorial_earth_moon.toml \
+    > /tmp/proteus_earth_moon_launch.log 2>&1 & disown
```
nichollsh: I'd suggest simplifying this to just: `proteus start -c input/tutorials/tutorial_earth_moon.toml`

docs/Tutorials/orbit_tides.md
```diff
@@ -0,0 +1,173 @@
...
+
+Add `--offline` to skip the reference-data check on later runs; the first
+run must be able to download any missing data (or download it beforehand,
+see the prerequisites above).
+
+Monitor progress with `tail -f output/tutorial_earth_moon/proteus_00.log`
+(the log appears once PROTEUS has initialized).
+
+!!! warning "Runtime"
+    This run takes several hours to overnight depending on hardware.
```
nichollsh: Would it be possible to choose a tutorial configuration which runs much faster? A tutorial that runs overnight is not so pedagogical. I think it's particularly slow because PROTEUS is falling back to Zalmoxis melting curves; the updated config above should help with this, although it's related to the bug here https://github.com/FormingWorlds/PROTEUS/issues/883

src/proteus/utils/coupler.py
```diff
@@ -1919,8 +1963,16 @@ def UpdatePlots(hf_all: pd.DataFrame, dirs: dict, config: Config, end=False, num
     plot_escape(hf_all, output_dir, plot_format=config.params.out.plot_fmt)
 
     # Planet and satellite orbit parameters
-    if config.orbit.evolve or config.orbit.satellite:
-        plot_orbit(hf_all, output_dir, config.params.out.plot_fmt)
+    if (
+        config.orbit.star_planet_model is not None
+        or config.orbit.planet_satellite_model is not None
+    ):
```
nichollsh, line 1970: This should also ideally call `plot_orbit_system()`.

src/proteus/utils/coupler.py
```diff
@@ -1979,6 +2031,23 @@ def UpdatePlots(hf_all: pd.DataFrame, dirs: dict, config: Config, end=False, num
             # Energy flux profiles
             plot_fluxes_atmosphere(output_dir, config.params.out.plot_fmt)
 
+    # Lovenumber spectra for tidal dissipation
+    if obliqua:
+        # Which times do we have tides data for?
+        ncs = glob.glob(os.path.join(output_dir, 'data', '*_obliqua.nc'))
+        nc_times = [int(f.split('/')[-1].split('_obliqua')[0]) for f in ncs]
+
+        plot_times_obliqua = nc_times
```
nichollsh: Is this line necessary?

pyproject.toml
```diff
@@ -307,11 +306,19 @@ url = "https://github.com/ExoInteriors/BOREAS.git"
...
 url = "https://github.com/nichollsh/LovePy"
 ref = "main"
 
+[tool.proteus.modules.obliqua]
+# Multi-phase tides module (Julia, with juliacall bridge).
+# Cloned in-place and instantiated by tools/get_obliqua.sh, mirroring
+# the AGNI pattern. Pinned to a commit on md/phys_tests (pre-main-merge);
+# bump once the branch lands on main.
+url = "https://github.com/FormingWorlds/Obliqua.git"
+ref = "89be36ee70a4a7c776c4d9b0d461ba1259b6b0c0"
```
nichollsh:
```suggestion
ref = "71950f274060fe404c246a02a9b005d81bda5a6c"
```

nichollsh (changes requested):
Thanks, @MarijnJ0. This is now taking good shape! I ran the tutorial for a while and it made plots which seem to replicate those you've included in the documentation. See below.

I still think the performance of this tutorial config could be improved. Could you somehow increase the timestep length or make other changes? Doing so would make the tutorial more useful.

Image

Image

To get this working in reasonable time, I made some adjustements to the config: 

[init_coupler.toml.txt](https://github.com/user-attachments/files/32279378/init_coupler.toml.txt)

[proteus_00.log](https://github.com/user-attachments/files/32279380/proteus_00.log)

----

The Obliqua-related tests seem to pass fine. However, I have some  further comments requesting changes below.

input/tutorials/tutorial_earth_moon.toml
MarijnJ0: Ok that makes sense, I will add it. I think it would still be useful to have an overview of the file in the docs, similar to the Obliqua docs/tutorial, e.g. with the `// ...` comments behind the inputs, what do you think?

docs/Tutorials/orbit_tides.md
MarijnJ0: @nichollsh I am not able to download the config, can you maybe send it over discord? Additonally, I tried setting `temperature_mode = "adiabatic_from_cmb"` and this reduced the total time from 1.15 hr to 1 hr, so not much of an improvement. It seems like `SPIDER` limits the time step to allow the surface flux to change only 4K per step when tides are enabled, this causes the solver to slow down substantially. For now, I have added a comment in the tutorial, suggesting the reader to spend this time reading up on the docs and look at some pregenerated plots :) . Overall, this is still much faster than running Aragog+Zalmoxis, which takes ~7 hr.

arround line 1193 in `spider.py`:
> if config.interior_energetics.heat_tidal and (np.amax(interior_o.tides) > 1e-10):
        dT_max = 4.0
        log.info('Tidal heating active; limiting dT_magma to %.2f K' % dT_max)

input/tutorials/tutorial_earth_moon.toml
nichollsh: The file path here is now incorrect, because the Obliqua satellite json file is now located at `input/tutorials/`.

nichollsh: The `//` comments cause syntax errors for me when Python parses the file.

nichollsh (dismissed):

nichollsh:
@MarijnJ0 please can you updated the pinned AGNI version in the PROTEUS `pyproject.toml` file to `1c4baad046c52ec2205508062fbcb962bce15375`? Thanks.

nichollsh (dismissed):

src/proteus/config/_orbit.py
```diff
@@ -65,6 +257,126 @@ def ax_valid(instance, attribute, value):
         raise ValueError(f'Initial axial period must be >0 hours, got {value}')
 
 
+@define
+class Satellite:
```
timlichtenberg, line 261: Four config files still use the removed `orbit.evolve` / `orbit.mass_sat` / `orbit.semimajoraxis_sat` keys and fail to load on this branch with `UnknownConfigKeyError`: `input/tutorials/tutorial_vapourise.toml`, `input/inference/example.toml`, `tests/integration/dummy.toml` and `tests/data/integration/zalmoxis_resume_mesh/init_coupler.toml`. `dummy.toml` is loaded by `test_integration_dummy`, `test_integration_observe_wrapper` and `test_integration_petitradtrans_helpers`, so those error at `Proteus` construction; the zalmoxis test skips on the error, so that coverage silently drops. `example.toml` also declares `[orbit.satellite]` as a non-table. We need to migrate all four to the 3.0 schema before merge: drop `orbit.evolve`, move the satellite keys into `[orbit.satellite]`, and set `config_version = "3.0"`.

timlichtenberg (changes requested):
Blocking issue in the inline comment: four shipped config files no longer load on this branch. Let's get those migrated before merge, otherwise looks good.

timlichtenberg (approved):
Re-checked after the latest commit. The four config files that failed to load under the 3.0 schema now parse correctly, and both integration tests that were affected pass, including the zalmoxis test that was previously silently skipping instead of running. That resolves the blocking issue from the last review.

I also looked at the open thread on wrapper.py:296 about the Time<=1 seeding gate, since it's still unresolved and touches the same orbit code path. In the dummy integration test's actual output, semimajorax and eccentricity are seeded to nonzero values at Time=0 and evolve smoothly through the run, so the gate is not being skipped for this config. Not blocking on my end; leaving the thread for whoever raised it to close.

Approving.

src/proteus/orbit/orbit.py
MarijnJ0: Descriptions are included in the docs now.

MarijnJ0:
Ok perfect, I went through the final open threads, they seem to be resolved. Now merging this into main!!

nichollsh:
Fantastic work! Exciting to see this merged.

PR #862, opened 2026-09-09
Resolve a get_* install path that does not exist yet

egpbos:
## Description

`tools/get_socrates.sh` takes an optional install path, and resolved it with `portable_realpath`. That helper fails on a path which does not exist yet (BSD `realpath` on macOS refuses a missing leaf, GNU `realpath` a missing parent), and since `set -euo pipefail` starts further down the script, the empty result travelled on to `git clone`:

```
$ bash tools/get_socrates.sh /tmp/does-not-exist
realpath: /tmp/does-not-exist: No such file or directory
fatal: could not create work tree dir '': No such file or directory
```

`proteus get socrates` hit the same failure on macOS, because `utils/data.py:get_socrates()` returns early when `socrates/` exists and so only ever passes a path which does not.

Fix: `portable_realpath` falls back to its existing `python3` resolver whenever `realpath` refuses the path. `git clone` creates the destination itself, parents included, so no `mkdir -p` is needed. Resolution of existing paths and symlinks is unchanged.

The helper is copied into nine `get_*.sh` scripts, so all nine are updated and a test pins the copies as identical text. `get_socrates.sh` additionally errors out if resolution returns empty, and its header now documents the positional argument.

No related issue; reported directly.

## Validation of changes

macOS 26.6.2 (arm64), Python 3.12, system bash 3.2 with BSD `realpath`.

- Five new tests in `tests/tools/test_install_scripts.py`, covering missing-leaf and missing-parent resolution, the install-path block, and the empty-resolution error. All five fail against the old script and pass against the new one.
- With `git` stubbed, both `bash tools/get_socrates.sh <new path>` and `proteus.utils.data.get_socrates()` now pass the full destination to `git clone` and create it; before the change the destination was empty.
- `pytest tests/tools/`: 226 passed. Unit tier: no new failures (the 23 failing `test_aragog*.py` tests fail identically without this change).
- `ruff check` and `ruff format --check` clean; `bash -n` clean on all scripts.

Not verified: a full SOCRATES clone and build (the clone was stubbed), and GNU `realpath` on a Linux host, which the CI Linux job exercises.

## Checklist

- [x] I have updated the docs, as appropriate (usage header in the script)
- [x] I have checked that all dependencies have been updated, as required (none changed)

### DIFF
--- a/tests/tools/test_install_scripts.py
+++ b/tests/tools/test_install_scripts.py
@@ -4,7 +4,6 @@ configuration invariants they depend on.
 
 Reusable shell logic replicated inline from ``tools/get_petsc.sh`` and
 ``tools/get_spider.sh``:
-- ``portable_realpath()``: cross-platform path resolution
 - ERR trap: exit-code and step-name capture
 - Platform detection: PETSC_ARCH assignment
 - Homebrew prefix fallback: architecture-aware default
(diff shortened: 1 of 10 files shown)


nichollsh (approved):
Good with me - tested on my laptop and it works as expected. This is using Fedora 44, `realpath` GNU-coreutils v9.10, `bash` RHEL  v5.3.9.

One optional question and suggestion... Why duplicate the function across all the `get*` files? I thought we were now using a file within `tools/` that contains these shared functions and `source`-ing it as required? Would be great to do that here rather than dupliate the `portable_realpath` function in all `get*` scripts.

egpbos:
@nichollsh Thanks! The deduplication PR #864 was actually set up later than this one :) I will merge this now and then add the deduplication of this to that PR (which anyway I still had to clean up).

PR #864, opened 2026-09-09
Source the get_* install scripts from one shared shell library

egpbos:
## Description

Each `tools/get_*.sh` carried its own copy of the same shell: `portable_realpath` in 9 scripts, the dirty-checkout guard in 7, the `--force` split in 6, the GitHub SSH probe in 4, the repo-root derivation in 12. A fix to one copy left the rest untouched, which is why #862 had to patch the same four lines nine times. This is the extraction Harrison asked for in that review.

`tools/_get_common.sh` now holds one copy of each. Every script sources it through `$(dirname "${BASH_SOURCE[0]}")` and stops if it is missing, since half of them set no `-e` and would otherwise run on with `proteus_root` empty and their work path at the filesystem root. Per-script differences became parameters rather than separate copies: `guard_dirty_checkout` takes a git pathspec, so SOCRATES still excludes its regenerable `make/Mk_cmd`, and `get_parse_args` reports both the `--force` switch and the optional install path.

Intended behaviour changes:

- A destination that does not exist yet now resolves. #862 has since merged its nine copies of that fix; this branch keeps the behaviour in the library's single copy, and drops the test that pinned the nine copies as identical text in favour of two invariants: no script defines a helper privately, and every caller sources the library first.
- `get_agni.sh` and `get_spider.sh` resolve the checkout root through symlinks, as the other nine already did.
- The guard no longer pipes its probes into `head`. A checkout git cannot read (unborn HEAD, incomplete `.git`) used to stop two scripts with no message and be deleted by the other six; it is now kept with the reason named.
- The SOCRATES cache key in the CI action hashes the library too, otherwise a change to a helper restores a build tree made by the old script.

Also merged from main: `get_obliqua.sh`, which arrived with its own root derivation and pin lookup, now sources the library, and a second derivation of the checkout root is flagged by the invariant, since that copy carried no helper name for the existing checks to catch.

Scripts: 397 lines removed, 110 added. Library: 167 lines. Docs: a "Module install scripts" section in `development_standards.md`.

## Validation of changes

macOS 26.6.2 (arm64), bash 3.2.57 (the only bash on PATH, so the bash 3.2 target is what ran), Python 3.12.

- Every script ran whole against stubbed `git` and `ssh` in a throwaway checkout, with the SSH probe accepting and refusing, on this branch and on `main`. All clone destinations and transports match, except the two intended cases above. `PETSC_DIR`, `PETSC_ARCH` and the SPIDER default destination are identical.
- `tests/tools/test_install_scripts.py` sources the shipped library instead of copying shell into the tests; helper cases run under both a plain shell and `set -euo pipefail`. Two invariants replace #862's stopgap: no script carries a private copy of a helper, and every caller sources the library first. Each invariant case was checked against the mutation it targets and fails under it.
- `pytest tests/tools/`: 275 passed. `pytest -m "unit and not skip and not slow and not integration"`: 3327 passed, 47 skipped, 9 failed, the same 9 failing on `main` in this environment (`tests/atmos_clim/test_janus.py`).
- `ruff check`/`ruff format --check`, `bash tools/validate_test_structure.sh`, `bash -n` on every script: clean.

Not covered: no real clone or build (git, ssh and the compilers were stubbed); GNU `realpath` not exercised on Linux; `shellcheck` not installed here.

## Checklist

- [x] I have checked that all dependencies have been updated, as required (none changed)

### DIFF
--- a/.github/actions/setup-proteus/action.yml
+++ b/.github/actions/setup-proteus/action.yml
@@ -353,7 +353,9 @@ runs:
       uses: actions/cache@v5
       with:
         path: socrates/
-        key: socrates-${{ runner.os }}-${{ runner.arch }}-portable-${{ steps.runner-image.outputs.version }}-${{ steps.pins.outputs.socrates_ref }}-${{ hashFiles('tools/get_socrates.sh') }}
+        # Both files: get_socrates.sh sources the shared helpers in
+        # tools/_get_common.sh, so a change there also changes what is built.
+        key: socrates-${{ runner.os }}-${{ runner.arch }}-portable-${{ steps.runner-image.outputs.version }}-${{ steps.pins.outputs.socrates_ref }}-${{ hashFiles('tools/get_socrates.sh', 'tools/_get_common.sh') }}
 
     - name: Build SOCRATES (cache miss)
       if: steps.cache-socrates.outputs.cache-hit != 'true'
--- a/docs/How-to/development_standards.md
+++ b/docs/How-to/development_standards.md
@@ -62,6 +62,32 @@ field sets) are written one entry per line with a trailing comma, grouped by
 module under a header comment, and ordered alphabetically within each group.
 This keeps two independent additions on different lines so they merge cleanly.
 
+**Module install scripts**
+
+The `tools/get_*.sh` scripts share one sourced library, `tools/_get_common.sh`.
+It provides `portable_realpath`, the checkout root (`proteus_root`) and tools
+directory (`proteus_tools_dir`), the `--force` and install-path argument split
+(`get_parse_args`), the dirty-checkout guard (`guard_dirty_checkout`), the
+GitHub SSH probe (`github_use_ssh`), the https-to-SSH URL rewrite
+(`github_ssh_url`), and the pin reader (`resolve_module_pin`).
+
+A new install script starts with the same bootstrap the existing ones use:
+
+```bash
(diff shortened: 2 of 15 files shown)


nichollsh (dismissed):
This is great and makes the code much easier to maintain. Thanks, @egpbos, for cleaning these up. I have tested this on my laptop (Fedora 44) and it seems to work well. That is, the scripts download/compile the respective modules as before.

Looks good to merge into main.

Do we want to perform a similar overhaul to the `get_*` scripts in the submodules?

egpbos:
Sorry, I actually wasn't happy with the code myself yet, Claude was a bit too trigger happy (I've instructed it to open PRs in draft mode from now on). Just doing a few more tweaks before reopening for review.

egpbos:
@nichollsh ready for rereview! I merged in main so that the path handling fixes from #862 are now also deduplicated. Also updated to include Oblique getter.

nichollsh (approved):
Thanks for this too, @egpbos. I appreciate you incorporating my suggestions regarding the shared `_get_common.sh` library. I tried this on my laptop too for a few of the `get*` scripts (not all) and it also works we described for these. Happy for this to be merged once updated from `main`.

timlichtenberg (approved):
Looks good, thanks! One small thing: the guard now keeps a checkout that git cannot read (for example after a killed clone) instead of re-cloning it, and install.sh and `proteus install-all` cannot pass `--force`. A line in the install docs on running the script with `--force` by hand would help.

PR #865, opened 2026-09-09
Pin gcc to keep SOCRATES shim symbols exported on macOS CI

timlichtenberg:
## Description
On macOS CI, a newer Homebrew gcc Darwin driver drops the SOCRATES shim's dlsym-visible symbols (`PS_real_kind_bytes`, `PS_create_StrDim`, ...) from `libSOCRATES_C.so`'s export table, so Julia's `dlsym` lookup in AGNI fails at runtime and the Integration and Slow tiers fail with `could not load symbol`.

This pins gcc to the homebrew-core formula revision whose Darwin driver exports those symbols, installed through a throwaway tap and verified to be both present and the linked keg before the dependent packages install. It also adds a macOS step that runs `nm -gU` on the freshly rebuilt shim and fails fast if `PS_real_kind_bytes` is no longer exported, so a future toolchain regression is caught at setup rather than deep in a test run.

## Validation of changes
I dispatched the full nightly science validation workflow against this branch (run 34328401618): all 21 jobs pass, including the macOS Integration tier and every Slow tier shard that previously failed on the `dlsym` lookup. The version gate and the shim symbol check were exercised on the GitHub-hosted macOS runner image.

## Checklist

### DIFF
--- a/.github/actions/setup-proteus/action.yml
+++ b/.github/actions/setup-proteus/action.yml
@@ -586,7 +586,26 @@ runs:
       run: |
         export RAD_DIR="$GITHUB_WORKSPACE/socrates"
         shim="$RAD_DIR/julia/lib/libSOCRATES_C.so"
+        obj="$RAD_DIR/julia/lib/Utilities_CF.o"
         echo "RAD_DIR=$RAD_DIR"
+        echo "-- gfortran --version --"
+        gfortran --version | head -n 1 || echo "(gfortran not on PATH)"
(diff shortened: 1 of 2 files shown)


nichollsh (approved):

egpbos:
If I'm not mistaken, this issue was already solved through https://github.com/FormingWorlds/SOCRATES/pull/29 and https://github.com/FormingWorlds/SOCRATES/pull/30. The version upper bound shouldn't be necessary anymore.

PR #866, opened 2026-09-10
Bump AGNI to restore the SOCRATES thermal source-function grid

timlichtenberg:
## Description

Bump the AGNI pin from `c7ffdf00` to `8a494d7c` to restore the SOCRATES thermal source-function tabulation grid.

The nightly Integration tier fails on `test_transparent_banded_olr_recovers_blackbody_emission`: the banded transparent OLR at `T_surf = 1500 K` overshoots the black-body emission by about 0.073% (`F_olr` 287271.57 vs 287062.70 W/m^2, tolerance `RTOL_SOCRATES = 5e-4`), on both ubuntu and macos with an identical value, so it is not a toolchain effect. It traces to AGNI commit `66a61da0`, which coarsened the `prep_spec` thermal source-function grid from a roughly 3.25 K step to a 25 K step; the coarser tabulation raises the banded-RT interpolation error enough to tip this tolerance. AGNI restored a finer grid upstream (`ba7bec1e`, then `8a494d7c` at a 5 K step as a runtime-vs-accuracy compromise), so moving the pin forward to `8a494d7c` recovers the accuracy while keeping the hydrostatic-integration and spin parameters that the previous pin bump added. The SOCRATES pin is unchanged and still satisfies AGNI's minimum-version check.

## Validation of changes

Test configuration: Linux and macOS, Python 3.12, real SOCRATES spectral files.

Built AGNI at `8a494d7c` and ran `test_transparent_banded_olr_recovers_blackbody_emission`: it passes over the full surface-temperature sweep (1000, 1500, 2000, 3000 K), all inside `RTOL_SOCRATES`. The full AGNI integration tier (`-m 'integration and not slow'`) passes (44 of 44). The change to `pyproject.toml` is the single-line pin bump; the diff carries no other file. I confirmed by reverting only the thermal-grid constants in the previous pin that the test flips from failing to passing, which isolates the grid coarsening as the cause.

## Checklist

### DIFF
--- a/docs/Reference/module_versions.md
+++ b/docs/Reference/module_versions.md
@@ -35,7 +35,7 @@ pinned commit.
 <!-- BEGIN GIT_TABLE -->
 | Module | Role | Pin | Docs |
 |--------|------|-----|------|
-| AGNI | Radiative-convective atmosphere (Julia) | [![AGNI](https://img.shields.io/badge/AGNI-c7ffdf00-green)](https://github.com/nichollsh/AGNI/commit/c7ffdf000a49439fc597eff3f586c2c54caba647){target="_blank" rel="noopener"} | [Docs](https://www.h-nicholls.space/AGNI/) |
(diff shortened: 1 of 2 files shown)


egpbos:
This was a (partial?) duplicate of #863.

PR #869, opened 2026-09-11
Fill zero-filled SPIDER and atmodeller helpfile columns

### LINKED ISSUE #842: SPIDER and atmodeller runs don't fill important helpfile columns
## Problem description

When using SPIDER and atmodeller, several of the helpfile columns remain unfilled with true values. They are left at zero. These seem to be:
* R_core
* P_center
* P_cmb
* core_density
* core_heatcap
* T_surface_initial (why is this even stored?)
* T_cmb_initial (idem)
* F_cmb

Potentially unrelated to SPIDER, but also the `*_mol_*` keys are also zero-filled. I suspect this is an atmodeller issue.

## Expected behavior

All helpfile columns to be filled appropriately.

## Evidence

[init_coupler.toml.txt](https://github.com/user-attachments/files/31686705/init_coupler.toml.txt)
[runtime_helpfile.csv](https://github.com/user-attachments/files/31686706/runtime_helpfile.csv)

## My computer
Describe the computer used to run the code when the bug was found; e.g Fedora Linux with Python 3.12

## Additional notes

IoA cluster running RHEL. Python 3.12.13 and Julia 1.12.7.
* PROTEUS hash `b06837ea77e9ba1f9b84748922a970b111c49740`.
* spider version 0.2.0
* PETSc version 1.3.19.0
* atmodeller version 1.0.2

## Relevant people

@timlichtenberg @maraattia

timlichtenberg:
## Description
Several helpfile columns stayed zero instead of holding true values. On the SPIDER side, R_core, core_density, core_heatcap, P_cmb, and F_cmb were never set. On the atmodeller side, the four `_mol_*` columns per gas species were never set. Closes #842 for both halves.

calculate_core_mass now writes R_core, core_density, and core_heatcap alongside M_core. ReadSPIDER now reads P_cmb and F_cmb from SPIDER's own timestep JSON, the same way it already reads T_cmb. P_center stays unfilled and is now documented as a SPIDER-side gap: only Zalmoxis provides it.

calc_surface_pressures_atmodeller now fills mol_atm, mol_liquid, mol_total, and mol_solid for every species branch: an in-solve species reads its own mole counts, an inactive noble gas derives them from its escape-owned kg budget, a reactive species excluded from the solve derives them from its kg values via molar mass. The helpfile scanner now carries TEMPLATE_OVERRIDES for these four columns so the reference docs generate cleanly again.

I also removed the T_cmb_initial column. Nothing in the codebase reads it; zalmoxis_solver was its only writer. A resume can load an old helpfile that still carries a retired column, so Proteus.start now drops any column absent from the current schema before the run continues, with a regression test on that path. T_surface_initial keeps its column: aragog and the shared interior_energetics code read it back for the accretion-mode surface temperature override, so it is not part of this bug and stays untouched.

## Validation of changes
Added a failing-then-passing test per writer: core-mass fields, the atmodeller mol-column derivation, and the resume migration. Ran the outgas, interior_energetics, and helpfile-reference test suites plus the docs generator check: 750 passed, 7 pre-existing unrelated skips.

Ran a full SPIDER+atmodeller simulation to completion (16 iterations, no errors). In the resulting helpfile, the atmodeller mol columns are non-zero and evolve smoothly over the run. R_core, P_cmb, core_density, and core_heatcap are non-zero; core_density and core_heatcap resolve from fixed config values for this setup, so their constancy over the run is expected, not a leftover bug. F_cmb evolves from about 8.6e6 to 5.9e4 W/m^2 over the run, consistent with a cooling, crystallizing magma ocean; I did not check its magnitude against literature or against the run's own viscosity and conductivity parameters, so I am filing that as a separate follow-up rather than folding it into this PR.

Test configuration: macOS, Python 3.12, conda env proteus-dev-1.

## Checklist


nichollsh (approved):
Thanks for addressing this issue, @timlichtenberg. I've tested it on my laptop and it seems to work as described. The filled columns are now filled with SPIDER/Atmodeller. I note they are left empty for the dummy/boundary modules, but that seems fine. Good to remove the redundant `T_cmb_initial` key too.

PR #871, opened 2026-09-11
Take netcdf-fortran from conda in the Kapteyn guide

egpbos:
## Description

The Kapteyn module system no longer provides NetCDF-Fortran. `module load netcdf-fortran` fails with `Unable to locate a modulefile`, no modulefile matches `netcdf` or `hdf`, and there is no `nf-config` on `PATH`. PROTEUS therefore cannot be installed by following the guide.

Take the library from conda-forge instead, and set `LD_LIBRARY_PATH` on the conda environment. Both steps are needed: SOCRATES builds against the absolute path `nf-config` reports, but records no run path, so its binaries fail at run time without the variable.

The `NetCDF Error` troubleshooting entry told the reader to rebuild SOCRATES outside conda against the system NetCDF. That now deletes a working checkout and cannot rebuild it, so it is replaced with conda-based checks.

Also fixes the installer pre-flight check, which accepted either `nc-config` or `nf-config` where SOCRATES needs both. On a host carrying only the C interface it reported `System dependencies: OK` and then failed in Phase 4 with `ERROR: NetCDF-Fortran library is not installed`. Kapteyn is such a host.

Habrok and Snellius are untouched.

## Validation of changes

Clean clone of `main` at `ecdfed28` on eddington (Rocky Linux 9.8, x86_64, gfortran 11.5.0, Julia 1.11.9), fresh conda environment created as the guide now describes.

- `bash install.sh --no-data` completes, all 8 phases
- `ldd socrates/bin/l_run_cdf`: `libnetcdff.so.7 => not found` without the variable, resolves to the environment's `lib` with it
- `proteus doctor`: `All checks passed`, reports `SOCRATES: 2603.10 (c3296586)`
- `pytest -m "unit and not skip and not slow and not integration" --ignore=tests/examples`: 3001 passed, 41 skipped
- `PROTEUS_CI_NIGHTLY=1 pytest tests/integration/test_integration_agni_transparent_limit.py tests/integration/test_integration_agni_greygas_interior.py`: 4 passed, including the banded case, which is the only one that runs the SOCRATES binaries

## Checklist

Two tests cover the pre-flight fix: one host with `nc-config` but no `nf-config`, one with neither. Both fail against the previous logic.

### DIFF
--- a/docs/How-to/kapteyn_cluster_guide.md
+++ b/docs/How-to/kapteyn_cluster_guide.md
@@ -61,13 +61,16 @@ Follow the instructions at [VS Code Instructions Kapteyn Cluster](https://docs.g
     source "$HOME/.bashrc"
     ```
 
-5. You can now follow the usual installation steps [here](installation.md), but, since your home folder is capped
-   at 9GB, you need to install Julia and miniconda or conda-forge in "/dataserver/users/formingworlds/<username>".
+5. Install conda and create the PROTEUS environment, by following the
+   [installation steps](installation.md) up to and including
+   `conda activate proteus`. Since your home folder is capped at 9GB, install
+   Julia and miniconda or conda-forge in "/dataserver/users/formingworlds/<username>"
+   rather than in the default location. See [Julia considerations](#julia-considerations)
+   and [Miniconda and conda-forge considerations](#miniconda-and-conda-forge-considerations)
+   below.
 
-    The Kapteyn module system does not provide NetCDF-Fortran, which SOCRATES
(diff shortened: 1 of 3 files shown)


egpbos:
Updated the PR body, also including the fix in commit 95750dba.

nichollsh (approved):
Works for me! I've tried the steps out and it seems to work as described. The description about how the system searches (and sometimes misses) the `netcdff` shared library matches my past experiences too. Also nice that you added some tests for this.

Thanks, @egpbos. Happy to merge.

PR #872, opened 2026-09-11
Fix resume selecting the wrong snapshot when filenames collide

### LINKED ISSUE #795: Resume can restore or delete the wrong snapshot when two rows share a filename
## Problem description

Snapshot files are named from the simulation time rounded to a whole year, while the helpfile stores `Time` at full precision. Two rows written inside the same integer year therefore map to one file, the second write silently overwrites the first, and `select_resumable_snapshot` (`src/proteus/utils/coupler.py`) treats the mapping as one-to-one: it accepts a row if a file with that row's derived name opens, and moves aside whatever exists when a row's pair is incomplete.

The conventions, all of which collapse sub-year steps onto one name:

| writer | filename | rounding |
|---|---|---|
| Aragog | `'%d_int.nc'` | truncates |
| SPIDER | `'%.0f.json'` | rounds |
| AGNI | `'%.0f_atm.nc'` | rounds |
| JANUS | `str(int(Time)) + '_atm.nc'` | truncates |

Two consequences, both on the resume path:

1. A row can be accepted whose file holds a different row's state, so the interior is restored out of step with the helpfile row the run continues from.
2. A row rejected for an incomplete pair takes the shared file with it, which strips the backing from an earlier row that would otherwise have been resumable, and the run rewinds much further than the crash cost it.

Steps to reproduce: run anything that writes more than one row inside one integer year, then kill it. Sub-year steps occur after a giant impact, where the timestep is cut hard, at a crystallization front, and in any configuration with `dt.minimum < 1` yr.

## Expected behavior

A resume either restores the interior state belonging to the row it lands on, or walks back to a row where it can. It should never restore another row's state, and never delete a file that a different row depends on.

## Evidence

Two helpfile rows at t = 70.4 and 70.6 yr with Aragog and AGNI. Aragog truncates, so both rows want `70_int.nc`; AGNI rounds, so they want `70_atm.nc` and `71_atm.nc`. The run is killed after the 70.6 interior snapshot is written and before its atmosphere half, which is the case this function exists to handle. Driving the real `select_resumable_snapshot` over that directory:

```
before: 0_atm 0_int 1_atm 1_int 2_atm 2_int 70_atm 70_int
resumes at t = 2.0   (last stored row was 70.6)
dropped      : [70, 70]
after : 0_atm 0_int 1_atm 1_int 2_atm 2_int
```

Row 70.6 is incomplete, so `70_int.nc` is moved aside and deleted. That file was the only interior backing for row 70.4 as well, so that row fails in turn and its atmosphere goes with it. The run rewinds 68 years and both files at t ≈ 70 are gone.

Reproduction script:

```python
import tempfile
from pathlib import Path
import numpy as np, pandas as pd
from netCDF4 import Dataset
from proteus.utils.coupler import select_resumable_snapshot

def make_nc(path, tag):
    ds = Dataset(str(path), 'w'); ds.row = tag
    ds.createDimension('x', 3)
    ds.createVariable('S', 'f8', ('x',))[:] = np.arange(3.0)
    ds.close()

with tempfile.TemporaryDirectory() as td:
    data = Path(td) / 'data'; data.mkdir()
    times = [0.0, 1.0, 2.0, 70.4, 70.6]
    hf = pd.DataFrame({'Time': times})
    for t in times[:3]:
        make_nc(data / ('%d_int.nc' % t), 'row_%g' % t)
        make_nc(data / ('%.0f_atm.nc' % t), 'atm_%g' % t)
    make_nc(data / '70_int.nc', 'row_70.4')
    make_nc(data / '70_int.nc', 'row_70.6')   # same name, overwrites
    make_nc(data / '70_atm.nc', 'atm_70.4')   # killed before 71_atm.nc
    kept, dropped = select_resumable_snapshot(
        str(Path(td)), hf, require_atm=True,
        interior_module='aragog', atmos_module='agni')
    print('resumes at t =', kept['Time'].tolist()[-1])
    print('dropped      :', dropped)
    print('after        :', sorted(p.name for p in data.glob('*')))
```

## My computer

macOS 15 on Apple silicon, Python 3.12, current `main` plus the accretion branch.

## Additional notes

Three ways to fix it, in the order I would take them:

1. Match rows to files by the time the file records for itself rather than by the name it derives. The interior writers already store it, so a file left by a different step can be told apart from the row's own. This needs no schema change and no new field, and a file that records nothing keeps the current behaviour, so directories written before still resume.
2. Give the filenames sub-year resolution, e.g. `'%.3f_int.nc'`. Cleanest conceptually, but it changes what every existing output directory looks like and needs a compatibility read path for runs already on disk.
3. As a narrow guard, never move aside a file that a different row owns, so a rejected row cannot strip the backing from another one.

**Done on the accretion branch (PR #789), commit `4337d41c`.** Items 1 and 3 are implemented, not by the filename but by the time each file records for itself. Both interior writers already store it, Aragog as a `time` variable in the netCDF and SPIDER as `time_years` in its JSON, so `select_resumable_snapshot` reads it back and matches it against the row: a file recording a different time is not that row's half whatever its name, and is neither resumed from nor moved aside when that row is dropped. That closes both consequences above for the interior half.

The margin is sized from the helpfile's own serialisation, which holds eleven significant digits and so moves a time by up to 4.94e-11 of its magnitude, and it is bounded so it can never reach the one-year filename bucket it exists to split. Past a few Gyr the helpfile cannot separate two rows inside one filename at all, and there the file is accepted on its name rather than a resumable row being refused. A file that records no time is accepted on its name exactly as before, so directories written before the field existed still resume.

**Left to do.**

- The naming itself, item 2 above. The mapping from rows to files is still many-to-one, so a run that writes several rows inside one year still keeps only the last one's snapshot, and a resume still walks back to it rather than to the row where the crash happened. Fixing that means changing the key in all four writers plus `read_last_Sfield`, the archive sweep and the discard path, and reading both conventions for directories written before the change.
- The atmosphere half is still matched on its name alone, because AGNI writes no time into `_atm.nc` (`src/interface/save.jl` defines no such variable). Adding one there, and reading it here, would extend the same guarantee to the atmosphere.
- SPIDER's path is covered at the unit tier (commit `1e400ad3`): a JSON recording another step's time is refused for the row, the row's own resumes, a JSON without the field is accepted on its name, and a dropped row leaves another step's JSON in place. The fixtures use two times that round into the same year, which is the collision SPIDER can produce: `monitor.c` names the file `llround(time * TIMEYRS).json` and writes `time_years` from the same `time` in the same call, so the name and the contents never diverge by more than the rounding tie. What is still untested is a real SPIDER run stopped and resumed, which needs a SPIDER build.

timlichtenberg:
## Description

closes #795

Resume could load, or the archive pruner could delete, a snapshot that did not belong to the helpfile row it was matched to. Snapshot files were named on the simulation time rounded to a whole year, so two rows less than a year apart, or two runs writing into the same output directory, produced files that share a name. Resume then picked a file by name alone, which could be a neighbour's state or a stale file another run left under the same name, and the pruner could remove the wrong one. On a resume this restores the wrong interior and atmosphere state; when two runs collide it can also delete a snapshot the other run still needs.

The fix names Aragog interior and atmosphere snapshots by sub-year time, so two rows less than a year apart no longer collide, and has the interior writers record the time they wrote: a `time` variable in the Aragog netCDF and a `time_years` entry in the SPIDER JSON. On resume the loader reads that recorded time back and confirms a file is the row's own state before loading it, and selects the latest snapshot by time rather than by filename order. A file that records no time is still accepted on its name, so a directory written before this change resumes unchanged. SPIDER keeps its whole-year JSON name because the SPIDER binary writes it and PROTEUS matches that name rather than choosing it; a sub-year rename there needs a SPIDER-side change and is not part of this.

## Validation of changes

Added per-writer regression tests that build a colliding-filename history and check that resume matches each row to the snapshot written for it, loads the correct interior and atmosphere state, and that the pruner keeps the right file. The tests fail on the previous name-only selection and pass with the change. Covered the Aragog and atmosphere writers and the AGNI snapshot glob, whose widest-match selection had the same name-collision gap. Ran the coupler and atmosphere test modules on macOS with Python 3.12.

## Checklist


nichollsh (changes requested):
Thanks for working on this, @timlichtenberg. I believe that it also addresses issue #774 as well?

Switching to a higher precision rounding in all cases (years to 3dp) makes sense. And, as I understand it, this function still provides backwards support for the `%.0f` rounding method, but all files written in the future will adopt the `%.3f` introduced by the new function `snapshot_path_for_time`?

I am a bit worried that the decimal points will cause problems when parsing the file names. E.g. the file `884.700_atm.nc` looks like it might have an extension `.700_atm.nc` depending on how it's parsed. Perhaps writing this as `884p700_atm.nc` is one option?

What do you think? Requesting changes on this PR for now.

timlichtenberg:
Thanks, good suggestions. I implemented the 'p' suggestion and connected fixes.

src/proteus/utils/helper.py
```diff
@@ -79,18 +80,60 @@ def snapshot_path_for_time(data_dir: str, time: float, suffix: str) -> str:
...
+    ----------
+    token : str
+        The numeric portion of a snapshot filename.
+
+    Returns
+    -------
+    float
+        The simulation time [yr].
+    """
+    return float(token.replace('p', '.'))
```
nichollsh, line 141: I think you should add a warning here which checks if there are multiple 'p' characters. For example, I can imagine this causing issues if we happen to end up using file suffixes which contain a 'p' character in the future.

nichollsh (changes requested):
Looks good except one minor suggestion! Thanks, Tim.

timlichtenberg:
Agreed, added a guard that raises ValueError when the token contains more than one p, so a malformed token fails immediately with a clear message instead of producing a confusing float conversion error downstream. Test added in test_helper.py.

nichollsh (approved):

PR #876, opened 2026-09-14
Inference: config validation and failure handling

### LINKED ISSUE #803: A bad reference config fails every inference worker with no reason given
## Problem description

`proteus infer` checks its reference config with `os.path.isfile` and nothing more (`src/proteus/inference/inference.py`, the block that resolves `config['ref_config']`). The file is copied into the output directory and handed to the workers without ever being read against the schema, so a mistake in it survives the whole setup.

The first worker is where it surfaces, and it surfaces badly. `run_proteus` in `src/proteus/inference/objective.py` writes the per-case config out of the reference file and runs `proteus start` as a subprocess with `stdout=subprocess.DEVNULL` and `stderr=subprocess.STDOUT`. The child does refuse an unrecognised key and does say which one, but it says it on stderr, and stderr follows stdout into `/dev/null`. What reaches the log is `Failed to run PROTEUS for worker=0 iter=0; exit code 1`, with nothing pointing at the reference config. Every worker then fails the same way, so a whole optimisation burns through its budget with no usable diagnosis.

The discarded output is not specific to this fault. Any reason a child run dies is thrown away the same way.

## Expected behavior

The reference config is loaded through `read_config_object` at startup, next to the existing `os.path.isfile` check, so a typo stops the run before any worker is spawned and names the offending key.

Separately, when a child run fails, its output is captured and logged rather than discarded.

## Evidence

The child reports the fault, and reports it exactly where the inference module has arranged not to look:

```
$ proteus start -c bad.toml --offline ; echo "exit=$?"
Error: Unrecognised configuration key in bad.toml:
  "not_a_real_key"
  This key is unrecognised, so setting it will have no effect. Check for typos or outdated option names.
exit=1
```

`bad.toml` is `input/dummy.toml` with one unrecognised key added. That message is on stderr, which `run_proteus` sends to `/dev/null`.

## My computer

macOS on arm64, Python 3.12.

## Additional notes

Two independent fixes here. Validating the reference config up front is the one that closes this specific hole. Capturing the child's output is worth doing on its own, since it applies to every way a worker run can fail.

## Relevant people

@nichollsh

stuitje:
## Description

This PR validates the reference PROTEUS config used in an inference run the same way PROTEUS does, so it rejects bad configs immediately without wasting computation time, and improves failure handling in PROTEUS workers. Covers #803. It does **not** yet change the way workers are dispatched (e.g., keeping the Julia instance alive between workers) which is a more drastic change and will be covered in a next PR. 

Files changed:

### Inference files
- `src/proteus/inference/inference.py`
Added `validate_reference_config` / `parameter_bounds`: runs the PROTEUS config checks on the
reference toml and on the two parameter bounds before inference launch. Moved the CPU-count and ref-config-exists checks ahead of safe_rm of the output dir. Implements `abort_on_failure` (user-activated, default off) and calls `summarise_failures` before the results summary.
- `src/proteus/inference/objective.py`
New `ProteusRunFailure` dataclass (status, exit code, logfile, swept parameters, failure/excluded) plus per-run JSON failure records (record_failure /read_failure_records). Child stdout/stderr now logged to <run>_console.log; timeout, non-zero exit, and unreadable helpfile become failures instead of RuntimeErrors. Distinguishes real failures (status 20–28, 0, 1, missing status) from excluded `failure_codes` (eg status 11) and logs them at warning. Added `set_abort_on_failure` / `abort_on_failure`, `apply_nested_updates` (from
update_toml, now during startup validation), `run_output_dir`, `read_status`, `find_run_logfile`. A missing P_surf column now raises explicitly.
- `src/proteus/inference/utils.py`
 New `summarise_failures`: writes failures.csv, logs the failure/excluded by cause. `print_results` warns on unscored optimisation evaluations. 
- `src/proteus/inference/async_BO.py`
worker split into a thin wrapper plus `_worker_loop,` so a dying worker logs its traceback and always releases its busy point. `parallel_process` reports workers that exited non-zero and raises when no optimisation step completed.
- `src/proteus/inference/BO.py`
Busy points selected by worker key rather than list position (a finished worker is absent from B, so positional selection picked the wrong entry); handles the no-other-busy-worker case.
- `src/proteus/inference/plot.py`
`plot_result_correlation` filters the i_* glob to directories.
- src/proteus/inference/gen_D_init.py
Docstring only. `failure_codes` reworded.

### Source files 
- `src/proteus/cli.py`
`ConfigRejectedError` refused config is now logged using the fwl logger at ERROR.
- `src/proteus/utils/logs.py` `attach_worker_logfile`, attaches a file handler to the fwl logger in a worker process that has none.

### Input files 
- `input/inference/example.infer.toml`
Describes`failure_codes` as exclusions, adds `abort_on_failure = false`

### docs 
- `docs/How-to/inference.md`
New 'Failed and excluded simulations' section.

### Other
- `.gitignore`
Ignores *.out and my local config folders I use to test the inference. 

## Validation of changes

I have ran the inference in a new set-up with working and old configs (which should raise). The logged failures work well in my opinion.

Test configuration: Linux, Python 3.12.14.

Tests run:

- `pytest tests/inference/`, all passed
- `pytest -m "unit and not skip and not slow and not integration" --ignore=tests/examples`, passed 

Claude Opus 5.0 generated the new tests according to the test rules. I ran:
```
bash tools/validate_test_structure.sh
python tools/check_test_quality.py --check
```

## Checklist


nichollsh:
Thanks for this @stuitje  - much needed. Let me know when this is ready for review :)

stuitje:
> Thanks for this @stuitje - much needed. Let me know when this is ready for review :)

Hi Harrison, it is ready. Apologies for the extra two commits: I ran another inference run yesterday where I had some different failures than before, and I found it a bit too spammy in the output log, so hence the extra fix.

docs/How-to/inference.md
```diff
@@ -144,10 +145,33 @@ Plots prefixed with `result_` show the results of the optimisation.
...
+
+**Exclusions.** The `failure_codes` field lists completion statuses the study
+does not want to fit against, such as `11` (maximum clock runtime) or `15`
+(volatiles escaped). These are reported at info level.
+
+At the end of the study both kinds are collected into `failures.csv`, which
+carries a `category` column of `failure` or `excluded`, and summarised. If more than half the evaluations went unscored, the summary says so as a warning, because the result then rests on far fewer real evaluations than the step count suggests. If no optimisation evaluation produced a fit quality there is no best fit to report, and the study stops with an error.
+
+Set `abort_on_failure = true` in the inference config to stop at the first failed
+simulation instead. This is useful while setting a study up, when the first failure is more likely to be a mistake in the reference config than an unrunnable corner of the parameter space. It applies to failures only; an excluded outcome never stops the study.
```
nichollsh: The phrasing in this whole section is really strange and very "claude". Please rephrase this to be clearer and less claude-sounding. Specifically, I don't understand why you are referring to "the study" here. Do you mean a given call to the inference scheme? I would also avoid mentioning numerical exit codes in case they change in the future - updating them here adds a lot of maintenance overhead.

src/proteus/inference/async_BO.py
```diff
@@ -103,6 +119,78 @@ def worker(
...
     - output_dir (str): Output directory for the whole inference call (abspath).
+    - logpath (str | None): Study logfile to reopen when this process has no
+      logging configuration of its own. None leaves logging untouched.
+    - log_level (int): Numeric level to log at, read from the parent.
+
+    Returns
+    ----------
+    - None
+    """
+    # A spawned worker inherits no logging configuration on MacOS,
```
nichollsh: Why does behaviour differ on MacOS?

src/proteus/inference/async_BO.py
```diff
@@ -300,4 +395,42 @@ def parallel_process(
...
+        else:
+            # Every worker exited on its first budget check. `max_steps` is
+            # reduced by one per additional worker, so this is what a study
+            # with fewer optimisation steps than workers looks like.
+            cause = (
+                'No worker failed. Each worker stops once the dataset reaches '
+                f'{max_steps} rows ({max_len} requested, less one per worker beyond '
+                f'the first), which the {n_init} initial samples already satisfy. '
+                f'Raise n_steps to at least n_workers ({n_workers}).'
+            )
```
nichollsh, line 420: Again, the phrasing  in this error message is extremely confusing. What exactly is this catching?

src/proteus/inference/objective.py
```diff
@@ -29,6 +32,303 @@
...
+# kept beside the run folder rather than inside it: PROTEUS empties its own
+# output folder once it starts, which would unlink a file held open there and
+# lose exactly the record this capture exists to keep.
+CHILD_CONSOLE_SUFFIX = '_console.log'
+
+# Status written by PROTEUS before its output folder is cleaned, and never
+# rewritten until the main loop starts. A child that dies in between leaves no
+# status file at all, so a missing file is reported as such rather than being
+# silently reported as a generic error.
+STATUS_MISSING = -1
```
nichollsh: I think status=0 would be more appropriate here.

src/proteus/inference/objective.py
```diff
@@ -29,6 +32,303 @@
...
+# Folder inside the study output holding one record per failed evaluation.
+# Written by the workers as they fail and read back once at the end, so that
+# the summary covers initial sampling and optimisation alike without the two
+# paths having to share any state while they run.
+FAILURE_RECORD_DIR = 'failures'
+
+# Fraction of evaluations that may fail before the summary escalates from a
+# report to a warning. Above this, the sampled region is mostly unrunnable and
+# the posterior is built on too few real evaluations to mean much.
+FAILURE_FRACTION_WARN = 0.5
```
nichollsh: This seems like overkill functionality.... Why not just have the code warn if *any* simulations fail. It would be simpler, and it's still useful to know.

src/proteus/inference/objective.py
```diff
@@ -29,6 +32,303 @@
...
+
+# Config entries every worker overwrites in the reference config, regardless of
+# which parameters are being swept. Shared with the startup validation so the
+# configuration that is checked is the configuration that is run.
+WORKER_CONFIG_OVERRIDES = {
```
nichollsh, line 46: You might also wish to consider adding the 'archive' functionality=true here too.

src/proteus/inference/objective.py
```diff
@@ -29,6 +32,303 @@
...
+    """Return the last `lines` lines of a child's console file.
+
+    Returns an empty string when the file is missing or unreadable, so a
+    failure report is still produced when the capture itself went wrong.
+    """
+    try:
+        with open(path, 'r', errors='replace') as f:
+            return _tail(f.read(), lines)
+    except OSError:
+        return ''
```
nichollsh: I am not convinced that showing the last (few) lines of the worker is the best route forward. It seems to strike the wrong balance between avoiding log spam and providing useful information. Instead, I really think that it would be better to record the whole output from a failed/errored child process into a specific file, and then direct the user to that file. I can forsee that only recording the last line will miss important/diagnostic information that should be retained.

nichollsh: Also, keeping the whole child-process output is simpler to do and doesn't require these extra functions.

src/proteus/inference/objective.py
```diff
@@ -29,6 +32,303 @@
...
+        )
+
+
+def set_abort_on_failure(abort: bool = False) -> None:
+    """Record whether a failed child run should abort the whole study.
+
+    Stored in the environment so it is visible to the main process and to any
+    spawned pool workers, matching how the child timeout is plumbed.
+    """
+    os.environ[_ABORT_ON_FAILURE_ENV] = '1' if abort else '0'
```
nichollsh: Is it necessary to have a whole function defined just to perform this one line of code? Instead, maybe just move this `if abort else` and `os.environ` call into the one place in which this function is actually called.

src/proteus/inference/objective.py
```diff
@@ -29,6 +32,303 @@
...
+
+def abort_on_failure() -> bool:
+    """Return whether a failed child run should abort the whole study.
+
+    Defaults to False: an inference sweep is expected to visit parameter
+    combinations the simulator cannot integrate, and treating those as fatal
+    would make most studies unrunnable. Set the inference config field
+    `abort_on_failure` to stop at the first failure instead.
+    """
+    return os.environ.get(_ABORT_ON_FAILURE_ENV, '0') == '1'
```
nichollsh: As above, seems redundant to have a whole function to perform one line of code...

src/proteus/inference/objective.py
```diff
@@ -29,6 +32,303 @@
...
+
+
+def read_status(out_abs: Path | str) -> int:
```
nichollsh: This function is a nice addition and would be better suited to placement within `src/proteus/utils/` rather than specific to the inference scheme. Suggest moving it there.

src/proteus/inference/objective.py
```diff
@@ -149,51 +477,110 @@ def run_proteus(
...
 
     # Handle case where atmosphere has escaped
     #   Set VMRs and MMW to zero
+    if 'P_surf' not in df_row:
+        # The helpfile schema is the same for every run, so a missing column
+        # is a fault of the setup rather than of this sample. Scoring it as a
+        # bad sample would let the study spend its whole budget returning the
+        # same failure value and then report success.
+        raise KeyError(
+            f"Simulator output has no 'P_surf' column ({out_csv}); "
```
nichollsh: Why would this be the case? What causes the helpfile to be missing a `P_surf` column? If this happens somehow, we should address the cause of that behaviour, rather than trying to 'patch' it here.

nichollsh (changes requested):
Thanks for this @stuitje. I've made some in-line comments below regarding the documentation and maintainability of the new code. Hope these are useful.

I ran the example inference config and it still seems to work well. 

Another suggestion: it would be useful to include here an added option for sharing a spectral file between workers/runs. Maybe worth a separate PR - your call - but it would fit within the work here.

src/proteus/inference/objective.py
stuitje: > I am not convinced that showing the last (few) lines of the worker is the best route forward. It seems to strike the wrong balance between avoiding log spam and providing useful information. Instead, I really think that it would be better to record the whole output from a failed/errored child process into a specific file, and then direct the user to that file. I can forsee that only recording the last line will miss important/diagnostic information that should be retained.

I agree, I believed I already fixed this but I see I have to check it again, thanks!

src/proteus/inference/objective.py
stuitje: perhaps I should install ponytail :)

src/proteus/inference/objective.py
stuitje: Makes sense. I do think there is a difference; I've had some runs with about 10% failures now, which still perform quite well overall (good objective, but also good match with ground truth initial params). I might just warn and state the percentage, which is already part of the current functionality.

src/proteus/inference/async_BO.py
stuitje: rephrased

src/proteus/inference/objective.py
stuitje: I am not sure, as status= 0 already exists, meaning "Proteus started". `STATUS_MISSING` should, I would say, really catch a missing status, set to -1, since proteus never writes status -1.

src/proteus/inference/async_BO.py
stuitje: I added this fix after a comment from Copilot in this thread. In multiprocessing in python, MacOS uses the 'spawn' start method by default, whereas Linux uses 'fork'. In the case of 'spawn', the child is fresh and doesn't inherit python objects such as handlers for the fwl logger.

I would force fork everywhere for simplicity but apparently fork is considered unsafe on Mac since python 3.8: https://bugs.python.org/issue33725

I could move everything to spawn since the logging fix is already there, or leave it like this. However, since I was  planning on taking a new look at the whole multiprocessing part of the BO, I could also leave it for that PR. What do you think?

src/proteus/inference/objective.py
stuitje: removed this extra warning

src/proteus/inference/objective.py
stuitje: will add it, good idea

stuitje:
@nichollsh Thank you for the thorough review. I agree with most points. This was the first time I used Claude Code in my work and I agree with you that it sometimes produces spammy or opaque logs, and that it can add unnecessary lines of code. I will fix this with more rigour next time.

I have implemented most of your comments. One vital change from last version: everything to do with failure handling (except `abort_on_failure`) has been moved to `failures.py` and out of `inference.py`, `utils.py` and `objective.py`. The latter was getting long and I think this makes the code much clearer. I am aware this might complicate your review, and I apologise. However I believe it is the best option in the long run.

The few comments that I didn't implement immediately I replied to, perhaps you can take a quick look :)

Before I ask for a re-review, I am looking into the spectral files now, updating the docs, and I want to test the current code again by doing a few inference runs. Thanks again!

src/proteus/inference/objective.py
stuitje: moved

nichollsh:
@stuitje please let me know when this is ready for another review. You'll need to update with changes from the main branch too.

stuitje:
> @stuitje please let me know when this is ready for another review. You'll need to update with changes from the main branch too.

Hi Harrison, I am aware main needs to be merged into this branch, I will do so when I am done. I could not finish the spectral files parts of the code yesterday. 

When I do finish them I will of course let you know. This will probably be on Monday morning, after the weekend.

stuitje:
@nichollsh, the tests are passing again, so this is ready for another review. I added the spectral file cache and exposed it as a config option: please let me know if you disagree; I can change the way it is handled.

stuitje:
@nichollsh, the tests are passing again, so this is ready for another review. I added the spectral file cache and exposed it as a config option: please let me know if you disagree; I can change the way it is handled.

timlichtenberg (changes requested):
Thanks, this is a big step for running inference studies on a cluster. A few things need fixing before this merges, most of them in the new failure path:

`J` crashes on every clean exit with a failure status. `run_proteus(parameters=raw)` adds `params.out.path`, `atmos_clim.spectral_cache` and the worker overrides to `raw` in place, and `ProteusRunFailure.report()` then formats every value with `:g`, which raises `ValueError` on the strings. This hits status 0, 1, -1, 20-28 and everything in `failure_codes`, so ordinary studies reach it. The same mutation gives failures.csv rows of different lengths (11 vs 16 fields), and `read_failure_records` then returns nothing and the summary reports every evaluation as usable. **Please pass a copy and write the columns from the swept keys only, with a test that uses the real `run_proteus` and only mocks `subprocess.run`.**

`abort_on_failure = true` does not stop the BO phase: only the failing worker exits, the others run to the budget and the study ends with exit 0 and a best fit. **A shared event checked in the loop, and a raise after join, would match what the docs promise.**

A log-scaled parameter with a lower bound of 0 passes validation and then fails in every worker, after the old study was already removed. **`parameter_bounds` should reject it.**

Status 29 (planet evaporated) now enters the fit, while main scores it as failed. Either keep it failed or document the change.

Smaller but real: concurrent failures can lose rows in failures.csv (a lock around the header check and the write fixes it), and the end-of-study correlation plot crashes on a case that died at start-up (no `init_coupler.toml`).

Also, 15 comment blocks are longer than 3 lines; please shorten them or move the explanation into the docstrings. And tests/test_cli.py conflicts with main now.

PR #880, opened 2026-09-15
Give `proteus get` a per-user logfile in the temp directory

### LINKED ISSUE #879: proteus get fails on a shared machine because its logfile has a fixed name in the temp directory
`proteus get` writes its logfile to a fixed path inside the system temporary directory (`src/proteus/cli.py`):

```python
log_path = Path(tempfile.gettempdir()) / 'proteus_get.log'
setup_logger(logpath=str(log_path), logterm=True, level='INFO')
```

The directory is portable but the filename is not unique, so on a shared machine every user resolves to the same path. `setup_logger` calls `os.remove(logpath)` on an existing file, and a sticky `/tmp` refuses removal of a file owned by someone else:

```
PermissionError: [Errno 1] Operation not permitted: '/tmp/proteus_get.log'
```

Observed on the Habrok cluster (RUG) on 2026-09-09: `/tmp/proteus_get.log` was owned by a different PROTEUS user from the previous day, `/tmp` was `drwxrwxrwt`, and `proteus get reference`, `proteus get spectral` and `proteus get stellar` all failed. The same commands in the same session succeeded once `TMPDIR` pointed at a private directory, which confirms the cause.

Impact is larger than the traceback suggests: `install.sh` phase 7 downgrades these failures to warnings and still reports `Installation complete!` with exit 0. On a shared cluster the second and later users get an install that claims success with no reference data, and only find out when runs fail or tests skip for missing spectral files.

egpbos:
## Description

Closes #879.

`proteus get` wrote its logfile to a fixed name in the system temporary directory, so on a shared machine every user resolved to the same path and a sticky `/tmp` refused removal of another user's file. The name is now scoped to the caller's uid, which is the ownership rule the kernel applies. A logfile that still cannot be opened, on a read-only or full temp directory, leaves the download running on terminal-only logging instead of aborting it.

## Validation of changes

Three tests in `tests/test_cli.py`, each verified to fail against the unfixed source:

- two users no longer target the same logfile;
- a foreign logfile whose removal raises `PermissionError` is left untouched and does not stop the command;
- `setup_logger` refusing the path warns and lets the download finish.

macOS with Python 3.12. `pytest -m "unit and not skip and not slow and not integration" --ignore=tests/examples` passes except for nine pre-existing `tests/atmos_clim/test_janus.py` failures caused by SOCRATES being absent in this checkout, identical before and after the change. `ruff check`, `ruff format --check` and `tools/validate_test_structure.sh` clean; `tools/check_test_quality.py --check` reports the same violations with and without the new tests.

## Checklist


nichollsh (approved):
Thanks, @egpbos - this works for me!

Perhaps we also need to implement a similar patch, regarding `/tmp` permissions, for the code in PROTEUS which creates temporary directories for other (non-logging) purposes? E.g. when storing `runtime.sf`. What do you think?

timlichtenberg (approved):
Looks good, thanks! Scoping the name to the uid is the right call, and the fallback to terminal logging is much better than aborting the download.

PR #888, opened 2026-09-19
Fix orbit-evection slow test after obliqua merge

timlichtenberg:
## Description
The orbit-evection slow-tier test errored at setup after the obliqua merge, https://github.com/FormingWorlds/PROTEUS/pull/859. Two problems surfaced together: the test helpfile row did not set `core_density`, so `get_C_planet` fell through to a config attribute the test namespace does not have (AttributeError at setup), and the Hansen table spanned k in [-50, 200] while the m=0 mirror requests k down to -200, which mismatched the array shapes in `orbitals()` (251 against 401 modes).

This adds `core_density` to the row, builds the table over [-200, 200], guards the table range against what the mirror requests, and refreshes the module docstring figures to values measured from the current run. Related: https://github.com/FormingWorlds/PROTEUS/pull/859.

## Validation of changes
- `pytest tests/integration/test_slow_orbit_evection_ctl.py` on the branch tip: 4 passed in 629 s. macOS, Python 3.12.
- The new fixture assertion checks that the Hansen table range covers `kmin_kmax_for_m0_mirror`; it evaluates False for the old [-50, 200] table and True for [-200, 200].
- The docstring peak, timing, capture and angular-momentum figures are re-measured from the run (peak e = 0.744 at a' = 12.33 R_earth, capture at t = 2.5e4 yr, `plan_sat_am` drift -4.2%), stated as one run on one machine. No test assertion changed.

## Checklist


timlichtenberg:
@MarijnJ0 Please review, the recent Obliqua PR merge broke the nightly apparently, this should it get back on track.

MarijnJ0 (commented):
These changes look good to me. Iverified the tests pass on my machine as well. The table span test is also a nice addition! 

The discrepancy between the original pinned values and the new values arise due to a slight shift in the evection angle phase that arises due to the inclusing of the new (neglegible) m=0 modes. This does not affect the overall migration substantially. Since the pinned values are still `in_band` the semi major axis and eccentricity still undergo rapid oscillations, so a small change (due to additional m=0 terms) affects the specific moment during this oscillation that we observe at the end, and can give rise to the observed ~4% offset.

Overall, the structure now matches the correct orbit solver structure in PROTEUS/orbit. 

Thanks for catching this Tim!

timlichtenberg:
Thanks a lot! :)

